PR #3, opened 2021-06-22
Add a feature for a `region` CodeList

danielhuppmann:
The `region` CodeList has a slightly different structure of the yaml files: to avoid repeating a "hierarchy" attribute many times (e.g., is this a country or a continent), the yaml files have a structure like
```yaml
<Hierarchy>:
  Region Code:
    Attribute: Attribute value
```
See the added unit test for a more applied example.

### DIFF
--- a/nomenclature/codes.py
+++ b/nomenclature/codes.py
@@ -35,13 +35,16 @@ class CodeList(Mapping):
 
         Parameters
         ----------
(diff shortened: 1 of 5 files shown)


danielhuppmann:
Closing and reopening to start the (just-added) tests

PR #9, opened 2021-06-29
Require path with initializing a Nomenclature, validate returns True

### LINKED ISSUE #8: Relative definition of path in Nomenclature
The relative definition of the `path` variable in the `__init__` of  `Nomenclature` `def __init__(self, path="definitions")` can lead to problems when the workflow where it is used is not started from the parent directory of `definitions`.

One possible fix would be to remove the default and require it to be set explicitly in every workflow in the following pattern:

`Nomenclature(Path(__file__).absolute().parent / "definitions").validate(df)`

Another fix would be checking if the directory specified in `path` actually exists and throw an error if it doesn't. Might also be a usecase for pydantic as it specifies an `FilePath` datatype which does this validation automatically (https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types)

danielhuppmann:
This PR makes `path` a required argument when initializing a Nomenclature, and it makes `validate()` a bit more user-friendly by casting its argument to an IamDataFrame if necessary. Also, `validate()` returns True so that it can be used as:

```
if nc.validate(df):
    <do something>
```

closes #8


tests/test_core.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+import nomenclature as nc
+
+
+def test_nonexisting_path_raises():
```
phackstock, line 5: If you change the test in `nomenclature/core.py` to check for a directory, this needs of course to be reflected in the unittest

nomenclature/validation.py
```diff
@@ -19,6 +19,9 @@ def is_subset(x, y):
 def validate(nc, df):
     """Validation of an IamDataFrame against codelists of a Nomenclature"""
 
+    if not isinstance(df, IamDataFrame):
+        df = IamDataFrame(df)
```
phackstock, line 23: At this conversion attempt we assume that converting into an IamDataFrame is possible. Should we do checks before trying to convert or does the constructor of the IamDataFrame take care of that?

phackstock (commented):
Some small comments about dataframe conversion and checking for a directory

nomenclature/core.py
```diff
@@ -7,10 +7,13 @@
 class Nomenclature:
     """A nomenclature with codelists for all dimensions used in the IAMC data format"""
 
-    def __init__(self, path="definitions"):
+    def __init__(self, path):
         if not isinstance(path, Path):
             path = Path(path)
 
+        if not path.exists():
```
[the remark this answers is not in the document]
danielhuppmann: thanks, good suggestion - implemented

tests/test_core.py
danielhuppmann: done

nomenclature/validation.py
danielhuppmann: I think it is helpful for the user to directly see which kind of error caused the error when `df` cannot be cast to an IamDataFrame

nomenclature/core.py
```diff
@@ -7,10 +7,13 @@
 class Nomenclature:
     """A nomenclature with codelists for all dimensions used in the IAMC data format"""
 
-    def __init__(self, path="definitions"):
+    def __init__(self, path):
         if not isinstance(path, Path):
             path = Path(path)
 
+        if not path.is_dir():
+            raise FileNotFoundError(f"Definitions directory not found: {path}")
```
phackstock, line 15: Maybe change to [`NotADirectoryError`](https://docs.python.org/3/library/exceptions.html#NotADirectoryError) since we're specifically looking for a directory. Maybe I'm also being overly knit-picky, in this case ignore ;)

danielhuppmann: 👍

phackstock: wonderful

phackstock (approved):
lgtm

PR #13, opened 2021-07-07
Add `to_excel()` method and test

danielhuppmann:
This PR adds a method to write the `variable` codelist to an Excel file.


danielhuppmann:
See #15

Issue #20, opened 2021-09-01
Region-Mapping for scenario explorer workflows


phackstock:
Opening the issue per discussion with @danielhuppmann. 

We want a function which can do region aggregation given two inputs: `df: IamDataFrame` and `mapping_path: pathlib.Path`. It should return the aggregated `IamDataFrame`.

Some open points: 

* Should the returned dataframe **only** contain the aggregated regions or also the original constituent regions?
* Should the function make use of the region aggregation capabilities of [`IamDataFrame.rename()`](https://github.com/IAMconsortium/pyam/blob/5f10925022f90f2b6b2ecbef14e935b38e3852a2/pyam/core.py#L1003) or [IamDataFrame.aggregate_region()](https://github.com/IAMconsortium/pyam/blob/5f10925022f90f2b6b2ecbef14e935b38e3852a2/pyam/core.py#L1328)
* How do we preform region aggregation for variables where a simple sum is not the correct way to aggregate?

stefanpauliuk:
>     Should the function make use of the region aggregation capabilities of IamDataFrame.rename() or IamDataFrame.aggregate_region()

The common way of aggregating data (assume data are in some multi-index table format) is a matrix multiplication:
Agg_Data = Agg_Matrix.dot(Full_Data)
The 'old' classification disappears and the new classification is returned.
Therefore, I prefer to only return data for the new (target) set of regions.

> How do we preform region aggregation for variables where a simple sum is not the correct way to aggregate?

for an n:1 aggregation, the Agg_Matrix can be condensed as list: which old position is added to which new position, and the method builds the Agg_Matrix with 1 at the relevant positions.
for an n:m aggregation, one option would to actually supply the aggregation matrix, containing fractions, could be supplied directly or constructed from a dictionary with the structure { Old_region: [New_region, Fraction]}

danielhuppmann:
Thanks @stefanpauliuk for chiming in. Your answer is correct in a matrix sense, but pyam uses a different data format internally - and the question was intended about which of the existing pyam methods to use.

About the "not just simple sum", this was meant about things like a weighted average, not splitting model-native regions across compare-regions. This would be an interesting use case, but is currently not supported by pyam, and would probably require some more thought to implement correctly.

[For example: how would "take 50% of region A" work in a emissions-weighted carbon price aggregation? But this is not a discussion for here, but should be a feature-request discussion in the pyam repository.]

phackstock:
[...]

PR #22, opened 2021-09-24
Feature/region mapping

phackstock:
This is a first draft of a PR regarding #20. 

As a first step I implemented 3 pydantic classes located in `nomenclature/region_mapping_models.py`:
1. **NativeRegion**: A model native region, has a name and optionally a renaming parameter for the scenario explorer)
2. **CommonRegion**: A common/comparison/aggregation region, has a name and a list of constituent model native regions
3. **RegionAggregationMapping**: A container for an entire model mapping, has the name of its model, and either a list of native regions or a list of common regions (or both)

The structure is inspired by the mapping used in the ar6-workflow (https://github.com/iiasa/ar6-workflow/blob/master/region_mapping/MESSAGEix-GLOBIOM_1.0.yml). 
This structure looks as follows:

```yaml
model: MESSAGEix-Materials 1.1
native_regions:
  - AFR: Africa
  - WEU
common_regions:
  - Africa:
    - AFR
  - Europe:
    - WEU
```

One problem I encountered with this structure is that is more implicit than explicit. Meaning that for the values given under 'native_regions' and 'common_regions' the user, or the computer for that matter, needs to know that in the case of 'native_regions' the first value is the name of the native region and an optional second value allows renaming. For 'common_regions', the outer level specifies the name of the common region while the inner list contains the names of this common region's constituent regions. 
In explicit notation this would look like this:
```yaml
model: MESSAGEix-Materials 1.1
native_regions: 
- name: AFR
  rename: Africa
- name: WEU
common_regions:
- name: Africa
  constituent_regions:
  - name: AFR
- name: Europe
  constituent_regions:
  - name: WEU
```
This would tell pydantic everything it needs to know to parse this mapping but it is of course rather verbose and not that easy for a human to type and read. 
Both of these file examples can be found in `tests/data/region_aggregation/`under `message_mapping.yml` and `message_mapping_proper.yml` respectively.
The first question to answer is which format we want to go with:
* The 'original/implicit' one which is shorter but cannot take advantage of the pydantic parsing features (at least to my knowledge we cannot, maybe @meksor has an idea)
* The 'explicit' one which is longer but can be parsed right out of the box

In case we want to stick with the first I have written a parsing function `convert_region_mapping` in `nomenclature/region_mapping_models.py` that can read a yaml file like that, transforms the resulting dictionary and feeds it into `RegionAggregationMapping`. Even though this is working it requires a lot more code than just straight up pydantic and is therefore more error prone and potentially more work to maintain. 

There is are two unit tests in `tests/test_region_aggregation` that verify that the data are correctly read and that the two approaches yield the same results.

For now I would say the  main question we need to answer is what format we want to go with.
Any other comments are of course welcome as well.


danielhuppmann:
Thanks for the initial PR

I think that the non-intuitive feel of the mapping is due to the very stylized example.  If you make this a bit more elaborate, I think it becomes more obvious how this should be used... 

```yaml
model: MESSAGEix-Materials 1.1
native_regions:
  - AFR: "MESSAGE-R11|Africa"
  - WEU: "MESSAGE-R11|Western Europe"
  - EEU: "MESSAGE-R11|Eastern Europe"
common_regions:
  - Africa:
    - AFR
  - Europe:
    - WEU
    - EEU
```

In response to the trade-off between 
1. verbose config files and short code 
2. more concise config files and some parsing-utility code, 
I'd have a slight preference for option 2.

nomenclature/region_mapping_models.py
```diff
@@ -0,0 +1,71 @@
...
+    native_regions: Optional[List[NativeRegion]]
+    common_regions: Optional[List[CommonRegion]]
+
+
+# should probably be moved inside RegionAggregationMapping as a classmethod
+def convert_region_mapping(file: Union[Path, str]) -> RegionAggregationMapping:
+    with open(file, "r") as stream:
+        mapping_input = yaml.safe_load(stream)
+    # check if we have a key called
+    # we need a model name and either native or common regions
```
danielhuppmann, line 32: Not sure if we need to have either native-regions or common-regions. From the point of view of a scenario submission, if there is no mapping, the data is imported as is (without any renaming or aggregation).

nomenclature/region_mapping_models.py
```diff
@@ -0,0 +1,71 @@
...
+    if not "model" and ("native_regions" or "common_regions") in mapping_input.keys():
+        raise ValueError(
+            "Region mapping needs an attribute 'model' and at "
+            "least one of the following: 'native_regions',"
+            "'common_regions'"
+        )
+    if "native_regions" in mapping_input.keys():
```
danielhuppmann, line 39: No need to specify `keys()`...
```suggestion
    if "native_regions" in mapping_input:
```

nomenclature/region_mapping_models.py
```diff
@@ -0,0 +1,71 @@
...
+                    {"name": list(nr)[0], "rename": list(nr.values())[0]}
+                )
+            else:
+                raise ValueError(
+                    f"Encountered error with {nr}, "
+                    "native_regions can only contain single "
+                    "entries, or single name: rename type mappings."
+                )
+        mapping_input["native_regions"] = native_region_list
+    if "common_regions" in mapping_input.keys():
```
danielhuppmann, line 55: No need to specify `keys()`...
```suggestion
    if "common_regions" in mapping_input:
```

nomenclature/region_mapping_models.py
```diff
@@ -0,0 +1,71 @@
...
+                native_region_list.append({"name": nr})
+            elif isinstance(nr, dict) and len(nr) == 1:
+                native_region_list.append(
+                    {"name": list(nr)[0], "rename": list(nr.values())[0]}
+                )
+            else:
+                raise ValueError(
+                    f"Encountered error with {nr}, "
+                    "native_regions can only contain single "
+                    "entries, or single name: rename type mappings."
```
danielhuppmann, line 52:
```suggestion
                    f"Invalid specification of native-region `{nr}`, "
                    "value must be a string or a mapping of the type `<name>: <rename>`."
```

danielhuppmann (commented):
A few comments inline

nomenclature/region_mapping_models.py
phackstock: If there is no mapping at all will we still have a yml file? Would this then just contain the model name and nothing else?
I wrote that comment under the assumption that we only create a yml file if we either do renaming or aggregation or both.

danielhuppmann: I'd see that as two separate questions:
 - Do we create a yaml file (e.g., as part of the model registration) if there is no renaming as part of the processing? This could be yes or no, but no need to make a decision on this now.
 - Do we require that a yaml file can **only** be valid if it has some renaming? This makes the code more complicated and limits our options for the first question later on, when we implement a model registration workflow. So I'd say that we do **not require** this...

phackstock: Ok, I've updated the PR. Now I allow only a model and there's a test for that.

danielhuppmann:
Thanks @phackstock. The example now used is really confusing because the native-to-db_region_name mapping has the same target as a "common region" name. Also, please remove the verbose example.

phackstock:
Ah yeah sorry @danielhuppmann for that. I've updated the PR now. The testcases should be more clear now.
The main change is that I now use a json schema in the validation process. The way that it works now is the following:
1. In `convert_region_mapping` I first read the input data as well as the json schema `region_mapping_schema.json`. _Two comments on that:_
    1.  Currently I use a [json schema](http://json-schema.org/) to validate a yaml file. It is working as json and yaml are strongly related languages but it might be a bit confusing to other maintainers. There is [yaml schema](https://asdf-standard.readthedocs.io/en/1.6.0/schemas/yaml_schema.html#yaml-schema) as well so I could convert it but I found it easier to wrap my head around json schema first. 
    1.  Once `convert_region_mapping` this is finished it should probably be turned into a classmethod of `RegionAggregationMapping` so that it can act almost as an alternative `__init__` function.
1. The input data validation is performed using [`jsonschema.validate`](https://python-jsonschema.readthedocs.io/en/latest/validate/#jsonschema.validate). This ensures that the data conforms to the schema `region_mapping_schema.json` which makes restructuring it to fit into `RegionAggregationMapping` a lot easier and requires much less code. 
1. The restructuring is performed, the resulting dictionary is fed into RegionAggregationMapping and the result is returned. _One question on that topic_:
    1. The way that I have written both the json schema and the data restructuring part right now to allow for two different ways to define the common regions: 
        1. Option 1 (found in `tests/data/working_message_mapping_cr_dict.yml`):
        ```yaml
        common_regions:
          Eastern Europe, Caucasus and Central Asia:
            - EEU
            - FSU
          Europe:
            - WEU
        ```
        2. Option 2 (found in `tests/data/working_message_mapping_cr_list.yml`):
        ```yaml
        common_regions:
          - Eastern Europe, Caucasus and Central Asia:
            - EEU
            - FSU
          - Europe:
            - WEU
        ```
    1. The difference being that in option 1 we don't have a hyphen ("-") in front of the common region name while for option 2 we do. This is read by the parser in two different ways: `{"common_regions": {"Eastern Europe...": ["EEU", "FSU"], "Europe": ["WEU"]}}` and `{"common_regions": [{"Eastern Europe...": ["EEU", "FSU"]}, {"Europe": ["WEU"]}]}`. In option 1 we get a dictionary of the common regions while in option 2 we get a list of dictionaries. 
    1. Currently both the json schema and the dictionary restructuring allow both both versions but that makes both of them more verbose and harder to understand. 
    1. Therefore my question, should we decide to only support one of the two options and if so which one? My preference would be to **only** support option **2**.

danielhuppmann (commented):
Looks great, a few suggestions inline, plus general remarks:

 - Please rename the failing-test files to more descriptive names than just enumeration
 - I'd suggest to use a more generic test case, like:
 ```
model: model_a
native_regions:
  - region_a: alternative_name_a
  - region_b
  - region_c: alternative_name_c
common_regions:
  - common_region_1:
    - region_a
    - region_b
  - common_region_2:
    - region_c
 ```

I'd actually then merge the PR as is and tackle the integration in the `DataStructureDefinition` as a follow-up PR - the discussion here is already quite long, and the current status (after clean-up) is a nice, compact addition: "Parse region-mapping config file". Maybe also change the PR name to make sure that this is reflected in the commit history of `main`.

nomenclature/region_mapping_models.py
```diff
@@ -0,0 +1,97 @@
...
+                elif isinstance(nr, dict):
+                    native_region_list.append(
+                        {"name": list(nr)[0], "rename": list(nr.values())[0]}
+                    )
+            mapping_input["native_regions"] = native_region_list
+
+        # Reformat the "common_regions"
+        if "common_regions" in mapping_input:
+            common_region_list: List[Dict[str, List[Dict[str, str]]]] = []
+            for cr in mapping_input["common_regions"]:
```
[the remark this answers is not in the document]
phackstock: Ah yeah that makes it a bit more clear.

phackstock:
Sounds good from my side. 
I'll rename the models and regions in the test cases to something more generic.
Regarding the names of the failing test cases, I've put a description of the error condition in the first line of each yaml file while keeping the name generic. I've done that so that it's clearly visible which yaml files are failure conditions and putting it in the file name would result in either a super long name or a loss of information. If you want the names to be descriptive though I can try be brief and descriptive while keeping the comment in the file for more details

nomenclature/region_mapping_models.py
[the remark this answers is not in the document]
phackstock: Sorry, actually that needs to be just `for cr in mapping_input["common_regions"]:` since "common_regions" is a list of common_regions and not a dictionary.

tests/test_region_aggregation.py
```diff
@@ -0,0 +1,92 @@
...
+        ),
+        (
+            "illegal_mapping_illegal_attribute.yaml",
+            ValidationError,
+            "Additional properties are not allowed.*",
+        ),
+        (
+            "illegal_mapping_conflict_regions.yaml",
+            ValueError,
+            ".*Conflict between \(renamed\).*common_region_1.*",
```
danielhuppmann, line 56: Could work to just add `r` before the string, see https://www.flake8rules.com/rules/W605.html

.gitignore
```diff
@@ -145,3 +145,7 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# prototyping files
+
+workbench.py
```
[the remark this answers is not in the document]
phackstock: Done

nomenclature/region_mapping_models.py
```diff
@@ -0,0 +1,103 @@
...
+            nr.target_native_region for nr in values["native_regions"]
+        }
+        common_region_names = {cr.name for cr in values["common_regions"]}
+        overlap = list(native_region_names & common_region_names)
+        if not overlap:
+            return values
+        raise ValueError(
+            "Conflict between (renamed) native regions and aggregation mapping"
+            f" to common regions: {overlap}"
+        )
```
danielhuppmann, line 66: Suggest to change the order for more logical reading...
```suggestion
        if overlap:
            raise ValueError(
                "Conflict between (renamed) native regions and aggregation mapping"
                f" to common regions: {overlap}"
            )
        return values
```

nomenclature/region_mapping_models.py
```diff
@@ -0,0 +1,103 @@
...
+    common_regions: Optional[List[CommonRegion]]
+
+    @validator("native_regions")
+    def validate_native_regions(cls, v):
+        target_names = [nr.target_native_region for nr in v]
+        doubles = [item for item, count in Counter(target_names).items() if count > 1]
+        if doubles:
+            raise ValueError(
+                f"Two or more native regions share the same name: {doubles}"
+            )
```
danielhuppmann, line 38: Minor suggestion for readability...
```suggestion
        duplicates = [item for item, count in Counter(target_names).items() if count > 1]
        if duplicates:
            raise ValueError(
                f"Two or more native regions share the same name: {duplicates}"
            )
```

nomenclature/region_mapping_models.py
```diff
@@ -0,0 +1,103 @@
...
+        return v
+
+    @validator("common_regions")
+    def validate_common_regions(cls, v):
+        names = [cr.name for cr in v]
+        doubles = [item for item, count in Counter(names).items() if count > 1]
+        if doubles:
+            raise ValueError(
+                f"Duplicated aggregation mapping to common regions: {doubles}"
+            )
```
danielhuppmann, line 48: As above
```suggestion
        duplicates = [item for item, count in Counter(names).items() if count > 1]
        if duplicates:
            raise ValueError(
                f"Duplicated aggregation mapping to common regions: {duplicates}"
            )
```

nomenclature/region_mapping_models.py
```diff
@@ -0,0 +1,103 @@
...
+        return v
+
+    @root_validator()
+    def check_illegal_renaming(cls, values):
+        """Check if any renaming overlaps with common regions"""
+        if values.get("native_regions") is None or values.get("common_regions") is None:
+            return values
```
danielhuppmann, line 55: Add a comment to explain why this fast-pass
```suggestion
        # Skip if only either native-regions or common-regions are specified
        if values.get("native_regions") is None or values.get("common_regions") is None:
            return values
```

nomenclature/validation_schemas/region_mapping_schema.yaml
```diff
@@ -0,0 +1,52 @@
+$schema: 'https://json-schema.org/draft/2020-12/schema'
+title: RegionAggregationMapping
+description: >
+  This schema is used to do the first stage of validation for yaml region
+  mapping files.
```
danielhuppmann, line 5: Shortening to one line (below 88 characters)
```suggestion
  This schema is used for the first stage of validation for yaml region mapping files.
```

nomenclature/validation_schemas/region_mapping_schema.yaml
```diff
@@ -0,0 +1,52 @@
...
+required:
+  - model
+additionalProperties: false
+definitions:
+  NativeRegion:
+    title: NativeRegion
+    oneOf:
+      - type: object
+        patternProperties:
+          ^.+$:
```
danielhuppmann, line 32: Is it possible to add a comment here? Maybe "the key of the dictionary is the native-region name"...

nomenclature/validation_schemas/region_mapping_schema.yaml
```diff
@@ -0,0 +1,52 @@
...
+            type: string
+        additionalProperties: false
+        maxProperties: 1
+        minProperties: 1
+      - type: string
+  CommonRegion:
+    title: CommonRegion
+    type: object
+    patternProperties:
+      ^.+$:
```
danielhuppmann, line 43: As above, if possible please add a comment

tests/test_region_aggregation.py
```diff
@@ -0,0 +1,92 @@
...
+def test_illegal_mappings(file, error_type, error_msg_pattern):
+    # This is to test a few different failure conditions
+
+    with pytest.raises(error_type, match=error_msg_pattern):
+        RegionAggregationMapping.create_from_region_mapping(test_folder / file)
+
+
+def test_model_only_mapping():
+    # test that a region mapping runs also with only a model
+    reference = {
```
danielhuppmann, line 84: Maybe rename this to `exp` and the other variable to `obs`?

tests/test_region_aggregation.py
```diff
@@ -0,0 +1,92 @@
...
+
+from conftest import TEST_DATA_DIR
+
+test_folder = TEST_DATA_DIR / "region_aggregation"
+
+
+def test_mapping():
+    mapping_file = "working_mapping.yaml"
+    # Test that the file is read and represented correctly
+    ram = RegionAggregationMapping.create_from_region_mapping(
```
danielhuppmann, line 13: Maybe rename this to `obs` and the other variable to `exp`?

PR #23, opened 2021-10-01
Add check for duplicates when creating yaml from xlsx

danielhuppmann:
This PR implements a validation that there are no duplicate codes (i.e., variables) when creating a yaml-file from an Excel spreadsheet.


phackstock (approved):
Looks good to me

Issue #16, opened 2021-07-08
Shorten `file` attribute in codelists


danielhuppmann:
All items in a codelist have an attribute `file` to identify from which file the code was imported - this might be helpful in debugging in more complex nomenclatures.

The file name should be saved relative to the definitions folder, not relative to the working directory.

PR #34, opened 2021-11-02
Feature/error enhancement region aggregation

### LINKED ISSUE #33: Add file name to errors in RegionAggregationMapping
Currently, if we have an error validating a `RegionAggregationMapping` we only get the cause of the error but not the file where that mapping is saved. In case of a big number of mapping files searching for the correct one can become a tedious task. Especially when thinking about further validation steps, i.e. checking for model collisions and validating against `DataStructureDefinition`.

### Proposed fix
Add the attribute `mapping_file` to `RegionAggregationMapping` and update the involved methods and tests.

What are your thoughts on that @danielhuppmann?

phackstock:
closes #33

### Changes
* Added a property `file` to `RegionAggregationMapping`. The type of `file` is [`pydantic.types.FilePath`](https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types) which conveniently validates that `file` points to a file. 
* Added this file information to validation errors inside `RegionAggregationMapping` for better error reporting.

As #32 should build on this, my preference would be to merge this first.

### Open Points

Right now the error specify the entire file path, maybe it would be better to limit it to the file name plus maybe one or two parent folders. Instead of `/home/user/.../mappings/some_mapping_file.yaml` we would get `mappings/some_mapping_file.yaml` or maybe only `some_mapping_file.yaml`. What do you think @danielhuppmann?
It might even be safety relevant as I'm not sure we want to give the entire file path of our server side folder structure in case this is running as part of a scenario explorer instance.

### DIFF
--- a/nomenclature/region_mapping_models.py
+++ b/nomenclature/region_mapping_models.py
@@ -26,7 +26,7 @@ class CommonRegion(BaseModel):
 
 class RegionAggregationMapping(BaseModel):
     model: str
-    mapping_file: FilePath
+    file: FilePath
     native_regions: Optional[List[NativeRegion]]
     common_regions: Optional[List[CommonRegion]]
 
@@ -66,7 +66,7 @@ class RegionAggregationMapping(BaseModel):
         if overlap:
             raise ValueError(
                 "Conflict between (renamed) native regions and aggregation mapping"
-                f" to common regions: {overlap} in {values['mapping_file']}"
+                f" to common regions: {overlap} in {values['file']}"
             )
         return values
 
@@ -86,7 +86,7 @@ class RegionAggregationMapping(BaseModel):
             raise ValidationError(f"{e.message} in {file}")
 
         # Add the file name to mapping_input
-        mapping_input["mapping_file"] = file
+        mapping_input["file"] = file
 
         # Reformat the "native_regions"
         if "native_regions" in mapping_input:
--- a/nomenclature/region_mapping_models.py
+++ b/nomenclature/region_mapping_models.py
@@ -1,9 +1,10 @@
 from pathlib import Path
 from typing import Dict, List, Optional, Union
 from collections import Counter
+from pydantic.types import FilePath
 
 import yaml
-from jsonschema import validate
+from jsonschema import validate, ValidationError
 from pydantic import BaseModel, validator, root_validator
 
 here = Path(__file__).parent.absolute()
@@ -25,28 +26,29 @@ class CommonRegion(BaseModel):
 
 class RegionAggregationMapping(BaseModel):
     model: str
+    mapping_file: FilePath
     native_regions: Optional[List[NativeRegion]]
     common_regions: Optional[List[CommonRegion]]
 
     @validator("native_regions")
-    def validate_native_regions(cls, v):
+    def validate_native_regions(cls, v, values):
         target_names = [nr.target_native_region for nr in v]
         duplicates = [
             item for item, count in Counter(target_names).items() if count > 1
         ]
         if duplicates:
             raise ValueError(
-                f"Two or more native regions share the same name: {duplicates}"
+                f"Two or more native regions share the same name: {duplicates} in {values['mapping_file']}"
             )
         return v
 
     @validator("common_regions")
-    def validate_common_regions(cls, v):
+    def validate_common_regions(cls, v, values):
         names = [cr.name for cr in v]
         duplicates = [item for item, count in Counter(names).items() if count > 1]
         if duplicates:
             raise ValueError(
-                f"Duplicated aggregation mapping to common regions: {duplicates}"
+                f"Duplicated aggregation mapping to common regions: {duplicates} in {values['mapping_file']}"
             )
         return v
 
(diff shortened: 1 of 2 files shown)


danielhuppmann:
Re the shorter paths, see the related issue #16 - in this case, I think the path should be relative to the "project" directory (ie. something like `/mappings/[<subfolder>/]<name>.yaml`)

danielhuppmann:
Appending the file path at the end makes for hard-to-read messages, see
```python
ValidationError: 1 validation error for RegionAggregationMapping
__root__
  Conflict between (renamed) native regions and aggregation mapping to common regions: ['common_region_1'] in tests/data/region_aggregation/illegal_mapping_conflict_regions.yaml (type=value_error)
```

Is it a lot of work to insert it before the first line break, i.e.,
```python
ValidationError: 1 validation error for RegionAggregationMapping in `region_aggregation/illegal_mapping_conflict_regions.yaml`:
__root__
  Conflict between (renamed) native regions and aggregation mapping to common regions: ['common_region_1' (type=value_error)
```

nomenclature/region_mapping_models.py
```diff
@@ -25,28 +26,29 @@ class CommonRegion(BaseModel):
 
 class RegionAggregationMapping(BaseModel):
     model: str
+    mapping_file: FilePath
```
danielhuppmann, line 29: Suggest to name that attribute just `file` in line with the codelist structure, see https://github.com/IAMconsortium/nomenclature/blob/72a142236050f6bda0ee4f80b96f77e8b17ea52b/nomenclature/codes.py#L122

phackstock:
It would be awfully convenient to be able to comment on comments to have threads ...
* Re the paths: Sounds good, how should we handle this for testing? Since the files don't live a `mapping` directory?
* Re error message: I really like that idea, I'll see if that can be done easily.

nomenclature/region_mapping_models.py

danielhuppmann:
Re the relative paths, I guess "relative to working directory" is the best we can do, and the code already implements that anyway... We just have to make sure that when the package is used in production mode, the working directory is the project directory.

phackstock:
Ah yeah that's a good, pragmatic solution.

phackstock:
Paths are now relative in the error messages.

phackstock:
From my view those changes are ready to be merged. The custom modified error messages (phackstock/nomenclature#1) are optional in my opinion.

nomenclature/region_mapping_models.py
```diff
@@ -25,28 +26,31 @@ class CommonRegion(BaseModel):
...
-    def validate_native_regions(cls, v):
+    def validate_native_regions(cls, v, values):
         target_names = [nr.target_native_region for nr in v]
         duplicates = [
             item for item, count in Counter(target_names).items() if count > 1
         ]
         if duplicates:
             raise ValueError(
-                f"Two or more native regions share the same name: {duplicates}"
+                f"Two or more native regions share the same name: {duplicates} in "
```
danielhuppmann, line 41: I think that the three ValueErrors are sufficiently similar so that moving the actual raising of the error to a utility method like `_raise_error(msg, values)` would improve the code.

danielhuppmann: In particular because of the addition `.relative_to(Path.cwd())`...

phackstock: Ah yes, good point but I think I'd rather create a custom error, something along the lines of `NameOverlapError`.

danielhuppmann: Nice!

nomenclature/region_mapping_models.py
```diff
@@ -77,7 +82,14 @@ def create_from_region_mapping(cls, file: Union[Path, str]):
...
         # Validate the input data using jsonschema
-        validate(mapping_input, schema)
+        try:
+            validate(mapping_input, schema)
+        except ValidationError as e:
+            # Add file information in case of error
+            raise ValidationError(f"{e.message} in {file}")
+
+        # Add the file name to mapping_input
+        mapping_input["file"] = file
```
danielhuppmann, line 92: Looking at this PR again, I guess it would be better to only store the relative path as attribute in the first place? (rather than doing that in the error messages)

phackstock: I was thinking about that too, if there's no value in keeping the full path (for debugging, error tracing, etc..) then I'll just change it to the relative path.

phackstock:
I created a custom error class `NameCollisionError` which inherits from `ValueError`. Now the raising of those errors in the validators is more concise. The only thing I couldn't get to work is for pytest to accept my new custom error as the error type in the assertion. For now I still used `ValueError` which is what `NameCollisionError` is subclassed from anyway. So it's not completely wrong just not as precise as possible. Not sure at this point if this the problem is connected to my custom error, pytest or pydantic or a combination of the above.

nomenclature/region_mapping_models.py
phackstock: I briefly tried saving only the relative path but that caused some tests to fail. I now included the conversion of the path in the `RegionNameCollisionError` class.

phackstock:
Should be all good for now. My above comment about not being able to assert the right type of error has been resolved.
For now the only thing remaining is that by default pydantic writes all attributes of an error class at the end of an error message. It's a pretty minor thing though.

danielhuppmann (approved):
Looks good, thanks @phackstock - please remove the commented-out section, then go ahead and merge!

PR #35, opened 2021-11-05
Check allowed regions

### LINKED ISSUE #26: Check for allowed regions
Building on #22, the next step should be to integrate `RegionAggregationMapping` with `DataStructureDefinition` in order to verify that all specified native regions as well as common regions are allowed.

phackstock:
closes #26 

For now a draft PR as I ran into an issue with the modified error messages while adding the validation for allowed regions using a `DataStructureDefinition` instance.
The issue has to do with how pydantic handles errors, in particular when multiple error occur during evaluation. Pydantic collects them all and wraps them in one `pydantic.ValidationError` at the end. The example below illustrates this:

```python
pydantic.error_wrappers.ValidationError: 2 validation errors for RegionProcessor in /tests/data/regionprocessor_working/mapping_2.yaml
mappings -> model_b
  Region(s) ['region_a', 'common_region_1'] in /tests/data/regionprocessor_working/mapping_2.yaml not defined (type=value_error.regionnotdefined; file=/tests/data/regionprocessor_working/mapping_2.yaml)
mappings -> model_a
  Region(s) ['region_a', 'common_region_1'] in /tests/data/regionprocessor_working/mapping_1.yaml not defined (type=value_error.regionnotdefined; file=/tests/data/regionprocessor_working/mapping_1.yaml)
```

The problem is now that when I previously overwrote the `__str__` function of `pydantic.ValidationError` it was under the assumption that one `ValidationError` would only ever correspond to one single mapping source file. Now we have potentially a lot of mapping files as we run through all mappings. I could just list them in the first line but I think that would make it not very clean.
So I would vote to pivot back to the original way pydantic formats errors and just have the file in the line with the error as it is also the case in the above example.
What do you think @danielhuppmann?

### DIFF
--- a/nomenclature/region_mapping_models.py
+++ b/nomenclature/region_mapping_models.py
@@ -1,9 +1,8 @@
-import copy
 from collections import Counter
 from pathlib import Path
 from typing import Dict, List, Optional, Set, Tuple, Union
 
-import pydantic
+from pydantic.errors import PydanticValueError
 import yaml
 from jsonschema import ValidationError, validate
 from pydantic import BaseModel, root_validator, validator, validate_arguments
@@ -13,42 +12,18 @@ from nomenclature.core import DataStructureDefinition
 
 here = Path(__file__).parent.absolute()
 
-# We take a deep copy of the original __str__ from
-# pydantic.error_wrappers.ValidationError. We do this to keep the changes minimally
-# invasive and get 'automatic' updates in case of any changes upstream
-original__str__ = copy.deepcopy(pydantic.error_wrappers.ValidationError.__str__)
-
-# Define a new __str__ method which adds file information in case it is present.
-# Otherwise the original __str__ method is used.
-
-
-def new__str__(self):
-    """Change __str__ from pydantic ValidationError to include the file name if
-    present"""
-    if "ctx" in self.errors()[0] and "file" in self.errors()[0]["ctx"]:
-        return original__str__(self).replace(
-            "\n", f" in {self.errors()[0]['ctx']['file']}\n", 1
-        )
-    return original__str__(self)
-
-
-# Overwrite the original __str__ with new__str__
-pydantic.error_wrappers.ValidationError.__str__ = new__str__
-
 Sequence = Union[List[str], Tuple[str], Set[str]]
 
 
-class RegionNameCollisionError(ValueError):
-    template = "Name collision in {location} for {duplicates}"
+class RegionNameCollisionError(PydanticValueError):
+    code = "region_name_collision"
+    msg_template = "Name collision in {location} for {duplicates} in {file}"
 
     def __init__(self, location: str, duplicates: Sequence, file: Path) -> None:
-        self.file = file.relative_to(Path.cwd())
         super().__init__(
-            self.template.format(
-                location=location,
-                duplicates=duplicates,
-                rel_file=self.file,
-            )
+            location=location,
+            duplicates=duplicates,
+            file=file.relative_to(Path.cwd()),
         )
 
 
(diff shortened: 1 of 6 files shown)


danielhuppmann:
Yes, makes sense to revert the error message approach.

phackstock:
Sounds good, I'll pivot back.

nomenclature/region_mapping_models.py
```diff
@@ -165,19 +181,39 @@ def __init__(
         super().__init__(self.template.format(model=mapping1.model, files=files))
 
 
+class RegionNotDefinedError(ValueError):
+    template = "Region(s) {region} in {file} not defined"
```
danielhuppmann, line 185:
```suggestion
    template = "Region(s) {region} in {file} not defined in the DataStructureDefinition"
```

nomenclature/region_mapping_models.py
```diff
@@ -165,19 +181,39 @@ def __init__(
...
+
+    def __init__(self, region, file) -> None:
+        self.file = file
+        super().__init__(self.template.format(region=region, file=file))
+
+
 class RegionProcessor(BaseModel):
     """Region aggregation mappings for scenario processing"""
 
+    dsd: DataStructureDefinition
```
danielhuppmann, line 195: Maybe better use `definition`?

nomenclature/region_mapping_models.py
```diff
@@ -165,19 +181,39 @@ def __init__(
...
     mappings: Dict[str, RegionAggregationMapping]
 
+    class Config:
+        # necessary since DataStructureDefinition is not a pydantic class
+        arbitrary_types_allowed = True
+
     @classmethod
-    @validate_arguments
-    def from_directory(cls, path: DirectoryPath):
+    @validate_arguments(config=dict(arbitrary_types_allowed=True))
```
danielhuppmann, line 203: Just asking - do you really need both the extra class and the config-dict?

nomenclature/region_mapping_models.py
```diff
@@ -151,6 +153,20 @@ def create_from_region_mapping(cls, file: Union[Path, str]):
             mapping_input["common_regions"] = common_region_list
         return cls(**mapping_input)
 
+    @property
+    def region_names_all(self) -> List[str]:
+        nr_list = (
+            [x.target_native_region for x in self.native_regions]
+            if self.native_regions is not None
+            else []
```
danielhuppmann, line 161: Shouldn't this work?
```suggestion
            [x.target_native_region for x in self.native_regions or list()]
```

nomenclature/region_mapping_models.py
```diff
@@ -151,6 +153,20 @@ def create_from_region_mapping(cls, file: Union[Path, str]):
             mapping_input["common_regions"] = common_region_list
         return cls(**mapping_input)
 
+    @property
+    def region_names_all(self) -> List[str]:
```
danielhuppmann, line 157: Given that we have an attribute `native_regions` and an attribute `common_regions`, I suggest for `all_regions` here...

danielhuppmann (commented):
A few initial thoughts inline...

nomenclature/region_mapping_models.py
phackstock: Yes because I am asking pydantic to validate the function inputs and one of the inputs is `DataStructureDefinition` which is not a pydantic class. Therefore pydantic would throw an error as it attempts to validate it so I need to set `arbitrary_types_allowed` to `True`. See [pydantic documentation](https://pydantic-docs.helpmanual.io/usage/types/#arbitrary-types-allowed) and this [stackoverflow post](https://stackoverflow.com/questions/68349150/best-practice-to-pydantic-validate-arguments-for-non-built-in-types-like-pandas) about the `@validate_arguments` decorator specifically.

nomenclature/region_mapping_models.py
phackstock: Ah nice, that should work.

nomenclature/region_mapping_models.py

phackstock:
Added some more tests, should be good from my side.

danielhuppmann (approved):

PR #37, opened 2021-11-09
Region processing

### LINKED ISSUE #20: Region-Mapping for scenario explorer workflows

phackstock:
closes #20 

Implements region processing which consists of two parts:

1. Renaming of model native regions
2. Aggregation of model native regions to comparison regions

Added the core functionality, `RegionProcessor.apply(df: IamDataFrame)`, and three region processing tests:

1. Test renaming only which is comprised of three aspects:
    1. Renamed regions should be renamed, 
    2. Named regions without a rename should be kept
    3. Regions which are not specified should be dropped
2. Test that a model **without** a mapping is simply passed through unchanged
3. Test that if common regions are specified they are correctly aggregated

Missing for now:
- [x] A test that combines all three features using an IamDataFrame which holds multiple different models
- [x] Weighted aggregation according to the specification in DataStructureDefinition. For now it just performs simple summation. 

Especially `test_region_aggregation.py` and `RegionProcessor.apply()` are a little on the rough side for now and would probably benefit from some clean up. 
Maybe the two parts of `RegionProcessor.apply()`, renaming and aggregating, should be moved to separate functions (perhaps something like `_rename` and `_aggregate`) to make for cleaner code. Additionally, it might be worth turning the list of target native region names (`[x.target_native_region for x in cm.native_regions]`) into a property of `RegionAggregationMapping`, again for cleaner code.


phackstock:
Especially `test_region_aggregation.py` and `RegionProcessor.apply()` are a little on the rough side for now and would probably benefit from some clean up. 
Maybe the two parts of `RegionProcessor.apply()`, renaming and aggregating, should be moved to separate functions (perhaps something like `_rename` and `_aggregate`) to make for cleaner code. Additionally, it might be worth turning the list of target native region names (`[x.target_native_region for x in cm.native_regions]`) into a property of `RegionAggregationMapping`, again for cleaner code.

nomenclature/region_mapping_models.py
```diff
@@ -188,3 +197,34 @@ def validate_mapping(cls, v, values):
...
+            if model not in self.mappings:
+                all_dfs.append(model_df)
+            # Otherwise we first rename, then aggregate
+            else:
+                cm = self.mappings[model]
+
+                # Rename
+                if cm.native_regions is not None:
+                    all_dfs.append(
+                        model_df.rename(region=cm.renaming_dict).filter(
```
danielhuppmann, line 216: I'm wondering... We assume that the native-region-renaming is a 1-to-1-mapping, so first filtering, then renaming should be faster...

nomenclature/region_mapping_models.py
```diff
@@ -134,6 +137,12 @@ def all_regions(self) -> List[str]:
         cr_list = [x.name for x in self.common_regions or []]
         return nr_list + cr_list
 
+    @property
+    def renaming_dict(self) -> Dict[str, str]:
```
danielhuppmann, line 141: I think the name `rename_mapping` more intuitive...

nomenclature/region_mapping_models.py
```diff
@@ -134,6 +137,12 @@ def all_regions(self) -> List[str]:
...
+        # NOTE: right now this implementation would imply that all not defined model
```
danielhuppmann, line 142: I would put that explicitly into a documentation page (for now as part of the main Readme of the repo)...

tests/data/region_processing/dsd/variables/tag_fuel.yaml
```diff
@@ -0,0 +1,3 @@
+- <Fuel>:
```
danielhuppmann, line 1: I would not use a `DataStructureDefintion` with a tag here - the test should be as focused as possible.

nomenclature/region_mapping_models.py
```diff
@@ -188,3 +197,34 @@ def validate_mapping(cls, v, values):
         if invalid:
             raise RegionNotDefinedError(invalid, v.file)
         return v
+
+    def apply(self, df: IamDataFrame) -> IamDataFrame:
+        all_dfs: List[IamDataFrame] = []
```
danielhuppmann, line 202: maybe rename to `processed_dfs`?

nomenclature/region_mapping_models.py
```diff
@@ -188,3 +197,34 @@ def validate_mapping(cls, v, values):
...
+        for model in df.model:
+            model_df = df.filter(model=model)
+
+            # If no mapping is defined the data frame is returned unchanged
+            if model not in self.mappings:
+                all_dfs.append(model_df)
+            # Otherwise we first rename, then aggregate
+            else:
+                cm = self.mappings[model]
```
danielhuppmann, line 211: Why `cm`?

nomenclature/region_mapping_models.py
```diff
@@ -188,3 +197,34 @@ def validate_mapping(cls, v, values):
...
+                        )
+                    )
+
+                # Aggregate
+                if cm.common_regions is not None:
+                    for cr in cm.common_regions:
+                        agg_df = model_df.aggregate_region(
+                            model_df.variable,
+                            cr.name,
+                            [r.name for r in cr.constituent_regions],
```
danielhuppmann, line 227: I was looking at this line for quite a while, thinking why `constituent_regions` has to be iterated over... After some digging, I realized that `CommonRegion.constituent_regions` is a list of `NativeRegion` instances - but why is that necessary? Would a list of strings be sufficient - and make the code a lot more readable?

danielhuppmann (commented):
Looks good, many thanks! A few suggestions inline, looking forward to the next iteration!

nomenclature/region_mapping_models.py
phackstock: Good point, changed

nomenclature/region_mapping_models.py
phackstock: Done

nomenclature/region_mapping_models.py
phackstock: You're right, it is not strictly necessary I made that design choice on more of a conceptual basis to emphasis the fact that a common region is comprised of one or more model native regions.
I can change it to just a list of strings if you want.
Alternatively, I could also add a property `constituent_region_names` or something along those lines to `RegionAggregationMapping`. This way we could move the list comprehension out of the apply function.

nomenclature/region_mapping_models.py
phackstock: Added some explanation to the README

tests/data/region_processing/dsd/variables/tag_fuel.yaml
phackstock: Done

nomenclature/region_mapping_models.py
phackstock: Done

nomenclature/region_mapping_models.py
phackstock: `cm` was supposed to be `current mapping` but I guess that's anything but obvious, I'll take it out and just use `self.mappings[model]`

nomenclature/region_mapping_models.py
danielhuppmann: Still a bit convoluted, but fine to add a new attribute as you suggest.

phackstock: Now that I think about it just a list of strings is probably the better way to go since I don't use the constituent regions for anything else other than selecting model native regions.
I would vote for simplifying it, we can always revert back to using the full `NativeRegion` model if need be.

phackstock:
Weighted aggregation now works. Any of the pyam.IamDataFrame.aggregate_region kwargs can be used.
Added one test for weighted sum aggregation.

README.md
```diff
@@ -11,7 +11,7 @@ This repository is licensed under the Apache License, Version 2.0 (the "License"
 ## Overview
 
 This package facilitates working with data templates that follow the format developed by
-the [Integrated Assessment Modeling Consortium (IAMC)](https://www.iamconsortium.org).
+the [Integrated Assessment Modeling Consortium (IAMC)](https://www.iamconsortium.org). Additionally, it supports region processing which consists of renaming and aggregation of model native regions to common regions.
```
danielhuppmann, line 14: Please also follow a max-88-character line length in the readme.
```suggestion
the [Integrated Assessment Modeling Consortium (IAMC)](https://www.iamconsortium.org). 
It supports validation of scenario data and region processing, which consists of
renaming and aggregation of model "native regions" to "common regions" used in a project.
```

README.md
```diff
@@ -58,7 +62,42 @@ When importing the codelist, any occurrence of `<Tag>` in a variable name will b
...
+  - region_b
+common_regions:
+  - common_region_1:
+    - region_a
+    - region_b
+  - common_region_2:
+    - ...
+```
+
+Some points to note:
```
danielhuppmann, line 92:
```suggestion
### Notes
```

README.md
```diff
@@ -58,7 +62,42 @@ When importing the codelist, any occurrence of `<Tag>` in a variable name will b
...
+
+* The names of the three top level key words `model`, `native_regions` and `common_regions` are fixed.
```
danielhuppmann, line 94:
```suggestion
* The names of the three top level keywords `model`, `native_regions` and `common_regions` are fixed.
```

README.md
```diff
@@ -58,7 +62,42 @@ When importing the codelist, any occurrence of `<Tag>` in a variable name will b
...
+* Inside the `native_regions` a **list** of model native regions serves as a selection as to which regions to keep.
+* In the above example `region_a` is to be renamed to `alternative_name_a`. This is done by defining a key: value pair of `model_native_name: new_name`.
```
danielhuppmann, line 96:
```suggestion
* In the above example `region_a` is to be renamed to `alternative_name_a`.
   This is done by defining a key-value pair of `model_native_name: new_name`.
```

danielhuppmann (approved):
Looks good, a few comments on the documentation, then good to be merged. Maybe structure the Notes-section explicitly as a list of the known keywords of the mapping file, with notes for each keyword - I think this would improve readability.

phackstock:
Another round of updates:
* README is now up to date
* `RegionAggregationMapping` now requires at least `native_regions` or `common_regions` to be provided. Both are fine too.
* From the above change follows that model only mappings are no longer allowed. This is reflected in a unit test using `tests/data/region_aggregation/illegal_mapping_model_only.yaml`
* `CommonRegion.constituent_regions` is now `List[str]` no longer `List[CommonRegions]`. The latter was adding unnecessary complexity.

danielhuppmann:
Looks good to me, fine to be merged!

PR #39, opened 2021-11-11
Add test_valid_mappings to CLI 'validate-project'

phackstock:
Edited after suggestion from @danielhuppmann:

~~This PR adds a function `assert_valid_mappings` to `nomenclature/testing.py`.~~
~~This function is called as the last step in `cli_valid_project`.~~

This PR adds checking the mappings to `assert_valid_structure`.

Added two unit tests to test the new feature, one to run, one to fail.

### DIFF
--- a/nomenclature/testing.py
+++ b/nomenclature/testing.py
@@ -22,6 +22,7 @@ def cli_valid_yaml(path: Path):
 def cli_valid_project(path: Path):
     assert_valid_yaml(path)
     assert_valid_structure(path)
+    assert_valid_mappings(path)
 
 
 def assert_valid_yaml(path: Path):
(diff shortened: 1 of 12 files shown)


nomenclature/testing.py
```diff
@@ -50,4 +51,11 @@ def assert_valid_structure(path: Path):
     nomenclature.DataStructureDefinition(path / "definitions")
 
 
+def assert_valid_mappings(path: Path):
```
danielhuppmann, line 54: With this implementation, the parsing of the DataStructureDefinition happens twice - that seems not very efficient. In my biew, the "structure" is both the definitions and the mappings, so this could be just one testing function.

phackstock: Ah right, ok, I'll update that.

phackstock: Done.

LauWien (approved):
Nice :)

danielhuppmann (approved):
Nice work!

phackstock:
Thanks @LauWien for getting the CLI started, made it really easy to add functionality :)

PR #44, opened 2021-11-12
Additional types of code attribute values

danielhuppmann:
This PR extends the list of allowed values of the attributes of individual code items in a codelist based on usage in the openENTRANCE nomenclature prototype:

In particular, this PR allows
 - the unit of a variable can be a list of strings (but not a list of strings and a null), because either a variable is dimensionless (null) or it has one or several units
 - the attributes of a variable (other than unit) can also be a boolean
 - the attributes of a region can also be boolean (e.g., EU-membership) or a list (e.g., list of synonyms for a country name)

And after a lot of back-and-forth, I came to the conclusion that using singular for the folders in a definition directory (`variable` & `region` instead of `variables` & `regions`) is preferable, because this will make it easier to add custom dimensions and it brings the folder name in line with the attribute of a `DataStructureDefinition` and an `IamDataFrame`.


nomenclature/validation_schemas/tag_schema.yaml
```diff
@@ -32,7 +32,7 @@ definitions:
         # Attributes-dictionary (replace `<Tag>` by `value` in the `key` attribute)
         type: object
         additionalProperties:
-          type: [string, number, "null"]
+          type: [string, number, boolean, "null"]
```
phackstock, line 35: Just tested it without the `additionalProperties` tag and all tests still run through fine. To me this suggests that either this property is not required or there is a unit test missing that covers this.

nomenclature/validation_schemas/variable_schema.yaml
```diff
@@ -17,11 +17,15 @@ definitions:
...
           unit:
-            type: [ string, "null" ]
+            oneOf:
+              - type: [ string, "null" ]
+              - type: array
+                items:
+                  type: string
         required:
           - unit
         additionalProperties:
```
phackstock, line 27: Also just tested this without the `additionalProperties` tag and all unit tests still run through perfectly. Maybe something's not right here.

nomenclature/codes.py
```diff
@@ -105,14 +104,14 @@ def from_directory(
             # validate against the schema of this codelist domain
             validate(_code_list, SCHEMA_MAPPING[name])
 
-            # if specified, set top-level key as attribute instead
-            if top_level_attr is not None:
+            # a "region" codelist assumes a top-level key to be used as attribute
+            if name == "region":
                 original_code_list, _code_list = _code_list.copy(), []
```
phackstock, line 109: `_code_list.copy()` creates a shallow copy. Is that fine or do we actually want a deep copy here?

nomenclature/core.py
```diff
@@ -16,21 +16,26 @@ class DataStructureDefinition:
...
+        self.region = CodeList.from_directory("region", path / "region")
 
-        args = [(self.region, "region"), (self.variable, "variable")]
-        for attr, name in args:
-            if not attr:
-                logger.warning(f"Attribute {name} is empty.")
+        self.dimensions = ["region", "variable"]
+        for dim in self.dimensions:
+            if not self.__getattribute__(dim):
+                logger.warning(f"Attribute '{dim}' is empty.")
```
phackstock, line 38: Does this need a warning? Are the cases where empty attributes are allowed not covered by the jsonschema?

phackstock (commented):
See comments below. 
One more clarification question, do you mean by "the attributes of a variable or region can be boolean" that the values of attributes can be boolean?

nomenclature/core.py
danielhuppmann: This is a warning not related to the attribute of a code (e.g., the variable Primary Energy), but if the entire CodeList is empty, e.g., if not a single variable is defined. This can happen if no "variable" folder is found. In this case, any validation will fail.

Might be worthwhile to make that an error.

nomenclature/validation_schemas/tag_schema.yaml
danielhuppmann: I didn't add a test, because we should have confidence that **jsonschema** works as expected with correct specifications.

nomenclature/validation_schemas/variable_schema.yaml
danielhuppmann: As above, I didn't add a test for all possible cases, because we should have confidence that **jsonschema** works as expected with correct specifications.

nomenclature/codes.py
danielhuppmann: Refactored the copy to avoid the copy alltogether...

danielhuppmann:
Thanks for the review @phackstock.

> One more clarification question, do you mean by "the attributes of a variable or region can be boolean" that the values of attributes can be boolean?

Yes, see this example:
https://github.com/openENTRANCE/nomenclature/blob/b26b5bdb330a7f714dce972212942002620a2fa8/nomenclature/definitions/region/countries.yaml#L6

nomenclature/validation_schemas/tag_schema.yaml
phackstock: I didn't mean testing that jsonschema works, I meant testing that the schema correctly specifies the structure that we expect.

nomenclature/validation_schemas/variable_schema.yaml
```diff
@@ -17,11 +17,15 @@ definitions:
...
-          type: [ string, number, "null" ]
+          type: [ string, number, boolean, "null" ]
```
phackstock, line 28: Since you say in the description of the PR that a variable can also have a list as an attribute value, is here and `array` missing?
```suggestion
          type: [ array, string, number, boolean, "null" ]
```

phackstock (commented):
Looks good just one small remark

nomenclature/core.py
phackstock: Ah I see, in that case, unless we want to allow a situation where no variables are defined I would vote for throwing an error.

nomenclature/validation_schemas/tag_schema.yaml
danielhuppmann: Yes, I added these additional types because I needed them to migrate the openENTRANCE-nomenclature...

Anyway, I added some more items to the test-data to cover more cases explicitly.

nomenclature/core.py
danielhuppmann: Done

nomenclature/validation_schemas/tag_schema.yaml
phackstock: Perfect

nomenclature/validation_schemas/variable_schema.yaml
danielhuppmann: No, only the unit can be a list at this point, all other attributes of a variable cannot be, because I encountered an issue with that - it's not needed right now, so let's keep that for another iteration.

phackstock: Ah ok, then I would suggest editing the PR description because under the second point I read it as any variable attribute can be a list.

danielhuppmann: Done

phackstock (approved):
Then I'd say we're good to merge.

PR #45, opened 2021-11-15
Make `DataStructureDefinition` dimensions customizable

### LINKED ISSUE #16: Shorten `file` attribute in codelists

danielhuppmann:
This PR allows to specify "custom" (aka generic) CodeList dimensions for a `DataStructureDefinition`. This can be used to validate against an IamDataFrame against a list of valid "scenario" names or (if it exists) a list of "subannual" categorical timeslices.

As a side bonus, the PR also closes #16.

### DIFF
--- a/nomenclature/core.py
+++ b/nomenclature/core.py
@@ -23,7 +23,8 @@ class DataStructureDefinition:
         path : str or path-like
             The folder with the project definitions.
         dimensions : list of str, optional
-            List of :meth:`CodeList` names, initialized from a sub-folder of `path`.
+            List of :meth:`CodeList` names. Each CodeList is initialized
+            from a sub-folder of `path` of that name.
         """
         if not isinstance(path, Path):
             path = Path(path)
@@ -32,13 +33,12 @@ class DataStructureDefinition:
             raise NotADirectoryError(f"Definitions directory not found: {path}")
 
         self.dimensions = dimensions
-        for dim in dimensions:
+        for dim in self.dimensions:
             self.__setattr__(dim, CodeList.from_directory(dim, path / dim))
 
         empty = [d for d in self.dimensions if not self.__getattribute__(d)]
         if empty:
-            _empty = ", ".join(empty)
-            raise ValueError(f"Empty codelist: {_empty}")
+            raise ValueError(f"Empty codelist: {', '.join(empty)}")
 
     def validate(self, df: IamDataFrame, dimensions: list = None) -> None:
         """Validate that the coordinates of `df` are defined in the codelists
--- a/nomenclature/core.py
+++ b/nomenclature/core.py
@@ -22,7 +22,7 @@ class DataStructureDefinition:
         ----------
         path : str or path-like
             The folder with the project definitions.
-        dimensions : list of str
+        dimensions : list of str, optional
             List of :meth:`CodeList` names, initialized from a sub-folder of `path`.
         """
         if not isinstance(path, Path):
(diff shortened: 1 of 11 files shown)


nomenclature/core.py
```diff
@@ -15,37 +15,41 @@
...
-        self.variable = CodeList.from_directory("variable", path / "variable")
-        self.region = CodeList.from_directory("region", path / "region")
+        self.dimensions = dimensions
+        for dim in dimensions:
+            self.__setattr__(dim, CodeList.from_directory(dim, path / dim))
 
-        self.dimensions = ["region", "variable"]
         empty = [d for d in self.dimensions if not self.__getattribute__(d)]
         if empty:
             _empty = ", ".join(empty)
```
phackstock, line 40: Unrelated to the PR itself but I would say that `_empty` is not needed, the concatenation can be done inline directly in the `ValueError(f"Empty codelist: {', '.join(empty)}")`

nomenclature/core.py
```diff
@@ -15,37 +15,41 @@
...
-    def __init__(self, path):
+    def __init__(self, path, dimensions=["region", "variable"]):
         """
 
         Parameters
         ----------
         path : str or path-like
             The folder with the project definitions.
+        dimensions : list of str, optional
+            List of :meth:`CodeList` names, initialized from a sub-folder of `path`.
```
phackstock, line 26:
```suggestion
            List of :meth:`CodeList` names, initialized from a sub-folder of `path` of the `CodeList` names.
```
or something along those lines to stress the fact that this attempts to read e.g. `path / "region"`

nomenclature/validation.py
```diff
@@ -16,48 +16,46 @@ def is_subset(x, y):
...
+                # fast-pass for unique units in df and the DataStructureDefinition
+                if dsd_unit == unit:
+                    continue
+                # full-fledged subset validation
+                if is_subset(unit, dsd_unit):
+                    continue
+                invalid_units.append((variable, unit, dsd_unit))
+
+        if invalid_vars:
+            log_error("variable", invalid_vars)
```
phackstock, line 44: Not sure I like the design of just logging the invalid values and the referring to them in the error. Having the invalid values directly in an error would be better in my opinion. I understand that getting all invalid values in one single go is more efficient but pydantic does it by wrapping all the errrors raised during validation into a single validation error that contains a list of all errors.
Maybe that's overkill though, but seeing as we already have a function called `log_error` which in effect acts as a custom error it might not be.

nomenclature/core.py
```diff
@@ -15,37 +15,41 @@
...
         if not isinstance(path, Path):
             path = Path(path)
 
         if not path.is_dir():
             raise NotADirectoryError(f"Definitions directory not found: {path}")
 
-        self.variable = CodeList.from_directory("variable", path / "variable")
-        self.region = CodeList.from_directory("region", path / "region")
+        self.dimensions = dimensions
+        for dim in dimensions:
```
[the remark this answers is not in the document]
danielhuppmann: Done

nomenclature/core.py
danielhuppmann: Done, I had a longer error message earlier which was too long for one line...

nomenclature/core.py
danielhuppmann: Done

nomenclature/validation.py
danielhuppmann: I agree that this pattern is not very elegant, but I'd keep it for now and then tackle this in a follow-up after release 0.1 - can you create an issue?

phackstock (approved):
Looks good to me, ready to merge.

PR #49, opened 2021-11-16
Skip region aggregation

### LINKED ISSUE #40: Region processing: skip variable
When refactoring the https://github.com/openENTRANCE/nomenclature to work with this new package, I realized that it would be helpful to have a "skip"-option for variables in a DataStructureDefinition.

Options:

1.
```yaml
Complicated Variable:
 - region-processor: skip
```
2.
```yaml
Complicated Variable:
 - skip-region-processing: true
```

Or maybe use "exclude" instead of skip?

Any thoughts @phackstock @peterkolp?

phackstock:
closes #40.

It is now possible to skip region aggregation by adding: `skip-region-aggregation: true` to a variable definition.

Three things to note:
* The boolean value `true` needs to be lowercase
* By default region aggregation is **not** skipped. The current pattern is opt-out.
* The variable will still be kept for model native regions.

### DIFF
--- a/nomenclature/region_mapping_models.py
+++ b/nomenclature/region_mapping_models.py
@@ -257,6 +257,7 @@ class RegionProcessor(BaseModel):
                             }
                             for var, kwargs in self.definition.variable.items()
                             if var in model_df.variable
+                            and not kwargs.get("skip-region-aggregation", False)
                         }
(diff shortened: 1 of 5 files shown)


danielhuppmann (approved):
Nice, thanks @phackstock!

PR #50, opened 2021-11-16
Performance improvement for RegionProcessor.apply

### LINKED ISSUE #48: Performance improvement for region-aggregation processing
The current implementation iterates over all common-regions, then creating the variable-kwargs-dictionary, then iterating over each variable.

Two ways to significantly improve performance:
1. create variable-kwargs-dictionary before iterating over common regions
2. The pyam `aggregate_region()` method can take a list of variables if there are no additional arguments (weight, method, ...). So the variable-kwargs-dictionary could be distinguished into a "summed variables"-list plus a "other-method variables" dictionary.

phackstock:
closes #48

Implemented the two points from #48:
* Moved the creation of the variable kwargs dict outside of the common region loop.
* Split aggregation between simple variables and variables that specify pyam aggregation kwargs (e.g. weighted aggregation, or function other than sum)


nomenclature/region_mapping_models.py
```diff
@@ -242,23 +242,23 @@ def apply(self, df: IamDataFrame) -> IamDataFrame:
...
-                        vars = {
-                            var: {
-                                key: value
-                                for key, value in kwargs.items()
-                                if key in PYAM_AGG_KWARGS
-                            }
-                            for var, kwargs in self.definition.variable.items()
-                            if var in model_df.variable
-                            and not kwargs.get("skip-region-aggregation", False)
+                    PYAM_AGG_KWARGS = {
```
danielhuppmann, line 245: I'd move that to the top of the file.

danielhuppmann:
Just tested this PR with a real-world example from the ECEMF project. Processing time went from ~330 seconds to 8 seconds.

nomenclature/region_mapping_models.py
```diff
@@ -241,29 +248,38 @@ def apply(self, df: IamDataFrame) -> IamDataFrame:
...
-                            agg_df = model_df.aggregate_region(
-                                var,
-                                cr.name,
-                                cr.constituent_regions,
-                                **kwargs,
+                        for var, kwargs in self.definition.variable.items()
+                        if var in model_df.variable
+                        and not kwargs.get("skip-region-aggregation", False)
+                    }
+                    simple_vars = [var for var, kwargs in vars.items() if not kwargs]
```
danielhuppmann, line 261: I'd go for a more descriptive name, like `vars_default_args`, and `vars_kwargs`.

nomenclature/region_mapping_models.py
```diff
@@ -241,29 +248,38 @@ def apply(self, df: IamDataFrame) -> IamDataFrame:
...
                 # Aggregate
                 if self.mappings[model].common_regions is not None:
-                    for cr in self.mappings[model].common_regions:
-                        PYAM_AGG_KWARGS = {
-                            "components",
-                            "method",
-                            "weight",
-                            "drop_negative_weights",
+                    vars = {
+                        var: {
```
danielhuppmann, line 252: for readability, I'd pull that out into an own utility function, e.g., `_filter_dict_args(kwargs, keys)` with `keys` defaulting to  `PYAM_AGG_KWARGS`

danielhuppmann (commented):
Looks great, a few minor comments inline

nomenclature/region_mapping_models.py

nomenclature/region_mapping_models.py

danielhuppmann (approved):
Nice, thank you!

PR #51, opened 2021-11-16
Use relative paths in RegionAggregationMapping

### LINKED ISSUE #46: Shorten `file` attribute of a RegionAggregationMapping
Similar to #16 and #45, ideally, the `file` attribute of a `RegionAggregationMapping` should be relative to the `mappings` folder of a project.

phackstock:
closes #46.

Changed paths from absolute to relative for shorter error messages.
Additional side benefit is that all of our custom pydantic errors no longer need an extra `__init__` and are therefore shorter

### DIFF
--- a/nomenclature/region_mapping_models.py
+++ b/nomenclature/region_mapping_models.py
@@ -24,13 +24,6 @@ class RegionNameCollisionError(PydanticValueError):
     code = "region_name_collision"
     msg_template = "Name collision in {location} for {duplicates} in {file}"
 
-    def __init__(self, location: str, duplicates: Sequence, file: Path) -> None:
-        super().__init__(
-            location=location,
-            duplicates=duplicates,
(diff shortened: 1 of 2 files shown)


phackstock:
Rebased of the latest version of `main`.
Added `get_relative_path(path: Path)` to `region_mapping_models`. In case a relative path is given, it is returned unchanged. In case of an absolute path, the path relative to the current working directory is returned.
Added a unit test to check relative path inputs.

@danielhuppmann I'm not sure if `get_relative_path` needs to be it's own function as it is only one line and only used twice in the entire `region_mapping_models`. On the other hand it might make for slightly cleaner code.

danielhuppmann (approved):
Looks great, tested it locally with openentrance.
Thanks @phackstock!

phackstock:
Ah perfect.

PR #52, opened 2021-11-17
Change name of package

phackstock:
Changed the `metadata.name` in `setup.cfg` to `nomenclature-iamc`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [metadata]
-name = nomenclature
+name = nomenclature-iamc
 author = Daniel Huppmann
 author_email = huppmann@iiasa.ac.at
 license = APACHE-2.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,7 @@
 [metadata]
 name = nomenclature-iamc
 author = Scenario Services team, ECE program, IIASA
-author_email = huppmann@iiasa.ac.at
+author_email = ece-scse@iiasa.ac.at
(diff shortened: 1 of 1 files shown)


danielhuppmann:
I'm wondering if this is the correct way to do this...?

In my understanding, the ideal solution is that the package is still "called" `nomenclature` in Python, i.e., 
```python
Import nomenclature
```
but is released on pypi as `nomenclature-iamc`.

I looked at both **pyam** and **iam-units**, but they are not consistent in how they set metadata/name and options.packages relative to the package name and how it is distributed on pypi.

phackstock:
From my understanding this exactly what this accomplishes. `metadata.name` is the name for pypi while in python it will be name of the package directory so 'nomenclature'. I just did a test install on my system and I can still only import 'nomenclature'.

phackstock:
Plus during testing we install it as well and I would guess that we would get an error with broken imports if the import name changed.

danielhuppmann (approved):
I guess this is a good time to also change the authors to "Scenario Services team, ECE program, IIASA" - then good to merge and see if this plays nice with pypi!

phackstock:
Sounds good, I'll do that. I'll keep your email though, right?

danielhuppmann:
On June 7 in the weekly standup, we discussed having a generic scenario-services email to include ece & services & admin / support / helpdesk (or similar) - not sure if that was actually set up, but it would be a good to take that  here?

Issue #53, opened 2021-11-18
Start proper docs


danielhuppmann:
Set up a proper Sphinx-docs structure with the following sections:

- Overview and scope (pointing to openENTRANCE as the first use case, referencing pyam as the main dependency, with openENTRANCE and Horizon 2020 ackowledgement as in the current README)
- Usage (focus on the yaml format)
  - Code lists (region, variable with units, tags, generic)
  - Region mappings
- API documentation
- CLI documentation

danielhuppmann:
- @LauWien, could you take a first pass with setting up the Sphinx structure (using the same layout as pyam), and then take care of setting this up on ReadTheDocs?
- @phackstock, could you add the content of the usage?

LauWien:
>     * setting this up on ReadTheDocs?

Does IAMconsortium has an own RTD account so I can set it up? Right now I think I can only set up a documentation for RTD for all IIASA repositories and my own repositories, but not for this package.

PR #55, opened 2021-11-18
Initial structure of the documentation

LauWien:
This PR creates the initial structure of the documentation. It supports the closing of issue #53.


README.md
```diff
@@ -161,7 +161,7 @@ the list of countries included in a macro-region (i.e., a continent or large reg
 
 ## The pyam package
 
-<img src="./_static/pyam-logo.png" width="133" height="100" align="right" alt="pyam
+<img src="./doc/source/_static/pyam-logo.png" width="133" height="100" align="right" alt="pyam
```
danielhuppmann, line 164: In the latest pyam release, I included the logo directly in the pyam GitHub repo so that it could be retrieved directly (like here) and one wouldn't need to copy-paste it in every repo. See https://github.com/IAMconsortium/pyam/tree/main/doc/logos

doc/source/conf.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = "nomenclature"
+copyright = "2021, IAMconsortium"
+author = "IAMconsortium"
+
+# The full version, including alpha/beta/rc tags
+release = "0.1"
```
danielhuppmann, line 25: Please retrieve the version number automatically, see https://github.com/IAMconsortium/pyam/blob/20df347799bc0f8dc28bae9863ed71a7ae704267/doc/source/conf.py#L74

doc/source/conf.py
```diff
@@ -0,0 +1,67 @@
...
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = "nomenclature"
+copyright = "2021, IAMconsortium"
```
danielhuppmann, line 21: I'd prefer to use the ECE SceSe team as author and IIASA as copyright holder for the time being, see https://github.com/IAMconsortium/nomenclature/blob/621fd88b73c7e03c1d7129b8ef5664f03e588540/setup.cfg#L3

danielhuppmann (commented):
Thanks for the quick work @LauWien - a few suggestions inline, otherwise it looks great!

I guess it makes most sense that @phackstock reviews, approves, you merge - and then there is a follow-up PR to add content.

README.md
LauWien: Ah sorry, I think I had an old version of pyam. Will change it!

LauWien:
Interestingly if I want to satisfy stickler, I need to delete the whitespaces before ":". If I want to satisfy black I need to add them again. It's a vicious cycle :D.

danielhuppmann:
Not sure where that file is coming from, but I think that how this is now done in pyam is more elegant, see 
https://github.com/IAMconsortium/pyam/blob/20df347799bc0f8dc28bae9863ed71a7ae704267/pyam/__init__.py#L29

Also, if we specify that **nomenclature** can only be used with python >= 3.8, having `importlib_metadata` as an extra dependency (and the try-except in line 5 of that file) is not needed. And I think dropping python 3.7 here is reasonable.

nomenclature/__init__.py
```diff
@@ -8,6 +11,11 @@
     RegionAggregationMapping,
 )
 
-
 # set up logging
 logger = logging.getLogger(__name__)
+
+# get version number either from git (preferred) or metadata
+try:
+    __version__ = get_version(Path(__file__).parent.parent)
```
phackstock, line 19: Pathlib actually has a function to traverse the entire file path.
```suggestion
    __version__ = get_version(Path(__file__).parents[1])
```
Should be the correct one.

nomenclature/__init__.py
```diff
@@ -1,4 +1,7 @@
 import logging
+from pathlib import Path
+from setuptools_scm import get_version
+from importlib.metadata import version
```
phackstock, line 4: If we require `importlib.metadata` we should add the fact that this package requires python 3.8 or higher to setup.cfg and the README.md

phackstock (commented):
Thanks @LauWien for getting sphinx started.
Some minor comments below.

LauWien:
Right now the sidebar is too small for the name "nomenclature" I think we should increase the width once we implement content.

phackstock (approved):
Looks good to me.
Thanks, and good point about adjusting the layout once we have content.

danielhuppmann (approved):
Looks great, many thanks! Could we just implement the change to requiring Python >=3.8 also on the badge and the text of the readme? Then, good to merge!

PR #59, opened 2021-11-19
Change package name when reading version from metadata

danielhuppmann:
Unit tests on https://github.com/openENTRANCE/openentrance fail, see https://github.com/openENTRANCE/openentrance/runs/4265491713?check_suite_focus=true - this is because the package is now called "nomenclature-iamc" in the metadata of the Python installation.

### DIFF
--- a/nomenclature/__init__.py
+++ b/nomenclature/__init__.py
@@ -18,4 +18,4 @@ logger = logging.getLogger(__name__)
 try:
     __version__ = get_version(Path(__file__).parents[1])
 except LookupError:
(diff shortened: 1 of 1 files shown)


danielhuppmann:
fyi @LauWien @phackstock

danielhuppmann:
Merging this without review as a quick bugfix.

danielhuppmann:
Proof of the pudding is in the testing - GitHub Actions now pass, see https://github.com/openENTRANCE/openentrance/actions/runs/1481459685

PR #60, opened 2021-11-22
Add readthedocs.yml for documentation build

LauWien:
This PR adds the _readthedocs.yml_ file for the documentation build on [RTD](https://readthedocs.org/). It supports the closing of issue #53.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,0 +1,29 @@
+# This workflow installs the package and builds the docs
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: build the docs
+
+on:
+  push:
+    branches: [ 'main' ]
+  pull_request:
+    branches: [ '**' ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Python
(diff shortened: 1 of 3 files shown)


danielhuppmann:
Looks good! But how about adding a build-the-docs GitHub Actions workflow, like https://github.com/IAMconsortium/pyam/blob/main/.github/workflows/build-docs.yml?

danielhuppmann (approved):
Looks good to me!

phackstock:
Just for understanding, this build-docs.yml is only for assuring that the docs can be built without any errors right? It does not do anything with the docs once they are built, correct?

danielhuppmann:
Correct, @phackstock:
 - RTD only sphinx-builds the docs **after** a PR is merged and does not raise an error if a build fails.
 - The additional GitHub Action workflow ensures that the sphinx-build succeeds **before** a PR is merged.

Technically speaking, because GitHub Action uses a slightly configuration from RTD, this will not cause all problems - but it saved me from a few silly mistakes in the past. What I did in the past if I really wanted to be sure about what will happen on RTD: create a branch on upstream (rather than my fork) and to add the branch to the list of branches built by RTD before starting a PR. See a real-world example here: https://github.com/IAMconsortium/pyam/pull/564 - though, usually, I then remove the branch from RTD again after merging the PR.

LauWien:
Thank you @danielhuppmann for reviewing!

PR #62, opened 2021-11-23
Add badges and reference to openENTRANCE to the docs

danielhuppmann:
While @phackstock is busy preparing the ENGAGE dataset, I did a little work on the docs page, contributing to the closing of #53


doc/source/index.rst
```diff
@@ -1,20 +1,59 @@
-.. nomenclature documentation master file, created by
-   sphinx-quickstart on Thu Nov 18 09:33:32 2021.
-   You can adapt this file completely to your liking, but it should at least
-   contain the root `toctree` directive.
+**nomenclature**: Working with IAMC-style scenario data
+=======================================================
 
-Welcome to nomenclature's documentation!
-========================================
+Release v\ |version|.
```
LauWien, line 4: This line shows _Release v0.2.dev12+g880ca57._. Not sure if that is intended or if I did something wrong in the implementation of the dynamic version tracking.

LauWien (approved):
Looks great! Only one question inline.
Not sure if it's an issue, but the side bar/layout changes when I click _The DataStructureDefinition class_, I guess due to the long name. Should we change the width of the side bar or is it fine like it is for now?

doc/source/index.rst
danielhuppmann: This is expected behavior: Sphinx bumps the smallest (in the sense major -> minor -> patch) part of latest version number by 1 and adds `.dev`, then the number of commits since the latest release, and then the short hash of the latest commit. If there are uncommitted changes, it also adds the term `dirty`.

LauWien: Thanks for the explanation!

danielhuppmann:
Thanks for noticing the bad layout, I added some fix, and we can play around with this as the page develops further...

LauWien:
Looks great now on RTD :)

PR #63, opened 2021-11-25
Add a Sphinx-docs page for the CLI

danielhuppmann:
This PR adds an automated CLI-documentation page (using **sphinx-click**, see https://sphinx-click.readthedocs.io).

In anticipation of future CLI methods that are not part of the testing module, I also moved the CLI-part to an independent CLI module.

Contributes to the closing of #53

### DIFF
--- a/nomenclature/__init__.py
+++ b/nomenclature/__init__.py
@@ -5,7 +5,7 @@ from importlib.metadata import version
 
 from nomenclature.codes import CodeList  # noqa
 from nomenclature.core import DataStructureDefinition, create_yaml_from_xlsx  # noqa
-from nomenclature.testing import assert_valid_yaml  # noqa
+from nomenclature.cli import cli  # noqa
 from nomenclature.region_mapping_models import (  # noqa
     RegionProcessor,
     RegionAggregationMapping,
--- a/doc/source/cli.rst
+++ b/doc/source/cli.rst
@@ -1,0 +1,6 @@
+Command line interface
+======================
+
+.. click:: nomenclature:cli
+   :prog: nomenclature
+   :nested: full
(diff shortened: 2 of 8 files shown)


nomenclature/__init__.py
```diff
@@ -5,7 +5,7 @@
 
 from nomenclature.codes import CodeList  # noqa
 from nomenclature.core import DataStructureDefinition, create_yaml_from_xlsx  # noqa
-from nomenclature.testing import assert_valid_yaml  # noqa
+from nomenclature.cli import cli  # noqa
```
phackstock, line 8: Maybe we want to additionally expose `assert_valid_yaml` and `assert_valid_structure` to the user.
`from nomenclature.testing import assert_valid_yaml, assert_valid_structure`

phackstock (approved):
Looks great. Very nice that the cli now lives in its own file. I found myself also splitting modules up into more logical chunks when I started preparing more documentation.
Just one small comment.

LauWien (approved):
Looks good :)

nomenclature/__init__.py
danielhuppmann: I actually wanted to align the behavior more closely with pandas and numpy (i.e., the most sophisticated packages I know). There, the testing-methods like `assert_frame_equal` are not exposed at the top level, but can be imported from the module, see for example 

https://github.com/IAMconsortium/pyam/blob/732752c9c1af9ab453b28584a4a4e05cc9930c0a/tests/test_core.py#L110

Equivalently, with the new structure, a user can still do
```python
from nomenclature.testing import assert_valid_structure
```

phackstock: Ok, following numpy and pandas sounds like a very sensible choice.

PR #64, opened 2021-11-26
Feature/enhance documentation

### LINKED ISSUE #53: Start proper docs

phackstock:
closes #53.

This PR adds the following:
* Install section in the docs
* Getting started section in the docs
* Moved the region processing errors to their own module
* Moved `get_relative_path` from `region_mapping_models` to a newly created `utils` module.
* Added the region mapping models to the API documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,8 @@ projects.
 
 ## Getting started
 
-To install this package, please install Python version 3.8 or higher. Nomenclature is on pypi so it can be installed like this:
+To install this package, please install Python version 3.8 or higher. Nomenclature is on
+pypi so it can be installed like this:
 
 ```bash
 pip install nomenclature-iamc
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ projects.
 ## Getting started
 
 To install this package, please install Python version 3.8 or higher. Nomenclature is on
-pypi so it can be installed like this:
+To install the latest release of the package, please use the following command:
 
 ```bash
 pip install nomenclature-iamc
--- a/README.md
+++ b/README.md
@@ -20,151 +20,13 @@ It supports validation of scenario data and region processing, which consists of
 renaming and aggregation of model "native regions" to "common regions" used in a
 project. It requires Python version >= 3.8.
 
-A **DataStructureDefinition** class contains **CodeLists** for *variables* (including
-units) and *regions* to be used in a model comparison or scenario exercise following the
-IAMC data format.
-
-A **CodeList** is a list of "allowed terms" (or codes), where each term can have several
-attributes (e.g., description, unit, parent region).
-
-A **RegionAggregationMapping** is a mapping that defines on a per-model basis how model
-native regions should be renamed and aggregated to comparison regions.
-
-A **RegionProcessor** is a class that holds a list of RegionAggregationMappings and a
-DataStructureDefinition. This class is used to facilitate region processing for model
-comparison studies.
-
-## The structure of a datastructure definition
-
-A **DataStructureDefinition** is initialized from a folder with the following structure:
-
-### Variables
-
-The *variable* codelist of the **DataStructureDefinition** will be read from all yaml
-files located  in a folder of that name (including any sub-folders). They must be
-formatted as a list of dictionaries mapping the variable (key) to its attributes.
-
-```yaml
-- Some Variable:
-    description: A short description
-    unit: A unit
-    <other attribute>: Some text (optional)
-```
-
-Every variable must have a **unit**, which should be compatible with the Python package
-[iam-units](https://github.com/iamconsortium/units). The unit attribute can be empty,
-i.e., the variable is *dimensionless*.
-
-#### Tags
-
-To avoid repetition (and subsequent errors), any number of yaml files can be used as
-"tags" using a nested list of dictionaries. The files defining the tags must have a name
-starting with `tag_`.
-
-```yaml
-- <Tag>:
-  - Some Key:
-      description: a short description of the key
-```
-
-When importing the codelist, any occurrence of `<Tag>` in a variable name will be
-replaced by every element in the Tag dictionary. The `<Tag>` will also be replaced in
-any of the variable attributes.
-
-There must be only one top-level entry in any yaml file to be used as tag.
-
-## Model Mappings
-
-Model mappings, defined in a .yaml format serve three different purposes on a per-model
-basis:
-
-1. Define a list of model native regions that are to be selected (and usually uploaded)
-   from an IAM result. This also serves as an implicit exclusion list for model native
-   regions, since only explicitly mentioned regions are selected.
-
-2. Allow for renaming of model native regions.
-
-3. Define how model native regions should be aggregated to common regions.
-
-This example illustrates how such a model mapping looks like:
-
-```yaml
-model: model_a
-native_regions:
-  - region_a: alternative_name_a
-  - region_b
-common_regions:
-  - common_region_1:
-    - region_a
-    - region_b
-  - common_region_2:
-    - ...
-```
-
-### Notes
-
-* The names of the three top level keywords `model`, `native_regions` and
-  `common_regions` are fixed.
-* Required properties are `model` and **at least** either `native_regions` or
-  `common_regions`. **Both** are **allowed** as well.
-* **`model`** (str): specifies the model name for which the mapping applies.
-* **`native_regions`** (list): list of model native regions serves as a selection as to
-  which regions to keep.
-  * In the above example `region_a` is to be renamed to `alternative_name_a`. This is
-    done by defining a key-value pair of `model_native_name: new_name`.
-  * `region_b` is selected but the name is not changed.
-  * Assuming `model_a` also defines a third region `region_c`, since it is not mentioned
-    it will be **dropped** from the data.
-* **`common_regions`** (list): list of common regions which will be computed as
-  aggregates. They are defined as list entries which themselves have a list of
-  constituent regions. These constituent regions must be model native regions.
-  * **Important to note** the names of the constituent regions **must** refer to the
-    **original** model native region names. In the above example `region_a` and
-    `region_b` and **not** `alternative_name_a`.
-
-#### Guidelines and naming convention
-
-The variable name (code) should adhere to the following conventions:
-
-- A `|` (pipe) character indicates levels of hierarchy
-- Do not use spaces before and after the `|` character, but add a space between words
-  (e.g., `Primary Energy|Non-Biomass Renewables`)
-- All words must be capitalised (except for 'and', 'w/', 'w/o', etc.)
-- Do not use abbreviations (e.g, 'PHEV') unless strictly necessary
-- Add hierarchy levels where it might be useful in the future, e.g., use `Electric
-  Vehicle|Plugin-Hybrid` instead of 'Plugin-Hybrid Electric Vehicle'
-- Do not use abbreviations of statistical operations ('min', 'max', 'avg') but always
-  spell out the word
-- Do not include words like 'Level' or 'Quantity' in the variable, because this should
-  be clear from the context or unit
-
-### Regions
-
-The *region* codelist of the nomenclature will be read from all yaml files located in a
-folder of that name (including any sub-folders). To avoid repeating a "hierarchy"
-attribute many times (e.g., country, continent), the yaml files must have a nested
-dictionary structure:
-
-```yaml
-- <Hierarchy Level>:
-  - Region Name:
-      Attribute: Attribute value
-```
-
-When importing the codelist, the hierarchy will be added as attribute, such that it can
-be retrieved as
-
-```python
-DataStructureDefinition.region["Region Name"]["Hierarchy"] = "<Hierarchy Level>"
-```
-
-Other attributes specified in the yaml file can include (for countries) ISO2/3-codes, or
-the list of countries included in a macro-region (i.e., a continent or large region).
+The full documentation is hosted on [Read the
+Docs](https://nomenclature-iamc.readthedocs.io/)
 
 ## The pyam package
 
-<img src="https://github.com/IAMconsortium/pyam/blob/main/doc/logos/pyam-logo.png" width="133" height="100" align="right" alt="pyam
-logo" />
+<img src="https://github.com/IAMconsortium/pyam/blob/main/doc/logos/pyam-logo.png"
+width="133" height="100" align="right" alt="pyam logo" />
 
 This package is intended to complement the Python package **pyam**, an open-source
 community toolbox for analysis & visualization of scenario data. That package was
@@ -177,11 +39,16 @@ projects.
 
 ## Getting started
 
-To install this package, please install Python version 3.7 or higher. Then, download or
-git-clone this repository and run the following command in the root folder:
+To install this package, please install Python version 3.8 or higher. Nomenclature is on pypi so it can be installed like this:
 
+```bash
+pip install nomenclature-iamc
 ```
-pip install --editable .
+
+Alternatively, it can also be installed directly from source:
+
+```bash
+pip install -e git+https://github.com/IAMconsortium/nomenclature#egg=nomenclature
 ```
 
 Then, open a Python console and import a suitable nomenclature structure from a folder
@@ -195,8 +62,8 @@ project.variable
 
 ## Acknowledgement
 
-<img src="./doc/source/_static/open_entrance-logo.png" width="202" height="129" align="right"
-alt="openENTRANCE logo" />
+<img src="./doc/source/_static/open_entrance-logo.png" width="202" height="129"
+align="right" alt="openENTRANCE logo" />
 
 This package is based on the work initially done in the [Horizon 2020
 openENTRANCE](https://openentrance.eu) project, which aims to  develop, use and
@@ -206,6 +73,6 @@ low-carbon transition pathways in Europe.
 Refer to the [openENTRANCE/nomenclature](https://github.com/openENTRANCE/nomenclature)
 repository for more information.
 
-<img src="./doc/source/_static/EU-logo-300x201.jpg" width="80" height="54" align="left" alt="EU
-logo" /> This project has received funding from the European Union’s Horizon 2020
-research and innovation programme under grant agreement No. 835896.
+<img src="./doc/source/_static/EU-logo-300x201.jpg" width="80" height="54" align="left"
+alt="EU logo" /> This project has received funding from the European Union’s Horizon
+2020 research and innovation programme under grant agreement No. 835896.
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -14,7 +14,7 @@ DataStructureDefinition
 
 
 
-Region processing
+RegionProcessor 
 -----------------
 
 .. automodule:: nomenclature.region_mapping_models
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -8,3 +8,9 @@ DataStructureDefinition
 
 .. autoclass:: DataStructureDefinition
    :members:
+
+Region processing
+-----------------
+
+.. automodule:: nomenclature.region_mapping_models
+   :members:
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -12,9 +12,7 @@ DataStructureDefinition
    :members:
 
 
-
-
-RegionProcessor 
+Region processing 
 -----------------
 
 .. automodule:: nomenclature.region_mapping_models
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -1,3 +1,5 @@
+.. _api:
+
 .. currentmodule:: nomenclature
 
 API documentation
@@ -8,6 +10,9 @@ DataStructureDefinition
 
 .. autoclass:: DataStructureDefinition
    :members:
+
+
+
 
 Region processing
 -----------------
--- a/doc/source/getting_started.rst
+++ b/doc/source/getting_started.rst
@@ -51,7 +51,7 @@ The following outlines how to use the nomenclature package:
    dsd = DataStructureDefinition("definitions/")
    rp = RegionProcessor.from_directory("mappings/", dsd)
 
-   # Read in the data using pyam
+   # Read the data using pyam
    iam_results_file = "some file"
    df = pyam.IamDataFrame(iam_results_file)
 
@@ -68,7 +68,7 @@ The following outlines how to use the nomenclature package:
 * *DataStructureDefinition* and *RegionProcessor* are initialized from directories
   containing yaml files. See :ref:`dir-structure` for details. 
 
-* *DataStructureDefinition.apply()* returns none if the data frame only contains   
+* *DataStructureDefinition.apply()* returns *None* if the data frame only contains   
   allowed values and raises an error otherwise.
 
 
--- a/doc/source/getting_started.rst
+++ b/doc/source/getting_started.rst
@@ -19,16 +19,64 @@ A **DataStructureDefinition** contains **CodeLists** for *variables* (including 
 and *regions* to be used in a model comparison or scenario exercise following the IAMC
 data format.
 
-A **CodeList** is a list of "allowed terms" (or codes), where each term can have several
-attributes (e.g., description, unit, parent region).
-
 A **RegionProcessor** holds a list of RegionAggregationMappings and a
 DataStructureDefinition. This class is used to facilitate region processing for model
 comparison studies.
 
+A **CodeList** is a list of "allowed terms" (or codes), where each term can have several
+attributes (e.g., description, unit, parent region).
+
 A **RegionAggregationMapping** is a mapping that defines on a per-model basis how model
 native regions should be renamed and aggregated to comparison regions.
 
+
+Before the :ref:`minimum-working-example` can be covered, the required directory
+structure for definitions and mappings needs to be addressed. 
+
+.. _dir-structure:
+
+Directory structure for definitions and mappings
+------------------------------------------------
+
+This is the directory structure that needs to be in place in order for the validation and region processing to work:
+
+.. code-block:: bash
+
+   .
+   ├── definitions
+   │   ├── region
+   │   │   ├── ...
+   │   │   └── regions.yaml
+   │   └── variable
+   │       ├── ...
+   │       └── variable.yaml
+   └── mappings [optional]
+       ├── model_a.yaml
+       └── ...
+
+**Notes**
+
+* The **DataStructureDefinition** will be initialized from the *definitions* folder.
+
+* The **RegionProcessor** will be initialized from the *mappings* folder. If the project
+  has no model specific mappings, this folder can also be omitted. In this case,
+  however, *RegionProcessor* **must not** be used as it would try to read a non-existent
+  directory causing the program to crash.
+
+* Inside the *definitions* directory each "dimension", in our case *variable* and
+  *region*, must live in its own sub-directory.
+
+* The directories inside the *definitions* folder, *variable* and *region* are special
+  names that must be kept.
+
+* The definitions can be spread across multiple yaml files. In the interest of keeping
+  this example minimal only one file for regions and variables is shown.
+
+* The *mappings* directory directly contains the model mappings. There are no special
+  sub-folders required. 
+
+
+.. _minimum-working-example:
 
 Minimum working example
 -----------------------
@@ -49,6 +97,7 @@ The following outlines how to use the nomenclature package:
    # Initialize DataStructureDefinition and RegionProcessor giving them the
    # directories where the codelists and mappings are defined as input.
    dsd = DataStructureDefinition("definitions/")
+   # Only to be used if there are mappings!
    rp = RegionProcessor.from_directory("mappings/", dsd)
 
    # Read the data using pyam
@@ -58,6 +107,7 @@ The following outlines how to use the nomenclature package:
    # Validate that the data frame only contains allowed regions and variables
    dsd.validate(df)
    # Apply region processing to the data
+   # Only to be used if there are mappings!
    df = rp.apply(df)
 
 **Notes**
@@ -71,43 +121,4 @@ The following outlines how to use the nomenclature package:
 * *DataStructureDefinition.apply()* returns *None* if the data frame only contains   
   allowed values and raises an error otherwise.
 
-
-.. _dir-structure:
-
-Directory structure for definitions and mappings
-------------------------------------------------
-
-This is the directory structure that needs to be in place in order for the validation and region processing to work:
-
-.. code-block:: bash
-
-   .
-   ├── definitions
-   │   ├── region
-   │   │   ├── ...
-   │   │   └── regions.yaml
-   │   └── variable
-   │       ├── ...
-   │       └── variable.yaml
-   └── mappings
-       ├── model_a.yaml
-       └── ...
-
-**Notes**
-
-* The names of the *definitions* and *mappings* directories are purely conventional and
-  don't carry any special meaning. As long as *DataStructureDefinition* and
-  *RegionProcessor.from_directory()* are pointed to the correct directories everything
-  will work.
-
-* Inside the *definitions* directory each "dimension", in our case *variable* and
-  *region*, must live in its own sub-directory.
-
-* As opposed *definitions* and *mappings*, *variable* and *region* are special names
-  that must be kept.
-
-* The definitions can be spread across multiple yaml files. In the interest of keeping
-  this example minimal only one file for regions and variables is shown.
-
-* The *mappings* directory directly contains the model mappings. There are no special
-  sub-folders required. 
+* *RegionProcessor* is only to be used if there are model mappings.
--- a/doc/source/getting_started.rst
+++ b/doc/source/getting_started.rst
@@ -51,8 +51,33 @@ Every variable must have a **unit**, which should be compatible with the
 Python package `iam-units <https://github.com/iamconsortium/units>`__.
 The unit attribute can be empty, i.e., the variable is *dimensionless*.
 
+Regions
+~~~~~~~
+
+The *region* codelist of the nomenclature will be read from all yaml
+files located in a folder of that name (including any sub-folders). To
+avoid repeating a “hierarchy” attribute many times (e.g., country,
+continent), the yaml files must have a nested dictionary structure:
+
+.. code:: yaml
+
+   - <Hierarchy Level>:
+     - Region Name:
+         Attribute: Attribute value
+
+When importing the codelist, the hierarchy will be added as attribute,
+such that it can be retrieved as
+
+.. code:: python
+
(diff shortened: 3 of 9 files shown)


doc/source/install.rst
```diff
@@ -0,0 +1,25 @@
...
+
+
+Via Pip
+-------
+
+.. code-block:: bash
+
+    pip install nomenclature-iamc
+
+.. attention::  The nomenclature package is distributed as "nomenclature-iamc" on pypi.
```
danielhuppmann, line 14: Suggest to move this ahead of the pip-install block...

nomenclature/region_mapping_models.py
```diff
@@ -24,35 +29,69 @@
...
+
     name: str
     constituent_regions: List[str]
 
 
 class RegionAggregationMapping(BaseModel):
+    """Holds information for region processing on a per-model basis.
+
+    Region processing is comprised of native region selection and potentiallay renaming
+    as well as common region aggregation.
```
danielhuppmann, line 81:
```suggestion
    as well as aggregation to "common regions" (regions used for reporting and comparison by multiple models).
```

nomenclature/region_mapping_models.py
```diff
@@ -115,7 +154,30 @@ def check_illegal_renaming(cls, values):
         return values
 
     @classmethod
-    def create_from_region_mapping(cls, file: Union[Path, str]):
+    def from_file(cls, file: Union[Path, str]):
+        """Effectively the constructor for RegionAggregationMapping.
```
danielhuppmann, line 158:
```suggestion
        """Initialize a RegionAggregationMapping from a file.
```

nomenclature/region_mapping_models.py
```diff
@@ -115,7 +154,30 @@ def check_illegal_renaming(cls, values):
...
+
+        Raises
+        ------
+        ValidationError
+            Raised in case there are any errors in the provided yaml mapping file.
+
+        Notes
+        -----
+
+        The reason that this is effectively the constructor is that the schema for region aggregation mapping yaml files is
```
danielhuppmann, line 179: This is not a complete or intellegible sentence...

nomenclature/region_mapping_models.py
```diff
@@ -203,9 +251,28 @@ class Config:
     @classmethod
     @validate_arguments(config=dict(arbitrary_types_allowed=True))
     def from_directory(cls, path: DirectoryPath, definition: DataStructureDefinition):
+        """Effectively the constructor of RegionProcessor
```
danielhuppmann, line 254:
```suggestion
        """Initialize a RegionProcessor from a directory of model-aggregation mappings.
```

doc/source/getting_started.rst
```diff
@@ -0,0 +1,190 @@
...
+      ``alternative_name_a``.
+
+Guidelines and naming convention
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The variable name (code) should adhere to the following conventions:
+
+-  A ``|`` (pipe) character indicates levels of hierarchy
+-  Do not use spaces before and after the ``|`` character, but add a
+   space between words (e.g., ``Primary Energy|Non-Biomass Renewables``)
```
danielhuppmann, line 142: Suggest to use parentheses and italics for variable names in any documentation (instead of `code`).

doc/source/getting_started.rst
```diff
@@ -0,0 +1,190 @@
...
+
+.. code:: python
+
+   DataStructureDefinition.region["Region Name"]["Hierarchy"] = "<Hierarchy Level>"
+
+Other attributes specified in the yaml file can include (for countries)
+ISO2/3-codes, or the list of countries included in a macro-region (i.e.,
+a continent or large region).
+
+The pyam package
```
danielhuppmann, line 178: Don't need to include the advertisement section for pyam here...

doc/source/getting_started.rst
```diff
@@ -0,0 +1,190 @@
...
+
+Notes
+~~~~~
+
+-  The names of the three top level keywords ``model``,
+   ``native_regions`` and ``common_regions`` are fixed.
+-  Required properties are ``model`` and **at least** either
+   ``native_regions`` or ``common_regions``. **Both** are **allowed** as
+   well.
+-  **``model``** (str): specifies the model name for which the mapping
```
danielhuppmann, line 113: Using both bold and codes doesn't work...

doc/source/getting_started.rst
```diff
@@ -0,0 +1,190 @@
...
+     - common_region_1:
+       - region_a
+       - region_b
+     - common_region_2:
+       - ...
+
+Notes
+~~~~~
+
+-  The names of the three top level keywords ``model``,
```
danielhuppmann, line 108: Using **bold** and `code` styling so often looks very busy, maybe bold for the expected/allowed keywords and italics for referring to values in the example (*region_a*)

doc/source/getting_started.rst
```diff
@@ -0,0 +1,190 @@
...
+   computed as aggregates. They are defined as list entries which
+   themselves have a list of constituent regions. These constituent
+   regions must be model native regions.
+
+   -  **Important to note** the names of the constituent regions
+      **must** refer to the **original** model native region names. In
+      the above example ``region_a`` and ``region_b`` and **not**
+      ``alternative_name_a``.
+
+Guidelines and naming convention
```
danielhuppmann, line 135:
```suggestion
Guidelines and variable naming conventions
```

danielhuppmann (commented):
Thanks @phackstock, a step in a good direction! A few suggestions to make the formatting easier to read and other minor issues...

doc/source/getting_started.rst
```diff
@@ -0,0 +1,190 @@
...
+
+A **RegionAggregationMapping** is a mapping that defines on a per-model
+basis how model native regions should be renamed and aggregated to
+comparison regions.
+
+A **RegionProcessor** is a class that holds a list of
+RegionAggregationMappings and a DataStructureDefinition. This class is
+used to facilitate region processing for model comparison studies.
+
+The structure of a datastructure definition
```
LauWien, line 29:
```suggestion
The structure of a DataStructureDefinition
```

doc/source/getting_started.rst
```diff
@@ -0,0 +1,190 @@
...
+         description: a short description of the key
+
+When importing the codelist, any occurrence of ``<Tag>`` in a variable
+name will be replaced by every element in the Tag dictionary. The
+``<Tag>`` will also be replaced in any of the variable attributes.
+
+There must be only one top-level entry in any yaml file to be used as
+tag.
+
+Model Mappings
```
LauWien, line 74:
```suggestion
Model mappings
```
or RegionAggregationMapping (?)

doc/source/getting_started.rst
```diff
@@ -0,0 +1,190 @@
...
+comparison or scenario exercise following the IAMC data format.
+
+A **CodeList** is a list of “allowed terms” (or codes), where each term
+can have several attributes (e.g., description, unit, parent region).
+
+A **RegionAggregationMapping** is a mapping that defines on a per-model
+basis how model native regions should be renamed and aggregated to
+comparison regions.
+
+A **RegionProcessor** is a class that holds a list of
```
LauWien, line 25: If the class "DataStructureDefinition" and "RegionAggregationMapping" gets explained in more detail, doesn't it make also sense to add the "RegionProcessor" class? If so I would drop "CodeLists" and move it under the "DataStructureDefinition" class. Then it maybe make sense to only use the classes as headings and skip the "Overview" heading and "The pyam package". Or maybe I just don't understand the structure of this page :)

doc/source/install.rst
```diff
@@ -0,0 +1,25 @@
...
+
+Installing From Source
+----------------------
+
+nomenclature can also be installed from source. This will get you the latest version
+of the main branch.
+
+.. code-block:: bash
+
+    pip install -e git+https://github.com/IAMconsortium/nomenclature#egg=nomenclature
```
LauWien, line 24: I this correct, or is it `pip install git+https://github.com/IAMconsortium/nomenclature.git@main`?

nomenclature/region_mapping_models.py
```diff
@@ -24,35 +29,69 @@
...
+    """
+
     name: str
     constituent_regions: List[str]
 
 
 class RegionAggregationMapping(BaseModel):
+    """Holds information for region processing on a per-model basis.
+
+    Region processing is comprised of native region selection and potentiallay renaming
```
LauWien, line 80:
```suggestion
    Region processing is comprised of native region selection and potentially renaming
```

nomenclature/region_mapping_models.py
```diff
@@ -115,7 +154,30 @@ def check_illegal_renaming(cls, values):
         return values
 
     @classmethod
-    def create_from_region_mapping(cls, file: Union[Path, str]):
+    def from_file(cls, file: Union[Path, str]):
+        """Effectively the constructor for RegionAggregationMapping.
+
+        Parameters
+        ----------
+        file : Union[Path, str]
```
LauWien, line 162: The layout is different to other modules e.g. _core.py_.

nomenclature/utils.py
```diff
@@ -0,0 +1,7 @@
+from pathlib import Path
+
+
+def get_relative_path(path: Path):
+    # Get the relative version of `path` relative to `path_relative_to`
+    # In case path does not contain path_relative_to it is returned unchanged
```
LauWien, line 6:
```suggestion
    # In case path does not contain `path_relative_to` it is returned unchanged
```

LauWien (commented):
Looks good! Just a few comments in line.
Looking ad the index.html - the layout is kind of changing when I click through the side bar.

README.md
```diff
@@ -177,11 +39,17 @@ projects.
 
 ## Getting started
 
-To install this package, please install Python version 3.7 or higher. Then, download or
-git-clone this repository and run the following command in the root folder:
+To install this package, please install Python version 3.8 or higher. Nomenclature is on
```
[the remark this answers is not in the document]
phackstock: Why would you remove the python version requirement here?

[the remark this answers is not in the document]
LauWien: Oh sorry, I thought it is already in the README.md, my bet!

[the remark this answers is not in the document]
phackstock: It is in fact in the README.md but I don't think that's a bad thing per se. Having redundancy for a critical piece of information such as the python version is a good thing in my opinion as the chance is higher that people will read it.
It of course brings the downside that once that changes we need to update it in multiple places.

doc/source/install.rst
phackstock: Very good question, I am afraid I can only give you what I understand of it which is what I gathered from [this stackoverflow post](https://stackoverflow.com/questions/62841000/how-is-egg-used-in-pip-install-e). 
Basically, if you run `pip install` from a source other than pypi you need to specify the egg option. Apparently this is required otherwise you get an error. I tried this on my machine and can confirm. Interestingly enough it seems that the actual value of the egg parameter does not matter. `pip install -e git+https://github.com/IAMconsortium/nomenclature.git@main#egg=asdf` worked perfectly fine for me. 
To actually answer your question, the `@main` addition is perfectly valid and can be combined with the egg option like so: `pip install -e git+https://github.com/IAMconsortium/nomenclature.git@main#egg=some-name-for-the-egg`. In our case where we install the latest version of the main branch. As this is the default for pip running it with or without the `@main` makes no difference.
This of course raises the question why there is that option in the first place and it is to specify a particular version, so if I actually want to install version 0.1 I would put `@v0.1`.
Hope that clears things up.
I the interest of being precise I'll include the `@main` as it does not rely on perhaps non-obvious default values.

nomenclature/region_mapping_models.py
phackstock: What do you mean by that?

doc/source/install.rst
phackstock: Done

doc/source/getting_started.rst
phackstock: Done.

doc/source/getting_started.rst
phackstock: Changed to just bold.

doc/source/getting_started.rst
phackstock: Done.

doc/source/getting_started.rst
phackstock: Do you mean variable names in the IAMC sense like 'Primay Energy' here or in the python sense, or both?
How should I use parentheses, like so (*Primary Energy*)?

doc/source/getting_started.rst
phackstock: I'm not sure I fully understand what you mean.
Do you mean adding a dedicated section for the RegionProcessor class? That would probably be a good idea.
I would vote for keeping the overview as it serves both as on overview and a glossary to explain the terms used in the following to the user.

doc/source/getting_started.rst
```diff
@@ -0,0 +1,190 @@
...
+       description: A short description
+       unit: A unit
+       <other attribute>: Some text (optional)
+
+Every variable must have a **unit**, which should be compatible with the
+Python package `iam-units <https://github.com/iamconsortium/units>`__.
+The unit attribute can be empty, i.e., the variable is *dimensionless*.
+
+Tags
+^^^^
```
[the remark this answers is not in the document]
phackstock: Good question I think it should not be there as its application is in principle not limited to variables but could also be extended to regions. Do I understand that correctly @danielhuppmann?
As regions are also part of a DataStructureDefinition I would say it makes sense to move the regions section also under the DataStructureDefinition.
I'm afraid I don't understand what you mean by:

> to many levels under "The structure of a datastructure definition" with to less headings, I guess.

phackstock:
Addressed all the non-resolved comments.
@LauWien Do you mean that the contents of the navbar change and only include the "Getting started" section if we are on index or on the actual getting started page? If so I have also noticed that, no idea what causes it but in the latest builds on my system the problem is gone. No idea what fixed it though.

doc/source/getting_started.rst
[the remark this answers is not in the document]
LauWien: Sorry for being not precise. I meant having, one heading with one subheading with one sub-subheading (heading 1, 2, 3). I just think it might look better to have at least two headings on each level (2, 3). I hope this is better understandable :sweat_smile:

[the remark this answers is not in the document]
phackstock: Ah right, ok, good point. If we move regions under DataStructureDefinition and move the Tags up one level we should have three subheadings.

doc/source/getting_started.rst
LauWien: Yes, I meant having for example the following structure analogue to the API documentation:

## Getting started
This package facilitates working with data templates that follow the format developed by the Integrated Assessment Modeling Consortium (IAMC). It supports validation of scenario data and region processing, which consists of renaming and aggregation of model “native regions” to “common regions” used in a project.

### DataStructureDefinition 

A DataStructureDefinition class contains CodeLists for variables (including units) and regions to be used in a model comparison or scenario exercise following the IAMC data format.
A CodeList is a list of “allowed terms” (or codes), where each term can have several attributes (e.g., description, unit, parent region).
...

### RegionAggregationMapping 
...

### RegionProcessor 
...

But as I said maybe I don't understand why the page is structured as it is :)

doc/source/install.rst
LauWien: Ah got it!

doc/source/getting_started.rst
danielhuppmann: Variable in the IAMC-sense

nomenclature/region_mapping_models.py
LauWien: Ah sorry, all good, I checked wrong.

doc/source/getting_started.rst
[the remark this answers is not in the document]
danielhuppmann: Tags should work for all types of codelists, but it is only explicitly tested for variables... Suggest to follow a strategy of "assume it works until proven otherwise, then fix and add tests"...

doc/source/getting_started.rst
danielhuppmann: This page is getting quite long for a "getting started" page... I'm afraid that this might scare off potential users.

LauWien:
> Addressed all the non-resolved comments. @LauWien Do you mean that the contents of the navbar change and only include the "Getting started" section if we are on index or on the actual getting started page? If so I have also noticed that, no idea what causes it but in the latest builds on my system the problem is gone. No idea what fixed it though.

I meant if I click for example on the navigation bar on "Installation" the zoom of my local build of the page is greater than if I click "Getting started" and if I click "Via Pip" it changes again. Is it the same for you or is it a problem on my side?

doc/source/install.rst
danielhuppmann: The proof of the pip is in the GitHub actions.

In the NetZero2040 project (the next customer for this package, started yesterday), the `requirements.txt` lists
```
git+git://github.com/IAMconsortium/nomenclature.git
```
see [here](https://github.com/netzero2040/netzero2040/blob/5482c3f7bd6456d594aac5f2f10eea433814cb1e/requirements.txt#L3)

and the GitHub Action workflow uses
```
 run: pip install -r requirements.txt
```

see [here](https://github.com/netzero2040/netzero2040/blob/5482c3f7bd6456d594aac5f2f10eea433814cb1e/.github/workflows/validation.yml#L26)

phackstock:
@LauWien I took a look at my local build and I don't see any changes in zoom on my side. I'm viewing it in firefox if that information helps

doc/source/getting_started.rst
phackstock: Alright, updated.

doc/source/getting_started.rst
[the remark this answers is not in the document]
phackstock: Sounds good, I put it at the same level as regions and variables then.

doc/source/install.rst
phackstock: I tried to install it on my system and read various different versions of whether or not the egg attribute is needed.
Looking at the [pip documentation](https://pip.pypa.io/en/stable/topics/vcs-support/#url-fragments), suggests that it is.
Weirdly though what you @danielhuppmann posted works perfectly fine.
Only once I started looking at the change log of pip for version [21.2](https://pip.pypa.io/en/stable/news/#id49) I found the following:

> Most of the fragment part, including egg=, is explicitly ignored. Only subdirectory= and hash values (e.g. sha256=) are kept.

So it seems that `git+git://github.com/IAMconsortium/nomenclature.git` will work for pip>=21.2 but will probably throw an error and require an egg for older versions.
My vote would therefore be to keep the egg option in as doesn't hurt in the latest versions of pip but is required in earlier ones.

doc/source/getting_started.rst
phackstock: Shortened the getting started page to a minimum working example.

phackstock:
Updated the "Getting started" section. It is much shorter now and contains a minimum working example. All further information will be moved into a "Usage" section.
Since this PR is already quite long, I'd say we should try and merge and I'll open a new PR for more details.

doc/source/getting_started.rst
```diff
@@ -0,0 +1,113 @@
...
+native regions should be renamed and aggregated to comparison regions.
+
+
+Minimum working example
+-----------------------
+
+This section aims to provide a minimum working example of how to use the nomenclature
+package. It is assumed that the variable templates and model mappings already exist.
+Details on how those are structured and how to start using nomenclature "from scratch"
+can be found here :ref:`usage`.
```
LauWien, line 39: I guess this reference will work once we implement a page with the filename usage.rst

doc/source/getting_started.rst
```diff
@@ -0,0 +1,113 @@
...
+A **CodeList** is a list of "allowed terms" (or codes), where each term can have several
+attributes (e.g., description, unit, parent region).
+
+A **RegionProcessor** holds a list of RegionAggregationMappings and a
+DataStructureDefinition. This class is used to facilitate region processing for model
+comparison studies.
+
+A **RegionAggregationMapping** is a mapping that defines on a per-model basis how model
+native regions should be renamed and aggregated to comparison regions.
+
```
LauWien, line 31: I would structure the classes either in alphabetical order or as you state by importance (first main and then auxiliary). I'd vote for the latter, so first "DataStructureDefinition" and "RegionProcessor" and then the rest.

doc/source/getting_started.rst
```diff
@@ -0,0 +1,113 @@
...
+   # Import the necessary libraries
+   import pyam
+   from nomenclature import DataStructureDefinition, RegionProcessor
+   
+   # Initialize DataStructureDefinition and RegionProcessor giving them the
+   # directories where the codelists and mappings are defined as input.
+   dsd = DataStructureDefinition("definitions/")
+   rp = RegionProcessor.from_directory("mappings/", dsd)
+
+   # Read in the data using pyam
```
LauWien, line 54:
```suggestion
   # Read the data using pyam
```

doc/source/getting_started.rst
```diff
@@ -0,0 +1,113 @@
...
+
+**Notes**
+
+* The pyam library is required as *DataStructureDefinition.validate()* and
+  *RegionProcessor.apply()* take a *pyam.IamDataFrame* as input.
+
+* *DataStructureDefinition* and *RegionProcessor* are initialized from directories
+  containing yaml files. See :ref:`dir-structure` for details. 
+
+* *DataStructureDefinition.apply()* returns none if the data frame only contains
```
LauWien, line 71:
```suggestion
* *DataStructureDefinition.apply()* returns *None* if the data frame only contains   
```

doc/source/getting_started.rst
```diff
@@ -0,0 +1,113 @@
...
+  containing yaml files. See :ref:`dir-structure` for details. 
+
+* *DataStructureDefinition.apply()* returns none if the data frame only contains   
+  allowed values and raises an error otherwise.
+
+
+.. _dir-structure:
+
+Directory structure for definitions and mappings
+------------------------------------------------
```
LauWien, line 78: I would move this section up, so one can see the structure first and has a better understanding of the following "Minimum working example". So you gather first all the information you need and it's nice to end the "Getting started" page with a working example.

doc/source/getting_started.rst
```diff
@@ -0,0 +1,113 @@
...
+
+* The names of the *definitions* and *mappings* directories are purely conventional and
+  don't carry any special meaning. As long as *DataStructureDefinition* and
+  *RegionProcessor.from_directory()* are pointed to the correct directories everything
+  will work.
+
+* Inside the *definitions* directory each "dimension", in our case *variable* and
+  *region*, must live in its own sub-directory.
+
+* As opposed *definitions* and *mappings*, *variable* and *region* are special names
```
LauWien, line 106: I think the tests will fail if you don't have a "definitions" folder.

doc/source/api.rst
```diff
@@ -8,3 +10,12 @@ DataStructureDefinition
 
 .. autoclass:: DataStructureDefinition
    :members:
+
+
+
+
```
LauWien, line 16: I think we can reduce here the empty lines

doc/source/api.rst
```diff
@@ -1,3 +1,5 @@
+.. _api:
+
 .. currentmodule:: nomenclature
 
 API documentation
```
LauWien, line 5: I would add also all the other classes as stated in the "Getting started" page "_..(a full list of all classes can be found in API documentation)."_ Maybe also in the same order like in the "Getting started" page.

doc/source/api.rst
```diff
@@ -8,3 +10,12 @@ DataStructureDefinition
 
 .. autoclass:: DataStructureDefinition
    :members:
+
+
+
+
+Region processing
```
LauWien, line 17:
```suggestion
RegionProcessor 
```

doc/source/api.rst
```diff
@@ -8,3 +10,12 @@ DataStructureDefinition
...
+-----------------
+
+.. automodule:: nomenclature.region_mapping_models
```
LauWien, line 20:
```suggestion
.. autoclass:: nomenclature.region_mapping_models
```
Not sure, but I think "automodule" doesn't catch everything important. At least it looks a bit different in the build.

doc/source/index.rst
```diff
@@ -28,12 +28,29 @@ the `Integrated Assessment Modeling Consortium (IAMC) <https://www.iamconsortium
 It supports validation of scenario data and region processing, which consists of
 renaming and aggregation of model "native regions" to "common regions" used in a project.
 
+Those two tasks are carried out by two classes:
+
+#. The **DataStructureDefinition** class handles the validation of scenario data. It
+   contains **CodeLists** for *variables* (including units) and *regions* to be used in
```
LauWien, line 34: I think I would rather use a word which describes "CodeLists" than introduce a new class here.

doc/source/index.rst
```diff
@@ -28,12 +28,29 @@ the `Integrated Assessment Modeling Consortium (IAMC) <https://www.iamconsortium
...
+   a model comparison or scenario exercise following the IAMC data format.
+
+#. The **RegionProcessor** class carries out renaming and aggregation based on information given in model mapping files.
+
+A starting point on how to use these two classes with existing data templates and model mappings can be found in :ref:`getting-started`.
+
+A more ground-up view including the structures for data templates and model mappings is
+given in :ref:`usage`.
```
LauWien, line 42: See earlier comment.

doc/source/install.rst
```diff
@@ -0,0 +1,27 @@
...
+
+Installing From Source
+----------------------
+
+nomenclature can also be installed from source. This will get you the latest version
+of the main branch.
+
+.. code-block:: bash
+
+    pip install -e git+https://github.com/IAMconsortium/nomenclature@main#egg=nomenclature
```
LauWien, line 26:
```suggestion
    $ pip install -e git+https://github.com/IAMconsortium/nomenclature@main#egg=nomenclature
```

doc/source/install.rst
```diff
@@ -0,0 +1,27 @@
...
+.. attention:: The nomenclature requires python >= 3.8
+
+
+Via Pip
+-------
+.. attention::  The nomenclature package is distributed as "nomenclature-iamc" on pypi.
+
+.. code-block:: bash
+
+    pip install nomenclature-iamc
```
LauWien, line 15:
```suggestion
    $ pip install nomenclature-iamc
```

nomenclature/region_mapping_models.py
```diff
@@ -24,35 +29,70 @@
...
+    """Holds information for region processing on a per-model basis.
+
+    Region processing is comprised of native region selection and potentially renaming
+    as well as aggregation to "common regions" (regions used for reporting and
+    comparison by multiple models).
+
+    Attributes
+    ----------
+    model: str
+        Name of the model for which this RegionAggregationMapping is defined.
```
LauWien, line 87:
```suggestion
        Name of the model for which RegionAggregationMapping is defined.
```

nomenclature/region_mapping_models.py
```diff
@@ -24,35 +29,70 @@
...
 
     @property
-    def target_native_region(self):
+    def target_native_region(self) -> str:
+        """Returns the resulting name, i.e. either rename or, if not given, name.
+
+        Returns
+        -------
+        str
+            Resulting name
```
LauWien, line 56:
```suggestion
            Resulting name.
```
I guess convention in this module includes a dot.

nomenclature/region_mapping_models.py
```diff
@@ -203,9 +253,28 @@ class Config:
...
     @validate_arguments(config=dict(arbitrary_types_allowed=True))
     def from_directory(cls, path: DirectoryPath, definition: DataStructureDefinition):
+        """Initialize a RegionProcessor from a directory of model-aggregation mappings.
+
+        Parameters
+        ----------
+        path : DirectoryPath
+            Directory which holds all the mappings.
+        definition : DataStructureDefinition
+            [description]
```
LauWien, line 263:
```suggestion
            ToDo: [description]
```

LauWien (changes requested):
Looks way better! Thank you for the time invested :) 
Just a few suggestions in-line.

doc/source/getting_started.rst
phackstock: Yes exactly, so I would vote to keep it in as I'll start the next PR for adding that as soon as this one is merged.

doc/source/getting_started.rst
phackstock: Ah yeah I was debating that, good point, I'll change the order by importance.

doc/source/install.rst
phackstock: I would not do that as it suggests that the user needs to actually type the dollar sign. See [this question](https://stackoverflow.com/questions/59795354/what-does-the-dollar-sign-represent-in-front-of-lines) on stackoverflow. I would put just the command that the user can copy and paste verbatim into the shell.

doc/source/install.rst
phackstock: See comment above

doc/source/getting_started.rst
phackstock: Do you mean the unit tests or the cli? I just took a look and you're right the cli `assert_valid_structure` would fail as it assumes that 'definitions' and 'mappings' are in place.
This is an inconsistency, though as for the actual classes `DataStructureDefinition` and `RegionProcessor` the constructors take a path argument that can be a directory of any name, not just the fixed 'definitions' and 'mappings'. 
I would vote to fix that inconsistency by either:

1. Changing the constructors to only look for these fixed directory names.
2. Add two optional arguments to the CLI to specify the names of the definitions and mappings directories is they are non-standard. The default values for the parameters would be 'definitions' and 'mappings' of course so it's transparent to the user.

I personally would prefer option 2.

doc/source/api.rst
phackstock: Good point, I was thinking of doing that or leaving it for another PR as some of the doc string are a little rough around the edges and this PR is already long enough. I can just dump them all here though and we can talk about them in a future PR.

doc/source/getting_started.rst
phackstock: Ah yeah I was thinking which one to introduce first, the working example or the layout of the directory. Good to hear that it should be the latter first.

doc/source/getting_started.rst
LauWien: Same. @danielhuppmann what do you think?

doc/source/getting_started.rst
```diff
@@ -0,0 +1,113 @@
...
+
+   .
+   ├── definitions
+   │   ├── region
+   │   │   ├── ...
+   │   │   └── regions.yaml
+   │   └── variable
+   │       ├── ...
+   │       └── variable.yaml
+   └── mappings
```
danielhuppmann, line 92:
```suggestion
   └── mappings [optional]
```

doc/source/getting_started.rst
```diff
@@ -0,0 +1,113 @@
...
+       ├── model_a.yaml
+       └── ...
+
+**Notes**
+
+* The names of the *definitions* and *mappings* directories are purely conventional and
+  don't carry any special meaning. As long as *DataStructureDefinition* and
+  *RegionProcessor.from_directory()* are pointed to the correct directories everything
+  will work.
```
danielhuppmann, line 101: No reason to invite users to start using alternative names...
```suggestion
* The **DataStructureDefinition** will be initialized from the *definitions* folder.
   The **RegionProcessor** will be initialized from the *mappings* folder, if it exists.
```

doc/source/getting_started.rst
danielhuppmann: No reason against extending the testing-module - but I still push users into adopting the conventions, see other suggestions.

doc/source/getting_started.rst
phackstock: The second part might be a bit misleading as it suggests to me that there is some sort of automated mechanism that checks if a mappings folder exists and only instantiates the region processor if it does.
This is not whats currently implemented. If you run:

```python
dsd = DataStructureDefinition("definitions")
rp = RegionProcessor("mappings", dsd)
```
without a directory called "mappings" you will get an error.
So if we want to advertise the fact that "mappings" is optional we should also make it clear that if there are no mappings, `RegionProcessor` must not be used, otherwise the program will crash.

danielhuppmann: Ok, how about "The optional folder *mappings* can be used to initialize a **RegionProcessor**.

doc/source/api.rst
phackstock: Not sure what you mean by that. What is missing if you say it doesn't catch everything important?
I put automodule on purpose since I want sphinx to take all classes from the module nomenclature.region_mapping_models. If I change it to autoclass I don't get anything, which is to be expected since nomenclature.region_mapping_models is a module and not a class.
I could of course do something like this:
```rst

.. autoclass:: nomenclature.region_mapping_models.RegionProcessor
.. autoclass:: nomenclature.region_mapping_models.NativeRegion
etc...
```
But I don't think that's a good idea as there is already a command to document everything from a module.

doc/source/index.rst
phackstock: Will be addressed in future PR.

doc/source/getting_started.rst
phackstock: I wrote an explanation for the next update.

nomenclature/region_mapping_models.py
phackstock: Wrote some description.

doc/source/api.rst
phackstock: If it's not a big deal I'd prefer leaving it as is and make adding the rest of the classes and discussing about that into its own PR.

doc/source/getting_started.rst
phackstock: Sounds good, I've updated the docs.

doc/source/getting_started.rst
phackstock: Done

phackstock:
Should be good to merge from my side.

danielhuppmann (approved):
Thanks @phackstock, looks like a great improvement! Good to be merged from my side, and then we can iterate on some details in a follow-up PR...

doc/source/api.rst
LauWien: Ah sorry my bet, I thought it is about the RegionProcessor class since you referred to api.rst in the getting_started.rst that here are all classes listet. But since all other classes will be dealt with in a new PR it's fine as it is.

LauWien (approved):
👍

PR #65, opened 2021-12-02
Add numpydoc for sphinx

phackstock:
While working on #64 I noticed that I got some sphinx warnings.

``` bash
WARNING: Unexpected section title.

Parameters
----------
```

It seems this has to do with sphinx not recognizing the numpydoc style. If you compare the [pyam api docs](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html) to the [nomenclature api docs](https://nomenclature-iamc.readthedocs.io/en/latest/api.html) you can see that the doc strings are not correctly rendered for nomenclature. 

The issue is that numpydoc was missing as an extension for sphinx.
This PR fixes that.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -34,6 +34,7 @@ release = nomenclature.__version__
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
+    "numpydoc",
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
     "sphinx_click",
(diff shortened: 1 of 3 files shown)


LauWien (commented):
Thank you!

danielhuppmann:
Thank you - but any idea why unit tests are failing now...?

phackstock:
Yes they are failing because in those two tests in question check the error message. In this check it is assumed that the files are read in a particular order. I fixed it here #64, but didn't yet in this branch.
I can check out the file from the branch from #64. That should address it.

phackstock:
Ah never mind it's a different issue related to some changes in the RegionAggregationMapping. I'll take a look.

LauWien (approved):

danielhuppmann (approved):
Thanks!

PR #66, opened 2021-12-02
Change continue to else in CodeList.from_directory

phackstock:
This is a minimal refactor changing the `continue` statement inside the `if` that checks if a yaml file starts with "tag_" to an `else`.

When looking at the code it took me a while to realize that the rest of the processing after the `if` only affects non-tag files. Changing it to an else makes that easier to understand in my opinion.
The functionality stays exactly the same.

### DIFF
--- a/nomenclature/codes.py
+++ b/nomenclature/codes.py
@@ -99,10 +99,10 @@ class CodeList(BaseModel):
                 for item in _code_list:
                     tag = Tag.from_dict(mapping=item)
                     tag_dict[tag.name] = [Code.from_dict(a) for a in tag.attributes]
+
             # if the file does not start with tag, process normally
             else:
-                # validate against the schema of this codelist domain (default
-                # `generic`)
+                # validate the schema of this codelist domain (default `generic`)
                 validate(
                     _code_list, SCHEMA_MAPPING.get(name, SCHEMA_MAPPING["generic"])
                 )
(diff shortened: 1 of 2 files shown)


nomenclature/codes.py
```diff
@@ -99,28 +99,33 @@ def from_directory(
...
+            # if the file does not start with tag, process normally
             else:
-                _code_list = [Code.from_dict(_dict) for _dict in _code_list]
-
-            # add `file` attribute to each element and add to main list
-            for item in _code_list:
-                item.set_attribute("file", str(f.relative_to(path.parent)))
-            code_list.extend(_code_list)
+                # validate against the schema of this codelist domain (default
+                # `generic`)
```
danielhuppmann, line 105: Make one line
```suggestion
                # validate the schema of this codelist domain (default `generic`)
```

nomenclature/codes.py
```diff
@@ -99,28 +99,33 @@ def from_directory(
...
-            if name == "region":
-                _region_code_list = []  # save refactored list as new (temporary) object
-                for top_level_cat in _code_list:
-                    for top_key, _codes in top_level_cat.items():
-                        for item in _codes:
-                            item = Code.from_dict(item)
-                            item.set_attribute("hierarchy", top_key)
-                            _region_code_list.append(item)
-                _code_list = _region_code_list
+            # if the file does not start with tag, process normally
```
danielhuppmann, line 102: Add an empty line for better structure of the code (easier to see that this is a different section)
```suggestion

            # if the file does not start with tag, process normally
```

danielhuppmann (approved):
Two minor suggestions, then good to be merged - thanks!

Issue #67, opened 2021-12-03
Order of validation & aggregation


phackstock:
This issue is in response to openENTRANCE/openentrance#130.

There is an issue that can occur related to the order of operations for validation and region aggregation.
The way these two steps were originally supposed to be run is like this:

``` python
dsd.validate(df)
df = rp.apply(df)
```

That brings a problem because `df` contains might contain regions that are not in the DataStructureDefinition. That might not be a problem though as these regions might be dropped or renamed as part of the region processing. 
This brought the following workaround:

``` python
# only validate the variables (or any other non region dimensions) at this point
dsd.validate(df, dimensions=["variable"]) 
df = rp.apply(df)
```

This works as the model mapping is checked in the region processor, upon instantiation so any illegal regions are going to be detected. There is, however, one case in which this falls short. If we have a model for which we don't have a model mapping defined nothing is checked so we would still need to verify the resulting data frame explicitly.

``` python
# only validate the variables (or any other non region dimensions) at this point
dsd.validate(df, dimensions=["variable"]) 
df = rp.apply(df)
dsd.validate(df, dimensions=["region"]) 
```

This is not the cleanest implementation and can easily lead to wrong errors or illegal regions in the data base. 

The crux of the problem is that the question whether or not the `DataStructureDefinition` should validate the regions of the incoming `df` directly hinges on whether or not a model mapping is defined for the data inside `df`:

* If there **is** a mapping, the `RegionProcessor` has already taken care of this step so the incoming regions **must not** be validated. 
* If there is **no** mapping, the regions **must** be validated by `DataStructureDefinition` as there is no other validation step.

The information of the existence of a model mapping is **only** available to the `RegionProcessor`, **not** the `DataStructureDefinition`. 

Therefore I would propose two different solutions:

1. Move the validation functionality also inside `RegionProcessor`. By creating `RegionProcessor.validate()` we can iterate through the models inside the dataframe and decide on a per model basis whether or not we validate the regions. This would give something like this:
  ```python
  rp.validate(df)
  df = rp.apply(df)
  ```
2. Alternatively, we can make region validation part of `RegionProcessor.apply()`. This would take minimal effort as there is already an if clause in `RegionProcessor.apply()` that checks whether or not a model mapping is present. As of right now it only logs the fact that there is none and then returns the data unchanged. Adding region validation here would be a single line. With this the validation process would look like this: 
```python
dsd.validate(df) # per default anything except regions
[...]

PR #72, opened 2021-12-09
Refactor/rename modules

phackstock:
First step in addressing #67.

Changes as discussed in #67:
* `core` -> `definition`
* `codes` -> `codelist`
* `codes_models` -> `code`
* `region_mapping_models` -> `processor.region`
* `region_mapping_errors` -> `error.region`

We now have dedicated sub-folders for `processor` and `error`.


danielhuppmann (approved):
Thanks, looks good!

PR #76, opened 2021-12-14
Add Usage to docs

phackstock:
Mostly self explanatory, added a "Usage" section to the docs which includes all parts of the documentation that did not fit into the index or the getting started page.

PR #77, opened 2021-12-15
Add intersphinx mapping

danielhuppmann:
This PR adds the intersphinx mapping to pyam and restructures the docs (rst pages and code docstrings) to use intersphinx markup.

### DIFF
--- a/doc/source/getting_started.rst
+++ b/doc/source/getting_started.rst
@@ -1,5 +1,7 @@
 .. _getting-started:
 
+.. currentmodule:: nomenclature
+
 Getting started
 ===============
 
@@ -10,13 +12,13 @@ processing, which consists of renaming of model “native regions” and aggrega
 “common regions” used in a project.
 
 There are two main classes that the user interacts with when using the nomenclature
-package, **DataStructureDefinition** and **RegionProcessor** (a full list of all classes can be found in :ref:`api`). 
+package, :class:`DataStructureDefinition` and :class:`RegionProcessor`.
 
-A **DataStructureDefinition** contains codelists which define allowed *variables*
+A :class:`DataStructureDefinition` contains codelists which define allowed *variables*
 (including units) and *regions* to be used in a model comparison or scenario exercise
 following the IAMC data format.
 
-A **RegionProcessor**  is used to facilitate region processing for model comparison
+A :class:`RegionProcessor` is used to facilitate region processing for model comparison
 studies. It holds a list of model specific mappings which define renaming of native
 regions and aggregation to common regions.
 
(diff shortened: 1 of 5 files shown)


doc/source/getting_started.rst
```diff
@@ -10,13 +12,13 @@ processing, which consists of renaming of model “native regions” and aggrega
 “common regions” used in a project.
 
 There are two main classes that the user interacts with when using the nomenclature
-package, **DataStructureDefinition** and **RegionProcessor** (a full list of all classes can be found in :ref:`api`). 
+package, :class:`DataStructureDefinition` and :class:`RegionProcessor`.
```
LauWien, line 15: Ah that's nice that `DataStructureDefinition` is clickable.

phackstock (approved):
Very nice addition of referencing both within the package and to pyam as well.
Right now, the link for ":class:`RegionProcessor`" is not working because in the api docs I included it as "automodule:: nomenclature.processor.region" and not "autoclass:: RegionProcessor". Should we fix it here or should I address it in #76?

doc/source/getting_started.rst
```diff
@@ -101,13 +103,13 @@ The following outlines how to use the nomenclature package:
...
-  containing yaml files. See :ref:`dir-structure` for details. 
+* :class:`DataStructureDefinition` and :class:`RegionProcessor` are initialized from
+  directories containing yaml files. See :ref:`dir-structure` for details.
 
-* The processor argument of *process* is optional and may only to be used if there are  
-  model mappings. See :ref:`toplevel-functions` for details.
+* The processor argument of func:`process` is optional and may only to be used if there
+  are model mappings. See :ref:`toplevel-functions` for details.
 
 * If not all dimensions of the **DataStructureDefinition** should be validated, a
```
LauWien, line 114:
```suggestion
* If not all dimensions of the :class:`DataStructureDefinition` should be validated, a
```
In the "Notes" of "Directory structure for definitions and mappings" the classes DataStructureDefinition and RegionProcessor could also be referred as :class:`DataStructureDefinition` and :class:`RegionProcessor` as implemented here if wanted.

nomenclature/core.py
```diff
@@ -28,20 +28,20 @@ def process(
...
-    df : pyam.IamDataFrame
-        Input data to be validated and aggregated.
-    dsd : DataStructureDefinition
+    df : :class:`pyam.IamDataFrame`
+        Scenario data to be validated and aggregated.
+    dsd : :class:`DataStructureDefinition`
         Codelists that are used for validation.
     dimensions : list, optional
         Dimensions to be used in the validation, defaults to all dimensions defined in
         *dsd*
```
LauWien, line 37:
```suggestion
        `dsd`
```

nomenclature/core.py
```diff
@@ -28,20 +28,20 @@ def process(
...
-    processor : RegionProcessor, optional
-        Region processor that will perform region renaming and aggregation if provided
+    processor : :class:`RegionProcessor`, optional
+        Region processor to perform region renaming and aggregation (if given)
```
LauWien, line 39: "(if given)" possible redundant since `processor` is "optional"

LauWien (approved):
Looks great :) 
Just a few suggestions inline.

danielhuppmann:
Response to @phackstock:

> Right now, the link for ":class:RegionProcessor" is not working because in the api docs I included it as "automodule:: nomenclature.processor.region" and not "autoclass:: RegionProcessor". Should we fix it here or should I address it in #76?

I think a priority is to make intersphinx operate smoothly also for other users pointing to our docs, so we should make sure that :class:`nomenclature.RegionProcessor` just works. So we should restructure the API documentation (ideally as a separate later PR - it's not a big problem if the link is not created right now).

phackstock:
@danielhuppmann ok, sounds good. Then good to merge from my side. I'll rebase #76 once we've merged.

nomenclature/core.py
danielhuppmann: Yes, redundant, but on purpose, because users may overlook the "optional" tag

PR #79, opened 2021-12-17
Clean up API docs

### LINKED ISSUE #78: Clean up API docs
There are three issues in the API docs:

 - The intersphinx mapping to `nomenclature.RegionProcessor` currently does not work because the API docs are built with automodule, see the comment by @phackstock: https://github.com/IAMconsortium/nomenclature/pull/77#issuecomment-994814351
 - The auto-docs of the pydantic-based classes show a lot of functions that are not relevant to the users
 - Having the type-hints in the function signature makes for pretty-much-useless docs, see below (from [here](https://nomenclature-iamc.readthedocs.io/en/latest/api.html#nomenclature.process))
image

phackstock:
closes #78.

Addresses the three points raised in #78. 
In addition I also added links to the source code. 

One thing that I found while looking into the type hint issue, is a sphinx add on called "sphinx-autodoc-typehints" (https://github.com/agronholm/sphinx-autodoc-typehints). Right now we are essentially writing the type hints twice, once in the function header and once in the docstring. This add on fixes that by taking the type hints directly from the function header.
Maybe an option for the future. What do you guys think @LauWien and @danielhuppmann?

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -19,9 +19,14 @@ DataStructureDefinition
 .. autoclass:: DataStructureDefinition
    :members:
 
+.. autoclass:: nomenclature.codelist.CodeList
+   :members:
 
 Region processing 
 -----------------
 
-.. automodule:: nomenclature.processor.region
+.. autoclass:: RegionProcessor
+   :members:
+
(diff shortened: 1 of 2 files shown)


danielhuppmann (approved):
Looks great, thanks!

PR #83, opened 2022-01-19
Allow region-processing with renaming and/or multiple targets

### LINKED ISSUE #82: Perform aggregation and rename
In the ECEMF project, we came across the issue where several region-aggregation configurations for the same variable should be performed. The particular use case: for carbon prices, it can be beneficial to compute weighted averages using alternative weights:

 - the "obvious" solution to use CO2 emissions as weights, but this can lead to strange results if weights go negative (see https://github.com/IAMconsortium/pyam/issues/446)
 - other weights like Final Energy or GDP, which cannot go negative 

When using alternative weights, it would be helpful to "remember" the alternative weight (and make it explicit) by changing the variable name, so when region-aggregating the "Price|Carbon" variable with "Final Energy" as weight, the resulting timeseries variable could be renamed to  "Price|Carbon (weighted by Final Energy)".

One solution to implement this would be to add a "rename" argument to the variable attributes, e.g.

```yaml
- Price|Carbon:
    description: Price of carbon (for regional aggregrates the weighted price of carbon
        by subregion should be used)
    unit: [EUR_2020/t CO2, USD_2010/t CO2]
    weight: Final Energy
    rename: Price|Carbon (weighted by Final Energy)
```

danielhuppmann:
This PR allows to implement region-processing with variable-renaming and/or performing multiple aggregations. In the example of the unit test, the carbon price is computed in the conventional way (weighted by CO2 emissions) as well as deriving a new variable "Price|Carbon (Max)", i.e. the maximum price across the subregions.

```yaml
- Price|Carbon:
    definition: Price of carbon
    unit: USD/t CO2
    region-aggregation:
      - Price|Carbon:
          weight: Emissions|CO2
      - Price|Carbon (Max):
          method: "max"
```

closes #82


danielhuppmann:
fyi @Renato-Rodrigues.

I chose a slightly different implementation than suggested by you with several lists, instead nesting a similar structure to the list of variables with attributes.

danielhuppmann:
Thinking about this one more time, the following use case (a slight modification of the example above) would only return the maximum-carbon-price variable. The weight-attribute would be ignored.

```yaml
- Price|Carbon:
    definition: Price of carbon
    unit: USD/t CO2
    weight: Emissions|CO2
    region-aggregation:
      - Price|Carbon (Max):
          method: "max"
```
But I think this is not a major issue for now...

And to also answer the questions by @phackstock https://github.com/IAMconsortium/nomenclature/issues/82#issuecomment-1015682099:

> Are the lists in weight and rename the same length

This is not relevant any more thanks to the structure as a list of dictionaries

> Are the entries in rename unique

This is not explicitly checked, but would raise an error via `pyam.concat()` because merging with an overlapping index is prevented by pyam.

phackstock (approved):
Looks good to me overall.
One way to prevent an overlap between `weight` or `method` directly inside a variable definition and the `region-aggregation` section would be to use `oneOf` in the variable yaml schema http://json-schema.org/understanding-json-schema/reference/combining.html#oneof.
This way you could say, either (`weight` or `method`) or `region-aggregation`. Alternatively you could also check during the parsing.
Maybe though this is a non-relevant corner case that we can ignore for now.

danielhuppmann:
I also updated the Usage page of the docs to reflect the new feature.

danielhuppmann:
Thanks for your suggestion @phackstock - I thought about checking for valid keyword-arguments as part of the schema-validation, but this would require keeping the list of pyam-region-aggregation keywords in sync in two locations.

But I'll add this as a "manual" check during DataStructureDefinition-initialization...

phackstock:
@danielhuppmann, ah yes a check like that sounds very good to me.

danielhuppmann:
I added two checks that the region-aggregation-attribute of a variable code must be well-defined (including adding tests for the correct errors)

doc/source/usage.rst
```diff
@@ -32,32 +32,64 @@ formatted as a list of dictionaries mapping the variable (key) to its attributes
...
+
+* Region aggregation for a particular variable can be skipped by using the attribute
+  *skip-region-aggregation*.
+
+    .. code:: yaml
+
+       - Some Variable:
+           skip-region-aggregation: true
+
+* Any attributes which are arguments of :meth:`aggregate_region` will
```
phackstock, line 67: Should this also be :meth:`pyam.IamDataFrame.aggregate_region`?

doc/source/usage.rst
```diff
@@ -32,32 +32,64 @@ formatted as a list of dictionaries mapping the variable (key) to its attributes
...
+  be passed to that method, e.g., *method*, *weight*.
+
+* It is possible to rename the variable returned by the region processing
+  using a *region-aggregation* attribute, which must have a mapping of
+  the target variable name to arguments for the :meth:`aggregate_region` method.
```
phackstock, line 72: Should this also be :meth:`pyam.IamDataFrame.aggregate_region`?

doc/source/usage.rst
```diff
@@ -32,32 +32,64 @@ formatted as a list of dictionaries mapping the variable (key) to its attributes
...
+    .. code:: yaml
+
+       - Some Variable:
+           skip-region-aggregation: true
+
+* Any attributes which are arguments of :meth:`aggregate_region` will
+  be passed to that method, e.g., *method*, *weight*.
+
+* It is possible to rename the variable returned by the region processing
+  using a *region-aggregation* attribute, which must have a mapping of
```
phackstock, line 71: 'must have a mapping' suggests that I have to provide arguments for IamDataFrame.aggregate_region. Maybe more of a conceptual question, would something like this also be possible:

```yaml
variable1:
  unit: unit1
  region-aggregation:
    - variable1_a:
```

Meaning that I just want to rename variable1 to variable1_a but otherwise aggregate normally?
If so maybe it would be good to provide and example for that as well.

nomenclature/codelist.py
```diff
@@ -52,6 +52,36 @@ def keys(self):
     def values(self):
         return self.mapping.values()
 
+    @classmethod
+    def from_list(
```
phackstock, line 56: Since this method is used only once I would prefer to remove it and just include the code in `DataStructureDefinition.__init__`. Unless you already have some more use cases in mind for this function, then I would say it's fine to keep.

nomenclature/definition.py
```diff
@@ -40,6 +53,33 @@ def __init__(self, path, dimensions=["region", "variable"]):
...
+                conflict_args = [i for i in attrs if i in PYAM_AGG_KWARGS]
+                if conflict_args:
+                    raise VariableRenameArgError(
+                        variable=name,
+                        file=attrs["file"],
+                        args=conflict_args,
+                    )
+
+                # ensure that mapped variables are defined in the nomenclature
+                rename_attrs = CodeList.from_list(
```
phackstock, line 74: I think this should also work without having to create another `CodeList` potentially eliminating the need for `CodeList.from_list`:

```suggestion
                invalid = [var_name for var in attrs['region-aggregation] for var_name in var if var_name not in self.variable]
```

nomenclature/error/variable.py
```diff
@@ -0,0 +1,17 @@
+from pydantic import PydanticValueError
+
+
+class VariableRenameArgError(PydanticValueError):
```
phackstock, line 4: Since both of the custom errors here and not used in a pydantic class but in a regular one, sub-classing pydantic errors feels a bit wrong to me. It perfectly works but to me it suggests that they are used inside a pydantic class which is not the case.
Something like this should work just as well:
```suggestion
class VariableRenameArgError(ValueError):
    def __init__(self, args, variable, file):
        super().__init__(f"Using attribute 'region-aggregation' and arguments {args} not supported, occurred in variable '{variable}' (file: {file})")
```
The main appeal of pydantic Errors is that all errors raised during validation are collected and only raised in the very end as a single container error containing a list of all individual errors. In our case a configuration like this:

```yaml
- Price|Carbon:
    definition: Price of carbon
    unit: USD/t CO2
    weight: Emissions|CO2
    region-aggregation:
      - Price|Carbon (Max):
          method: max
```

should raise both errors in this module. Since `DataStructureDefinition` is not a pydantic class I would prefer to use errors sub-classed from 'normal' python exceptions.

nomenclature/error/variable.py
```diff
@@ -0,0 +1,17 @@
...
+    code = "variable_rename_conflict_error"
+    msg_template = (
+        "Using attribute 'region-aggregation' and arguments {args} not supported, "
+        "occurred in variable '{variable}' (file: {file})"
+    )
+
+
+class VariableRenameTargetError(PydanticValueError):
```
phackstock, line 12: Same argument as for `VariableRenameArgError`.

phackstock (changes requested):
Looks good to me overall.
See a few comments in line.

nomenclature/definition.py
danielhuppmann: One additional benefit of casting to a `CodeList` is that this automatically guards against duplicate keys in the list. And I think "cast to a CodeList" is easier to grasp than the nested list comprehension...

doc/source/usage.rst
danielhuppmann: done

doc/source/usage.rst
danielhuppmann: done

nomenclature/codelist.py
danielhuppmann: Done

doc/source/usage.rst
phackstock: Ok, adding that feature once there is a need for it sounds good.

nomenclature/definition.py
```diff
@@ -40,6 +53,33 @@ def __init__(self, path, dimensions=["region", "variable"]):
         if empty:
             raise ValueError(f"Empty codelist: {', '.join(empty)}")
 
+        # check that any variable region-processing-rename mappings are valid
+        if "variable" in self.dimensions:
+            items = [
```
phackstock, line 58: Mabye one more comment, I would put this variable validation code into its own function that is called in the `__init__`. Something like `_validate_variable` or along those lines. 
This should make it more obvious to the user that all these lines are responsible for variable validation

nomenclature/definition.py
phackstock: Ah yes, that safe guard is a fair point. I like the change to adding a pydantic validator to `CodeList`.
Could this validator also be used for the 'normal' `from_directory`?

danielhuppmann: I'm working on it...

phackstock: Very nice.

danielhuppmann:
Thanks for your suggestions @phackstock - I moved the variable-region-aggregation validation into the CodeList class, plus refactored some more errors to PydanticValueError-subclasses.

phackstock (approved):

phackstock:
Looks all good from my side, thanks for implementing everything.
Good to merge.

PR #89, opened 2022-01-26
Add setuptools_scm to requirements

phackstock:
As I recently tried installing nomenclature I saw that `setuptools_scm` is a required dependency that was missing from `setup.cfg`.
It was in there but only under `setup_requires` and since `nomenclature/__init__.py` uses it to get the current version it is an actual dependency.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,7 @@ install_requires =
     pydantic
     pyyaml
     jsonschema
+    setuptools_scm
 setup_requires =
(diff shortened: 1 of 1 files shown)


danielhuppmann (approved):
Thanks for catching this!

phackstock:
Pandas was also missing. Not too big of an issue since that should be covered by pyam. Still good to explicitly list it since we explicitly import it.

PR #90, opened 2022-01-27
Feature/automate release

### LINKED ISSUE #57: Automate release
For future releases of the nomenclature package on pypi we want a GitHub action that automatically updates the pypi entry.

Inspiration may be taken from the [`publish.yaml`](https://github.com/IAMconsortium/units/blob/main/.github/workflows/publish.yaml) workflow from the units package.

phackstock:
closes #57.

Added the github action to publish to pypi in the event of a release. 
I tested it on my fork with test-pypi. Deployment keys are also added.
I've also linked the repo to Zenodo. For the next release we should get a DOI.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,39 @@
+name: Publish to PyPI and TestPyPI
+
+on: push
+  branches:
+    - feature/automate-release
+
+jobs:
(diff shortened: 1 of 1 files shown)


danielhuppmann (approved):
Looks good to me, thanks!

PR #91, opened 2022-01-28
Refactor tags to use f-string formatting

### LINKED ISSUE #88: Refactor tags to f-string style
The current use of `<Tag>` could be changed to `{Tag}` to be closer to Python f-string formatting.

Given that there are currently only two or three repositories using tags, this should be a manageable task to implement the change and update the repos simultaneously.

Thoughts @phackstock?

danielhuppmann:
This PR changes the use of the tag feature to use a format similar to f-strings instead of the custom `<..>` style.

It is not possible to have keys of the format `{Tag}` (starting with a curly bracket) in a yaml-dictionary file, therefore the structure of the yaml files is changed such that only the tag name is the key (without curly brackets).

closes #88

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -3,9 +3,6 @@ from typing import Union, List, Dict
 from pydantic import BaseModel, validator
 
 
-TAG_PATTERN = compile("^<.*>$")
-
-
 class Code(BaseModel):
     """A simple class for a mapping of a "code" to its attributes"""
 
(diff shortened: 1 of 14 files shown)


nomenclature/code.py
```diff
@@ -61,11 +51,11 @@ def _replace_tags(code, tag, target_list):
     _code_list = []
 
     for target in target_list:
-        key = code.name.replace(tag, target.name)
+        key = code.name.replace("{" + tag + "}", target.name)
```
phackstock, line 54: Could use an f-string for that ;) but maybe with escaping the curly brackets the string concatenation is easier to read.
```suggestion
        key = code.name.replace(f"{{{tag}}}", target.name)
```

PR #92, opened 2022-01-28
Add a quickfix for conflict between codes and tags

danielhuppmann:
PR #91 introduced a bug: if a code (variable name) contains the name of a tag (e.g., the tag "Industry" and the variable "Final Energy|Industry"), then the variable was added multiple times during the replace-tag phase, causing a duplicate-code error.


danielhuppmann:
FYI @phackstock

PR #93, opened 2022-02-03
Collect errors during reading of region 

phackstock:
As the recent PR in the navigate workflow has shown it would be very convenient to get all errors related to region mappings in one single go instead of getting one, fixing one and running `RegionProcessor.from_directory()` over and over again.

This PR addresses that, errors during reading are collected and returned as a single `pydantic.Validation` error that contains all individual errors. This way mappings can be fixed much faster.

I added a test `test_region_processor_multiple_wrong_mappings` which reads in a total of 8 mappings, 7 of which cause errors.

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -6,7 +6,7 @@ from typing import Dict, List, Optional, Set, Union
 import pyam
 import pydantic
 import yaml
-from jsonschema import ValidationError, validate
+import jsonschema
 from pyam import IamDataFrame
 from pydantic import BaseModel, root_validator, validate_arguments, validator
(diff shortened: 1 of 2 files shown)


phackstock:
Test currently fails because `test_region_aggregation.test_region_processor_multiple_wrong_mappings` checks for the error message. In this error message the order in which the mappings are read is important and different on the GitHub servers compared to my local machine.

phackstock:
Simplified the test to only check for the number of errors reported without implicitly testing for the order in which files are read in. 
Now all the tests pass.

danielhuppmann (approved):
Looks good to me, thanks @phackstock

PR #94, opened 2022-02-03
Feature/add yml parsing

### LINKED ISSUE #85: Only `yaml` files are found not `yml`
When scanning a directory for region aggregation mappings, only `.yaml` files are found while  `.yml` is ignored.

phackstock:
closes #85.

I've added support for reading yaml files which have the `.yml` suffix instead of `.yaml`. For now this only applies to region mappings and the test function `testing.assert_valid_yaml`. 

For reading code lists right now, the user can provide if a suffix so a choice between `.yaml` and `.yml` is possible there. What's not possible right now is to combine the two.

What are your thoughts @danielhuppmann and @LauWien? Should we allow combining the two or should I modify the region mappings and the test function so that one and only one has to be selected. Because as it stands right now there is some inconsistency which is never a good thing.


danielhuppmann:
Thanks @phackstock! I think that it could be confusing to accept both `.yaml` and `.yml` in the region-processing automatically, but require a user choice for the code-list files.

phackstock:
Totally agreed, what do you think should we require a choice everywhere or accept both `.yaml` and `.yml`?
My preference would be for the latter as I don't think that there are any downsides to accepting both while it can cause some friction especially with 'old-style' region aggregation mappings if you have to manually switch between them.

danielhuppmann:
Agree, set both as default, and also allow to manually set the file-type as a keyword argument in all places.

phackstock:
Is there a use case where we manually want to set only one file type?
If there is I'm happy to include that but if there is (currently) not I would say to just hardcode `.yaml` and `.yml` reducing complexity. Seeing as this would also affect the CLI, giving the option to chose would need to be implemented in a few places.

nomenclature/processor/region.py
```diff
@@ -274,24 +274,25 @@ def from_directory(cls, path: DirectoryPath):
...
-                    errors.append(
-                        ErrorWrapper(
-                            ModelMappingCollisionError(
-                                model=mapping.model,
-                                file1=mapping.file,
-                                file2=mapping_dict[mapping.model].file,
-                            ),
-                            "__root__",
+        for file in path.glob("**/*"):
+            if file.suffix in {".yaml", ".yml"}:
```
danielhuppmann, line 278: Suggest to use this more elaborate formulation to avoid double-indenting
```suggestion
        for file in [f for f in path.glob("**/*") if f.suffix in {".yaml", ".yml"}]:
```

nomenclature/codelist.py
```diff
@@ -153,51 +150,51 @@ def from_directory(
...
-                    tag_dict[tag.name] = [Code.from_dict(a) for a in tag.attributes]
-
-            # if the file does not start with tag, process normally
-            else:
-                # validate the schema of this codelist domain (default `generic`)
-                validate(
-                    _code_list, SCHEMA_MAPPING.get(name, SCHEMA_MAPPING["generic"])
-                )
+        for f in path.glob(file):
+            if f.suffix in {".yaml", ".yml"}:
```
danielhuppmann, line 157: Suggest to use this more elaborate formulation to avoid double-indenting
```suggestion
        for file in [f for f in path.glob("**/*") if f.suffix in {".yaml", ".yml"}]:
```

nomenclature/testing.py
```diff
@@ -12,13 +12,14 @@ def assert_valid_yaml(path: Path):
...
     error = False
-    for file in path.glob("**/*.yaml"):
-        try:
-            with open(file, "r", encoding="utf-8") as stream:
-                yaml.safe_load(stream)
-        except (yaml.scanner.ScannerError, yaml.parser.ParserError) as e:
-            error = True
-            logger.error(f"Error parsing file {e}")
+    for file in path.glob("**/*"):
+        if file.suffix in {".yaml", ".yml"}:
```
danielhuppmann, line 16: Suggest to use this more elaborate formulation to avoid double-indenting

```suggestion
    for file in [f for f in path.glob("**/*") if f.suffix in {".yaml", ".yml"}]:
```

phackstock:
Removed the unnecessary nesting.

nomenclature/codelist.py
```diff
@@ -153,17 +150,15 @@ def from_directory(
...
         """
         code_list, tag_dict = [], CodeList(name="tag")
-
         # parse all files in path if file is None
         file = file or "**/*"
 
-        # parse all files
-        for f in path.glob(f"{file}{ext}"):
-            with open(f, "r", encoding="utf-8") as stream:
+        for yaml_file in (f for f in path.glob(file) if f.suffix in {".yaml", ".yml"}):
```
danielhuppmann, line 156: Maybe still refactor to simply „file“ for consistency?

danielhuppmann (approved):
Looks great, one suggestion inline - good to merge either way!

nomenclature/codelist.py
phackstock: Yes that what I had originally, but `file` is already the name of the input value for the file pattern in `from_directory`. Two lines above we have `file = file or "**/*"`. I didn't want to overwrite the variable but I also did not want to change the name of the input values of `from_directory`.
If you're fine with overwriting `file`, changing the name of the input parameter or a better name than `yaml_file` (which admittedly I'm also not super happy with), I'm all up for suggestions.

danielhuppmann: Ok, got it, then keep as is!

PR #95, opened 2022-02-03
Silence logger during region processing

danielhuppmann:
When working on the navigate-refactoring, I noticed that calling region-processing on non-existing regions or (lists of) variables prints 
```
pyam.utils - WARNING: Formatted data is empty!
```
quite a few times to the log, with no actually helpful information.

This PR uses the pyam `adjust_log_level()` function to silence the pyam-logger for that part of the code.

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -315,13 +315,13 @@ class RegionProcessor(BaseModel):
 
             # If no mapping is defined the data frame is returned unchanged
             if model not in self.mappings:
-                logging.info(f"No region aggregation mapping found for model {model}")
+                logger.info(f"No region aggregation mapping found for model {model}")
                 processed_dfs.append(model_df)
             # Otherwise we first rename, then aggregate
             else:
                 # before aggregating, check that all regions are valid
                 self.mappings[model].validate_regions(dsd)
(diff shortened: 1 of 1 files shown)


nomenclature/processor/region.py
```diff
@@ -315,57 +316,59 @@ def apply(self, df: IamDataFrame, dsd: DataStructureDefinition) -> IamDataFrame:
...
-                    ]
-                    vars_kwargs = {
-                        var: kwargs
-                        for var, kwargs in vars.items()
-                        if var not in vars_default_args
-                    }
-                    for cr in self.mappings[model].common_regions:
-                        regions = [cr.name, cr.constituent_regions]
-                        # First perform 'simple' aggregation (i.e. no aggregation args)
+                with adjust_log_level():  # silence empty filter
```
phackstock, line 330: I looked up the default value for the log level which I didn't know was "ERROR". Maybe just to make it clear to the reader you could explicitly set it.
```suggestion
                with adjust_log_level(level="ERROR"):  # silence empty filter
```

phackstock (commented):
Just one small remark otherwise looks good to me.

phackstock (approved):
Looks good to be merged.

PR #98, opened 2022-02-11
Add a `CodeList.to_excel` method

danielhuppmann:
# Overview

This PR adds a method (and test) to write the items from a CodeList to an xlsx spreadsheet.

# Usage

If a user has cloned the project-specifications repository from Github, she can use the following to translate a dimension e.g., the list of *variables*, to an excel spreadsheet.

```python
from pathlib import Path
from nomenclature import CodeList

CodeList.from_directory("Variable", Path("definitions/variable")).to_excel(<file>)
```

where *file* is the (folder and) name of the target file.

### DIFF
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -3,7 +3,9 @@ import yaml
 from typing import Union, Dict, List
 from pydantic import BaseModel, root_validator
 from jsonschema import validate
+import pandas as pd
 
+from pyam.utils import write_sheet
 
 from nomenclature.code import Code, Tag, replace_tags
 from nomenclature.error.codelist import DuplicateCodeError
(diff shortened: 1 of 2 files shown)


danielhuppmann:
fyi @strefler @MathijsHarmsenPBL

nomenclature/codelist.py
```diff
@@ -201,3 +203,36 @@ def from_directory(
 
         # iterate over the list to guard against silent replacement of duplicates
         return CodeList(name=name, mapping=code_list)
+
+    def to_excel(self, excel_writer, sheet_name="definitions"):
```
phackstock, line 207: Is there a specific reason that the user would want to provide their own `excel_writer`?
If not I would find it more convenient that the argument would only contain the path to the target file and be renamed `target`, `output_file_path` or something along those lines.

phackstock (commented):
Looks good to me. I like the how this the counterpart to `to_yaml`, very nice.
Just one small question about the input parameters of `to_excel` below.

nomenclature/codelist.py
danielhuppmann: If you look at my analysis scripts for the IPCC SR15 (see last cells [here](https://data.ene.iiasa.ac.at/sr15_scenario_analysis/assessment/sr15_2.0_categories_indicators.html)), I wanted to write several sheets within one xlsx file. For that, using the **ExcelWriter** is quite helpful.

phackstock: Ah alright, that is a fair use case. This could also be useful in case we ever want to implement `to_excel` on a `DataStructureDefinition` basis where we would export and entire `DataStructureDefinition` in a single excel file into different sheets.

phackstock (approved):
Good to merge from my side.

PR #99, opened 2022-02-15
Partial region aggregation

### LINKED ISSUE #96: Partial region-aggregation for common-regions
There is a use case where part of the timeseries data for a region is included in the **IamDataFrame** (e.g., scenario data uploaded to the IIASA Scenario Explorer), and other part of the data is aggregated from the sub-regions (as is currently done by the region processor).

A simple solution would be to keep any timeseries data for common-regions and only compute the aggregate-region-data for other variables.

Questions to be discussed:
- Should we check for conflicts? e.g., a value provided at "World" level is different from the value computed by the RegionProcessor... Should this be an error or warning?

phackstock:
closes #96.

## New Feature

This PR implements partial region aggregation. 
Partial region aggregation happens when a common region is already being reported from a model natively. 
This can encompass all variables or only a subset. Additionally this also works for variables that are **only** reported on this region level.

### Example

Say we have model native results that look like this:

* *model_a* reports a total of three regions: *region_a*, *region_b* and *World*
* Regions *a* and *b* are native regions and *World* falls under the category of common regions (although it is reported by the model directly)
* For region aggregation _World_ is comprised of regions _a_ and _b_.
* For region _region_a_ and _region_b_ we report variables _var_a_ and _var_b_.
* For region _World_ we report _var_b_ and _var_c_.

The corresponding model mapping like like this:

```yaml
model: model_a
native_regions:
- region_a
- region_b
common_regions:
- world:
  - region_a
  - region_b
```

The region processor now applies the following logic:

1. Variable _var_a_ is **not** reported on _World_ level, therefore it is calculated through region aggregation of regions _region_a_ and _region_b_.
2. Variable _var_b_ **is** reported on _World_ level **as well as** _region_a_ and _region_b_. **By convention** the model native results (on _World_ level) take precedence over preforming aggregation.
3. Variable _var_c_ is **only** reported on _World_ level so it taken from there.

The resulting data frame then contains the following:
* For _region_a_ and _region_b_: variables: _var_a_ and _var_b_
* Fro _World_: variables _var_a_ (calculated by aggregation) , _var_b_ (taken from model native results by precedence), _var_c_ (taken from model native results)

## Open Points

* As mentioned in point 2. of the region processor logic, we currently simply take the model native results over the aggregation. Additionally we could also calculate the aggregation and compare the results giving an error if the results do not match.


tests/data/region_processing/dsd/variable/variables.yaml
```diff
@@ -7,3 +7,5 @@
 - Share|Coal:
     definition: Share of Coal in the total primary energy mix
     unit:
+- World Variable:
```
danielhuppmann, line 10: Global mean temperature (increase) is a variable that is typically only reported at the World level...
```suggestion
- Temperature|Mean:
```

nomenclature/processor/region.py
```diff
@@ -349,6 +349,21 @@ def apply(self, df: IamDataFrame, dsd: DataStructureDefinition) -> IamDataFrame:
                         }
                         for cr in self.mappings[model].common_regions:
                             regions = [cr.name, cr.constituent_regions]
+                            # Check for already present variables
+                            model_native_res = model_df.filter(region=cr.name)
+                            if len(model_native_res):
+                                processed_dfs.append(model_native_res)
+                                model_native_vars = set(model_native_res.variable)
+                                # Remove all already present variables
+                                vars_default_args = list(
```
danielhuppmann, line 358: I think that you have to create a copy here...
If you perform this operation for several common regions, you remove arguments from the dictionary in each iteration - so if the first region has *Variable A* (and is omitted from the aggregation), that variable will also be ignored for the second region...

danielhuppmann (commented):
I think that implementation was too easy... Best to extend the test, I guess...

nomenclature/processor/region.py
phackstock: Ah yes, good catch, the copy will be necessary. 
I'll also create a dedicated, more focused test case that covers multiple common regions.

phackstock: done.

nomenclature/processor/region.py
```diff
@@ -345,6 +343,7 @@ def apply(self, df: IamDataFrame, dsd: DataStructureDefinition) -> IamDataFrame:
 
                     # Aggregate
                     if self.mappings[model].common_regions is not None:
+                        # processed_dfs_region: List[IamDataFrame] = []
```
danielhuppmann, line 346: Is this a mistake?

nomenclature/processor/region.py
```diff
@@ -403,6 +408,43 @@ def _filter_dict_args(
...
+def _aggregate_with_model_native(
+    model_df: IamDataFrame,
+    vars: Union[str, List[str]],
+    common_region: str,
+    constituent_regions: List[str],
+    **aggregation_kwargs,
+) -> IamDataFrame:
+
+    # Get all results for common_region
+    model_native = model_df.filter(region=common_region, variable=vars)
```
danielhuppmann, line 420: Why is this called `model_native` if it is the selection of the common regions?

nomenclature/processor/region.py
```diff
@@ -403,6 +408,43 @@ def _filter_dict_args(
...
+    # Aggregate for comparison
+    if aggregation_kwargs:
+        aggregated = _aggregate_region(
+            model_df, vars, *[common_region, constituent_regions], **aggregation_kwargs
```
danielhuppmann, line 424: Why not use as args directly?
```suggestion
            model_df, vars, common_region, constituent_regions, **aggregation_kwargs
```

nomenclature/processor/region.py
```diff
@@ -403,6 +408,43 @@ def _filter_dict_args(
...
+            model_native.filter(variable=comp_vars),
+            aggregated.filter(variable=comp_vars),
+            left_label="model native",
+            right_label="aggregated",
+        )
+        if len(diff):
+            logging.warning(
+                f"Differences found between model native and aggregated results\n{diff}"
+            )
+        # Take all model native results and add the ones that are only available from
```
danielhuppmann, line 442: Please rephrase to fit in one line

danielhuppmann (commented):
Thanks @phackstock - the mechanics of the implementation look great, but I have a few implementation suggestions...

nomenclature/processor/region.py
phackstock: Ah yes, I forgot to remove it. It's taken out now.

nomenclature/processor/region.py
phackstock: I called it that way since it's results that are natively reported from the model and not obtained through region aggregation. I can use a different name though.

nomenclature/processor/region.py
```diff
@@ -403,6 +408,43 @@ def _filter_dict_args(
...
+        # aggregation
+        return model_native.append(aggregated.filter(variable=comp_vars, keep=False))
+    return model_native
```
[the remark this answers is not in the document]
phackstock: Yes exactly, this is an implicit `else`, I'll make it explicit and put a comment in place.

nomenclature/processor/region.py
```diff
@@ -356,29 +355,35 @@ def apply(self, df: IamDataFrame, dsd: DataStructureDefinition) -> IamDataFrame:
...
                         }
                         for cr in self.mappings[model].common_regions:
-                            regions = [cr.name, cr.constituent_regions]
                             # First, perform 'simple' aggregation (no arguments)
                             processed_dfs.append(
-                                model_df.aggregate_region(vars_default_args, *regions)
+                                _aggregate_with_model_native(
+                                    model_df,
+                                    vars_default_args,
+                                    cr.name,
```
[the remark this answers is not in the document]
phackstock: Ah yes good point, I'll go back to that.

nomenclature/processor/region.py
phackstock: Good point, I've changed it to your suggestion further down of using `regions` directly.

nomenclature/processor/region.py
danielhuppmann: I think this is rather confusing here: there are native regions and common regions, and there are originally-submitted results versus a processed IamDataFrame (including the aggregated results). I would still rather call it `common_region_df`.

nomenclature/processor/region.py
```diff
@@ -403,6 +408,43 @@ def _filter_dict_args(
         }
 
 
+def _aggregate_with_model_native(
```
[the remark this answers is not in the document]
danielhuppmann: You are right, I originally created this as an own function to avoid duplication, but this is not a concern anymore because of the restructuring...

nomenclature/processor/region.py
phackstock: Ah ok, so using `common_region_df` for results that are (potentially) comprised of originally-submitted as well as processed results. I'll use that name and add a comment to make it more clear.

nomenclature/processor/region.py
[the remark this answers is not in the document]
phackstock: Ok sounds good, I'll put it into one function then.

nomenclature/processor/region.py
phackstock: done.

phackstock:
Thanks for the suggestions @danielhuppmann, I updated the code accordingly.
The two `_aggregate_region...` functions are now combined into a single one.

I also added a check for the warning message in `test_region_processing_weighted_aggregation`, in the third parameter configuration. I can also split this third one into its own test if you think that would bring more clarity. I could also revert this commit and add it its own PR since it's only partially related to this one. Additionally, we could also test the weighted aggregation for the missing variable failure scenario. I ran a quick check and the error it produces is the same "Inconsistent index between ...". It would be quite straightforward to distinguish between the two cases, we would just need to include a check for the variables present in the data frame inside the except clause. But that's probably good to include in a separate PR. 

Coming back to the open points I raised in my comment two days ago:
* Should I include a specific check on a scenario level in what is now `_aggregate_region`? As it stands the comparison is attempted for one region, model and variable but for all scenarios. Right now we would lose a variable that's missing some scenarios in the uploaded results but is present in the aggregated ones. Is edge case realistic or can we live without having to explicitly iterate over all scenarios?
* The renaming is done after the aggregation so model native results would still take precedence here. I could see a (admittedly far fetched) use case where someone might want to upload model native results under the original variable name and the aggregated results under the new name, but I think this is too convoluted and we should keep it as it currently is. 
* If results for common regions are reported from the model directly but a variable has `skip-region-aggregation` as `True` this variable will not be uploaded even though technically we don't need region aggregation to obtain it. Is this behavior ok?

danielhuppmann:
> The two `_aggregate_region...` functions are now combined into a single one.

Thanks, looks good!

> I also added a check for the warning message in `test_region_processing_weighted_aggregation`, in the third parameter configuration.

Nice, fine to leave here.

> Additionally, we could also test the weighted aggregation for the missing variable failure scenario. I ran a quick check and the error it produces is the same "Inconsistent index between ...". It would be quite straightforward to distinguish between the two cases, we would just need to include a check for the variables present in the data frame inside the except clause. But that's probably good to include in a separate PR.

Not quite sure what you mean, maybe make an issue and tackle later...

> 
> Coming back to the open points I raised in my comment two days ago:
> 
> * Should I include a specific check on a scenario level in what is now `_aggregate_region`? As it stands the comparison is attempted for one region, model and variable but for all scenarios. Right now we would lose a variable that's missing some scenarios in the uploaded results but is present in the aggregated ones. Is edge case realistic or can we live without having to explicitly iterate over all scenarios?

Now that you specifically ask... It might make sense to change the implementation strategy. Do all aggregation first, then do the comparison. I gave it a try, see https://github.com/phackstock/nomenclature/pull/2

Additional benefits: faster performance (because filtering and comparison only once, not in every iteration), and only one warning message per model.

> * The renaming is done after the aggregation so model native results would still take precedence here. I could see a (admittedly far fetched) use case where someone might want to upload model native results under the original variable name and the aggregated results under the new name, but I think this is too convoluted and we should keep it as it currently is.

The suggestion would also take care of this.

> * If results for common regions are reported from the model directly but a variable has `skip-region-aggregation` as `True` this variable will not be uploaded even though technically we don't need region aggregation to obtain it. Is this behavior ok?

The suggestion would also take of it, because it filters to all valid variables provided at a common-region.

phackstock:
Thanks a lot for the PR @danielhuppmann, we can continue the discussion over there.
Overall it looks very promising to me, both taking care of some corner cases and providing a speed boost is a really nice improvement.

phackstock:
## Updates

### Tests

Added unit tests for partial region aggregation covering the following cases:
* Variable is available in provided and aggregated data and the same
* Variable is only available in the provided data
* Variable is only available in the aggregated data
* Variable is not available in all scenarios in the provided data
* Using skip-aggregation: true should only take provided results
* Using the region-aggregation attribute to create an additional variable
* Variable is available in provided and aggregated data but different

Tests also check logging output.
Thanks a lot again for your solution @danielhuppmann, it worked for all the edge cases I threw at it.

### Compare & merge function

@danielhuppmann I took your compare & merge for provided and aggregated results and put it in a dedicated function. Now `RegionProcessor.apply` is a little bit easier to read.

### Adjust pyam log level

In order to get rid of some warnings from pyam regarding empty data frames as a result of the comparison between provided and aggregated results I extended the context manager `with adjust_log_level(logger="pyam", level="ERROR"):` to include this part.

## Open points

From my side the only thing that's remaining is to update the docs. I would put this in the here https://nomenclature-iamc.readthedocs.io/en/latest/usage.html#regionprocessor, if that works for you @danielhuppmann.

phackstock:
When writing the documentation I stumbled upon a possible bug in the way partial region aggregation is performed.
Say we have a variable that is defined like this:

```yaml
- Variable B:
    unit: EJ/yr
    region-aggregation:
        - Variable B:
            method: max
```

with the following model mapping:

```yaml
model: m_a
common_regions:
  - World:
    - region_A
    - region_B
```

then providing data such as this: 
```python
[
    ["m_a", "s_a", "region_A", "Variable B", "EJ/yr", 1, 2],
    ["m_a", "s_a", "region_B", "Variable B", "EJ/yr", 3, 4],
    ["m_a", "s_a", "World", "Variable B", "EJ/yr", 4, 6]
]
```

and performing region aggregation would give us, according to our "provided data takes precedence over aggregated"-rule:

```
["m_a", "s_a", "World", "Variable B", "EJ/yr", 4, 6]
```

which would be incorrect as `Variable B` is defined as the max and not the sum, so we should instead get:

```
["m_a", "s_a", "World", "Variable B", "EJ/yr", 3, 4]
```

What are your thoughts on this @danielhuppmann? 
One option would be to disallow renaming a variable to itself when using the `region-aggregation` attribute as part of a variable code list.

phackstock:
Added some documentation about partial region aggregation. As touched on before, the docs will probably need another re-work soon, but that's a separate issue.

danielhuppmann:
Re the potential bug or inconsistency in the partial region aggregation... I do not think that this is a bug or problem (from a technical-implementation point of view).

If "Variable B" is defined such that the maximum across regions should be reported at the "World" level, and a team reports the sum instead of the max, this is clearly a reporting error. The package can help identify reporting errors by writing a warning to the log, but it is not the purpose of the package to be smarter than the user...

phackstock:
Ah yes fair enough, then it should be all good. The warning is written already anyway.
Since we have a few cases now where the user should read the log file maybe we can think about a way how to present the log file to the user in the scenario explorer.

danielhuppmann (approved):
Looks good to me!

PS: We should revisit the docs and harmonize the notation (provided data vs. original) in the not-too-distat future...

phackstock:
Agreed on the docs part, I'm also not super happy with the distinction between "Getting Started" and "Usage". I find myself looking for things, not knowing in which part they are.

PR #100, opened 2022-02-16
Support list of models in a region mapping

### LINKED ISSUE #84: Allow multiple "model" values in a region-mapping file
The "model" attribute in a region-processing mapping-file currently only takes a string value. It would be beneficial to allow this to be a list so that one region-mapping can be used to several (minor/patch) versions of a model.

So the extended use of the region-mapping file could look like the following (based on https://github.com/openENTRANCE/openentrance/pull/148):

```yaml
model:
  -  OSeMBE v1.0.0
  -  OSeMBE v1.1.0
  -  OSeMBE v1.2.0
native_regions:
  - AUT: OSeMBE v1|AUT
  ...

 common_regions:
   - EU27:
     - AUT
     ...
```

or

```yaml
model:
  -  OSeMBE v1.0.0
  -  OSeMBE v1.0.1
  -  OSeMBE v1.0.2
native_regions:
  - AUT: OSeMBE v1.0|AUT
  ...

 common_regions:
   - EU27:
     - AUT
     ...
```
depending on the likelihood of region-changes over the model versions

danielhuppmann:
Per this PR and comment https://github.com/iiasa/navigate-workflow/pull/29#discussion_r808207954, the feature to support a list of models in a region-mapping become a bit more relevant...

closes #84

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -93,7 +93,7 @@ class RegionAggregationMapping(BaseModel):
         Optionally, list of common regions where aggregation will be performed.
     """
 
-    model: str
+    model: Union[List[str], str]
     file: FilePath
     native_regions: Optional[List[NativeRegion]]
     common_regions: Optional[List[CommonRegion]]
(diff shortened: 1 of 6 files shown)


nomenclature/processor/region.py
```diff
@@ -93,7 +93,7 @@ class RegionAggregationMapping(BaseModel):
         Optionally, list of common regions where aggregation will be performed.
     """
 
-    model: str
+    model: Union[List[str], str]
```
phackstock, line 96: Just one (mostly stylistic) comment. Instead of allowing both `str` and `List[str]` you can also just allow a list of strings:

```suggestion
    model: List[str]
```

in combination with a pre validator that converts it to a list if it is provided as just a string

```python

@validator("model", pre=True)
def convert_to_list(cls, v):
    return pyam.utils.to_list(v)
```

this would mean that for any interaction with `RegionAggregationMapping` you can be sure that `model` is a list and iterate over it directly instead of having to check if it is a string or a list.

phackstock (approved):
Looks good to me. 
One small stylistic comment below but I don't have strong feeling about either way.

nomenclature/processor/region.py
danielhuppmann: Thanks for the suggestion - implemented.

phackstock:
Thanks for the changes @danielhuppmann, good to merge from my side now.

Issue #87, opened 2022-01-24
Weighted region aggregation fails when weight is missing


phackstock:
When performing weighted region aggregation it needs to be made sure that both the variable and the weight are present in **all** model, scenario combinations. 
From my point of view it would be desirable if it would produce a warning if there are only a couple of model, scenario combinations where weight and/or variable are missing and an error if that's the case for all.
Maybe this is more well suited to be addressed on the pyam side though.
What are your thoughts on the matter @danielhuppmann?

PR #101, opened 2022-02-17
Quickfix for region-aggregation with missing weights

danielhuppmann:
This PR is a first quickfix for #87. Added to-do notes for a better separation between "weight does not exist at all" versus "weight exists but index is inconsistent".

Because in a related section of the code and urgent, I also added a quickfix for https://github.com/IAMconsortium/pyam/issues/631 directly to this PR. Renaming an empty IamDataFrame raises an unintended error, so I separated the aggregation-and-rename into separate steps with an if in-between.

FYI @HauHe @Renato-Rodrigues

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -347,6 +347,7 @@ class RegionProcessor(BaseModel):
                         vars_default_args = [
                             var for var, kwargs in vars.items() if not kwargs
                         ]
+                        # TODO skip if required weight does not exist
                         vars_kwargs = {
                             var: kwargs
                             for var, kwargs in vars.items()
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -361,19 +361,27 @@ class RegionProcessor(BaseModel):
                             # Second, special weighted aggregation
                             for var, kwargs in vars_kwargs.items():
                                 if "region-aggregation" not in kwargs:
-                                    processed_dfs.append(
-                                        model_df.aggregate_region(
-                                            var, *regions, **kwargs
-                                        )
+                                    _df = _aggregate_region(
+                                        model_df,
+                                        var,
+                                        *regions,
+                                        **kwargs,
                                     )
+                                    if _df is not None:
+                                        processed_dfs.append(_df)
                                 else:
                                     for rename_var in kwargs["region-aggregation"]:
                                         for _rename, _kwargs in rename_var.items():
-                                            processed_dfs.append(
-                                                model_df.aggregate_region(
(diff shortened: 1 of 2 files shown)


phackstock:
Should I give this a review @danielhuppmann?

danielhuppmann:
Yes, a brief one, please…

tests/test_core.py
```diff
@@ -162,10 +188,21 @@ def test_region_processing_complete(directory):
...
             ],
+            None,
+        ),
+        # check that region-aggregation with missing weights passes (inconsistent index)
+        # TODO check the log output
+        (
+            "weighted_aggregation",
+            [
+                ["model_a", "scen_a", "World", "Primary Energy", "EJ/yr", 4, 6],
+                ["model_a", "scen_a", "World", "Emissions|CO2", "Mt CO2", 5, np.nan],
```
phackstock, line 199: Just a question for my understanding. The filtering dictionary which removes the 2010 entry for Emissions|CO2 is needed because directly creating a data frame with the 2010 data missing for one variable would throw an error, right?

nomenclature/processor/region.py
```diff
@@ -361,21 +365,32 @@ def apply(self, df: IamDataFrame, dsd: DataStructureDefinition) -> IamDataFrame:
...
+                                            if _df is not None and not _df.empty:
+                                                processed_dfs.append(
+                                                    _df.rename(variable={var: _rename})
+                                                )
 
-        return pyam.concat(processed_dfs)
+        if processed_dfs:
+            return pyam.concat(processed_dfs)
+        else:
+            return IamDataFrame(pd.DataFrame([], columns=df.dimensions + ["value"]))
```
phackstock, line 393: If none of the region processing worked I would throw an error as that might lead to confusions along the lines of, "The region processing went through without any errors but I don't see any data".

```suggestion
            raise ValueError(The resulting data frame is empty please check the logs as to why that might be the case.)
```
Or something along those lines

phackstock (commented):
Looks good, just one minor comment

tests/test_core.py
danielhuppmann: No, the filtering applies to `test_df`. This is to use the same test-data for all different versions of this test. Removing this one value creates an inconsistent index.

nomenclature/processor/region.py
danielhuppmann: Good suggestion, implemented - but without reference to the log because I don't think anything meaningful is written to the log in most cases.

phackstock: Pefect, thanks.

PR #106, opened 2022-03-07
Fix handling of boolean attributes in codelists

danielhuppmann:
When working on the NAVIGATE template, I noticed that the nomenclature package does not correctly handle boolean attributes in the codelists.

After some investigation, I realized that pydantic casts booleans to string, unless ones uses the strict types.

This PR changes the codelist-types to the strict alternatives and updates the `to_yaml()` method to keep the order of the dictionaries.

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -1,13 +1,14 @@
-from re import compile, match
 from typing import Union, List, Dict
-from pydantic import BaseModel, validator
+from pydantic import BaseModel, StrictStr, StrictInt, StrictFloat, StrictBool
 
 
 class Code(BaseModel):
     """A simple class for a mapping of a "code" to its attributes"""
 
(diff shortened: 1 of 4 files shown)


phackstock (commented):
Looks good to me overall, one small remark regarding the testing.
I think with `test_codelist_to_yaml()` we don't actually test if the boolean was passed correctly.
Since it just checks for the string output, the value of `bool` could be `"true"` as well.
Using the same input data you use a basic test like this:
```python
code = CodeList.from_directory("variable", TEST_DATA_DIR / "simple_codelist")
assert code.bool is True
```

danielhuppmann:
Thanks @phackstock - good comment, explicit check added...

phackstock (approved):
Perfect, good to merge.

Issue #103, opened 2022-02-22
Check if resulting IamDataFrame is empty


phackstock:
In `processor.region.RegionProcessor.apply` we currently check if the list `processed_dfs` is empty and raise an error if it is.
In addition to this check we need to perform another check to see if the individual frames are longer than 0.
If all of them are 0 we raise an error.

phackstock:
Almost completely addressed by #99.

dc-almeida:
Closing because in the current implementation, models without a mapping are returned unchanged, and models with mappings that result in empty dataframes are checked via `merge_with_preaggregated_data`.

PR #107, opened 2022-03-08
Update error message for empty dataframe

phackstock:
I took a look at #103 and realized its pretty much taken care of with the merging of #99 as data frames are only added to `processed_dfs` if they are not empty.
I only found a slight inconsistency in the error message. In case `processed_dfs` is empty it would only report: "The region aggregation for model {model} resulted in an empty dataset" which can be misleading in case of multiple models.
I adjusted the error message and modified a unit test to test for that.


nomenclature/processor/region.py
```diff
@@ -408,7 +408,8 @@ def apply(self, df: IamDataFrame, dsd: DataStructureDefinition) -> IamDataFrame:
 
         if not processed_dfs:
             raise ValueError(
-                f"The region aggregation for model {model} resulted in an empty dataset"
+                f"The region aggregation for model(s) {df.model} resulted in an empty "
+                "dataset."
```
danielhuppmann, line 412: For consistency with the class **RegionProcessor** (if you agree, please also change in other parts of the logging)...

```suggestion
                f"The region-processing for model(s) {df.model} returned an empty dataset"
```

danielhuppmann (commented):
A few suggestions for improving consistency...

danielhuppmann (approved):
Looks good, thank you @phackstock

PR #109, opened 2022-03-15
Add a test for a variable with brackets in the name

danielhuppmann:
@strefler reported (via) email that variable names that include a bracket seem to cause problems during validation.

This PR adds a test to check that the validation works as expected (which it does, as far as I can tell)...


phackstock (approved):
Looks good to me.

Issue #108, opened 2022-03-11
Restructuring the documentation


danielhuppmann:
The current documentation sections "Getting Started" and "Usage" are not as useful or easy-to-understand as they should be.

Here is a suggestion for a new structure:

0. Remove "Installation" as own page
1. Change "Getting started" to have the following sections:
   - Installation: what previously was on the "Installation" page
   - Project-specific configuration: explain that each project has a configuration folder usually hosted on Github, explain that a user has to download/clone the GitHub repo to run the validation for a specific project locally
   - Data validation: One paragraph showing the CLI for `nomenclature validate-project .`
2. Change "Usage" to have the following *sub-pages*
   - Folder structure
   - Definitions
   - Mappings for region processing

The Usage pages should be aimed at novice Python users, so avoiding Python jargon. Also, "Notes" is probably not an idea header, because it does not tell a user what kind of notes to expect... Better to have descriptive sub-section titles.

danielhuppmann:
What do you think, @phackstock?

phackstock:
Thank you for the proposed outline @danielhuppmann, agree with all of the above. That structure makes a lot more sense.

PR #110, opened 2022-03-18
Unexpected region check

### LINKED ISSUE #102: Raise/warn if unexpected regions exist for region-processing
The **RegionProcessor** currently silently removes all regions that are not native regions (and once PR #99 is merged, common regions).

This can be confusing for users who successfully upload a data file to the Scenario Explorer but then do not find the removed regions in the Explorer.

Question:
1. fail loudly by raising an error if there are unexpected regions
2. write a warning to the log of the removed regions

Option 1 is more annoying for users, but Option 2 has the caveat that users usually do not look at the log file, in particular if the color is green and some data arrived in the database.

Preferences? @phackstock @Renato-Rodrigues @JohannesEmm @HauHe @robertpietzcker

phackstock:
closes #102.

## Overview

Implemented a check in `nomenclature/processor/region._check_unexpected_regions` which raises a ValueError in case the provided IamDataFrame contains regions which are not mentioned in one of the three sections:

* *native_regions* as the **original** region name.
* *common_regions* as the name of the region **or** one of the constituent regions. (I have a question about the second option, see below)
* *exclude_regions* as one of the entries in a list of strings.

In order to get the desired behavior, I added a parameter called `exclude_regions` in `RegionAggregationMapping`. There are checks that assure that a region is not present in *exclude_regions* and *native_regions* or *common_regions*.

I added two unit tests that test for errors raised in two different error scenarios:
1. Regions which are mentioned in *exclude_regions* as well as *native_regions* or *common_regions*. 
2. An unexpected region is present in the input data.

When it comes to testing the expected outcome without any error scenarios I just had to modify a few model mappings to include *exclude_regions* as there were already a few tests that contained unexpected regions. Since I only had to add *exclude_regions* and all expected results stayed the same I did not see the need to add another test case.

## Remarks and questions

* I added documentation about the new `exclude_region` attribute in the model mappings under the Usage#model-mapping section. This is not ideal but for now I'd not change it too much as I'll tackle the documentation as a whole (#108) next.
* I currently accept regions which are only mentioned as part of a common region. In a model mapping like this:

```yaml
model: model_a
common_regions:
  - common_region_1:
    - region_a
    - region_b
```

when provided data for *region_a* and *region_b* I would not raise an error since even though *region_a* and *region_b* **will not** be part of the processed data frame, they are used as constituent regions of *common_region_1*. @danielhuppmann do you think this a good way to go or should I raise an error as they are not part of the processed data frame?

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -3,6 +3,8 @@ linters:
     black:
     flake8:
         python: 3
-        ignore: W605
+        ignore: 
+            - W605
+            - W503
         max-line-length: 88
         max-complexity: 20


danielhuppmann:
The proposed solution implicitly allowing regions that are part of a common-region-list makes a lot of sense to me. (The alternative to include it as a component AND as an exclude-region would be confusing).

nomenclature/processor/region.py
```diff
@@ -459,3 +491,40 @@ def _merge_with_provided_data(
...
+            const_reg
+            for comm_reg in mapping.common_regions or []
+            for const_reg in comm_reg.constituent_regions
+        ]
+        + (mapping.exclude_regions or [])
+    ):
+        raise ValueError(
+            f"Did not find region(s) {regions_not_found} in 'native_regions', "
+            "'common_regions' or 'exclude_regions' in model mapping for "
+            f"{mapping.model} in {mapping.file}. If they are not meant for upload add "
```
danielhuppmann, line 527: I would not mention the “upload” because the package per se is independent from the IIASA Explorer infrastructure

phackstock: Ah right, good point I was thinking about it too much in terms of the Scenario Explorer.

phackstock: Updated it

phackstock:
It seems that in the case of line brakes before a binary operator black and flake8 disagree (see output of stickler).
Apparently it's supposed to be updated for flake8 in the near future (https://www.flake8rules.com/rules/W503.html). Should we add W503 it to the ignore section of `.stickler.yaml`?

danielhuppmann:
> 

Yes, sounds good! Maybe create an issue to remind us to clean that up in a while…

danielhuppmann (approved):
Looks good to me, thanks @phackstock

PR #113, opened 2022-04-13
Check if region-processing results are empty

### LINKED ISSUE #112: Raise error if region processing returns empty result
As it currently stands, the region processor assembles the final result of the processing by concatenating a list of `IamDataFrame` objects stored in `processed_dfs` (https://github.com/IAMconsortium/nomenclature/blob/main/nomenclature/processor/region.py#L447). 
Before that it is checked that the list `processed_dfs` is not empty, however the way the test is currently done checks only if the list contains at least one `IamDataFrame` object. This is not ideal since the object can be empty. 
In the extreme case `processed_dfs` contains only empty `IamDataFrame` objects which does not raise any errors but produces an unexpected result for the user while uploading. The upload appears 'green' in the scenario explorer UI while no data is added.

Adding a check as simple as:

```python
if all(df.empty for df in processed_dfs):
    raise ValueError("...")
```

should resolve the issue.

phackstock:
closes #112.

## New feature

Previously at the end of `Regionprocessor.apply` there was a check if the resulting list of `pyam.IamDataFrame` objects `processed_dfs` was non-empty. If it was in fact empty, an error is raised.
The PR now adds a explicit check to see if `processed_dfs` actually contains any non-empty `IamDataFrames`. In case they are all empty, the same error is raised.

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -345,8 +345,9 @@ class RegionProcessor(BaseModel):
         Raises
         ------
         ValueError
-            In case there are regions present in the input data which are not mentioned
-            in the corresponding model mapping.
+            * In case there are regions present in the input data which are not
+              mentioned in the corresponding model mapping.
+            * In case the region-processing results in an empty data frame.
         """
         processed_dfs: List[IamDataFrame] = []
         for model in df.model:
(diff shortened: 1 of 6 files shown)


nomenclature/processor/region.py
```diff
@@ -345,8 +345,9 @@ def apply(self, df: IamDataFrame, dsd: DataStructureDefinition) -> IamDataFrame:
         Raises
         ------
         ValueError
-            In case there are regions present in the input data which are not mentioned
-            in the corresponding model mapping.
+            * In case there are regions present in the input data which are not
+              mentioned in the corresponding model mapping.
```
danielhuppmann, line 349:
```suggestion
            * If *df* contains regions that are not listed in the model mapping, or
```

danielhuppmann (approved):
Looks good to me, one minor suggestion for a shorter formulation of the docstring...

PR #118, opened 2022-05-18
Add explicit tests for all supported Python versions

danielhuppmann:
And also add cross-references so that we remember to update the badges when supporting new Python versions (or dropping support for old versions).

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python 3.8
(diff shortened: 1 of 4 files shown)


phackstock (approved):
Looks all good to me. Just one question since we're talking about testing different versions. Should we also test MacOS and Windows for the operating system side?

danielhuppmann:
Briefly thought about it, but the list of dependencies is quite short (mostly pyam plus add-ons), so I wouldn't expect anything to not work across different operating systems.

phackstock:
Sure, but it's of course the unexpected stuff you catch using all those combinations.
Maybe to keep the testing time reasonable we could introduce nightly tests that test a full combination of python versions and operating systems.

PR #119, opened 2022-05-18
Migrate CLI-page to top-level docs

danielhuppmann:
The CLI docs are more like the API documentation, directed at expert-level users or developers, rather than the "general" users.


doc/source/cli.rst
```diff
@@ -0,0 +1,26 @@
...
+
+Standard usage
+--------------
+
+Run the following in a command line to ensure that a project folder is a valid
+configuration for the **nomenclature** package.
+
+.. code-block:: bash
+
+  nomenclature validate-project /folder/to/project/
```
phackstock, line 19: Although definitely better than my initial path placeholder maybe we just want to do `/project/folder`?
```suggestion
  nomenclature validate-project /project/folder/
```

phackstock (commented):
One small comment other than that, looks good.

danielhuppmann:
Thanks for the suggestion @phackstock

PR #122, opened 2022-05-30
Rewrite the pyam-references

danielhuppmann:
This PR rewrites the reference to the pyam package and adds it to the docs-pages hosted on RTD (previously, the reference was only included in the Readme).

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,7 @@ the IPCC and in many model comparison projects at the global and national level,
 including several Horizon 2020 & Horizon Europe projects.
 
 The validation and processing features of the **nomenclature** package
-work with scenario data in the form of a [**pyam.IamDataFrame**](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html)
-object as input.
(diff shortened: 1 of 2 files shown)


phackstock (approved):
Looks good to me, one really minor comment below. Even without that one, good to merge from my side.

README.md
```diff
@@ -34,22 +34,25 @@ The **nomenclature** package supports three main use cases:
...
+an open-source community toolbox for analysis & visualization of scenario data.
+The **pyam** package was developed to facilitate working with timeseries scenario data
+conforming to the format developed by the IAMC. It is used in ongoing assessments by
+the IPCC and in many model comparison projects at the global and national level,
+including several Horizon 2020 & Horizon Europe projects.
 
-[Read the Docs](https://pyam-iamc.readthedocs.io) for more information!
+The validation and processing features of the **nomenclature** package
+work with scenario data in the form of a [**pyam.IamDataFrame**](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html)
+class as input.
```
phackstock, line 53: Since we're not using the class itself but rather an instance of pyam.IamDataFrame, using `object` or `instance` would be more correct here. 
```suggestion
object as input.
```

PR #123, opened 2022-06-01
Single constituent region common region bypass

### LINKED ISSUE #116: Single constituent common region can lead to unexpected results
This is the result from a conversation with @pweigmann.
The partial region aggregation feature can lead to unexpected results if it is only comprised of a single common region, e.g.:

```yaml
model: model_a
common_regions:
  - Region_A:
    - region_a
```

this configuration is not super rare as it can often be the case that a region, that a model natively reports, `region_a` in the example, falls under the category of a comparison region. In this case, no aggregation is necessary, a renaming from `region_a` to `Region_A` is sufficient to ensure comparability with other models.

This is where unexpected behavior can occur as `common_regions` in essence means "perform region-aggregation". At the same time, region aggregation works on a principle we call "partial region-aggregation" (https://nomenclature-iamc.readthedocs.io/en/latest/usage.html#partial-region-aggregation) which means that nomenclature will perform the aggregation of the constituent regions but only use that for comparison. The results of the processing will be the model native results. 
However, this only works if the name of the **resulting** common region, in our case `Region_A` remains the same, i.e.:

```yaml
model: model_a
common_regions:
  - Region_A:
    - Region_A
```
if we have a renaming only the aggregated results will be used. 
This is normally not a problem as they _should_ be identical to the provided model native values. 
However, in a variable uses weighted aggregation such as `Price|Carbon` and the weight variable contains negative values we lose data points where the weight is negative.

There are a few options that this problem can be fixed/circumvented:
1. Disallow single constituent common regions, instead force them to be in the `native_regions` section.
2. Implement a check in the region aggregation to skip aggregation in case a common region is only comprised of a single region and only rename.

My preference would be option 1. as it is cleaner to implement, the region aggregation comparison feature does not really make sense if the region is only comprised of a single region, and there is another related issue with the current implementation:

Assuming a model natively reports two regions, "region_a" and "region_b". Assuming further that they should both be renamed to "Region_A" and "Region_B" respectively. Minus the potential issues with missing data, the following model mapping works:
```yaml
model: model_a
native_regions:
  - region_a: Region_A
common_regions:
  - Region_B:
    - region_b 
```
The inconsistency here is that for `native_regions` the value on the **right** is the one that will be in the result, For `common_regions` (here used for renaming only), on the other hand, the one on the **left/above** is the desired one. 
One way to solve this would be go with option 1. mentioned above and disallow single-constituent common regions. This would also be in line with the [zen of python](https://peps.python.org/pep-0020/) "There should be one-- and preferably only one --obvious way to do it.". In this case the renaming of model native regions could be done **only** by adding it to the `native_regions` list. 
If we go with this option we might also have to reconsider the naming of the two categories `native_regions` and `common_regions` as they would then more closely correspond to `do-not-aggreate` and `aggregate`.

@danielhuppmann, please let me know what you think.
Also @pweigmann, @orichters and @Renato-Rodrigues (you guys came to mind first as modellers that have current experience with using nomenclature, please feel free to loop in other colleages), your input is most welcome.

phackstock:
closes #116.

## Features added

For model mappings that look like this:

```yaml
model: model_a
common_regions:
  - region_A:
    - region_a
```

where a model native region corresponds exactly to a common region we effectively only want to do a renaming.

There was an issue previously where data was still aggregated and in case the aggregation failed data could be lost.
Now there is a check in place to see if a common region only consists of a single constituent region and if so, region processing is bypassed and only a renaming (if necessary) is performed.

I appended two existing unit tests to test or the correct behavior, `test_core.test_region_processing_rename` and `test_core.test_region_processing_skip_aggregation`.

Since this is more of a bugfix than a new feature I did not add anything to the docs but I can of course do that if that's deemed necessary.

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -416,6 +416,7 @@ class RegionProcessor(BaseModel):
                                     ).rename(region=cr.rename_dict)
                                 )
                                 continue
+                            # if there are multiple constituent regions, aggregate
                             regions = [cr.name, cr.constituent_regions]
                             # First, perform 'simple' aggregation (no arguments)
                             _processed_dfs.append(
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -72,6 +72,19 @@ class CommonRegion(BaseModel):
     name: str
     constituent_regions: List[str]
 
+    @property
+    def is_single_constituent_region(self):
+        return len(self.constituent_regions) == 1
+
+    @property
+    def rename_dict(self):
+        if self.is_single_constituent_region:
+            return {self.constituent_regions[0]: self.name}
+        else:
+            raise AttributeError(
+                "rename_dict is only available for single constituent regions"
+            )
+
 
 class RegionAggregationMapping(BaseModel):
     """Holds information for region processing on a per-model basis.
@@ -394,6 +407,20 @@ class RegionProcessor(BaseModel):
                             if var not in vars_default_args
                         }
                         for cr in self.mappings[model].common_regions:
+                            # If the common region is only comprised of a single model
+                            # native region, just rename
+                            if (
+                                cr.is_single_constituent_region
+                                and not model_df.filter(
+                                    region=cr.constituent_regions[0]
+                                ).empty
+                            ):
+                                _processed_dfs.append(
+                                    model_df.filter(
+                                        region=cr.constituent_regions[0]
+                                    ).rename(region=cr.rename_dict)
+                                )
+                                continue
                             regions = [cr.name, cr.constituent_regions]
                             # First, perform 'simple' aggregation (no arguments)
                             _processed_dfs.append(
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -409,12 +409,7 @@ class RegionProcessor(BaseModel):
                         for cr in self.mappings[model].common_regions:
                             # If the common region is only comprised of a single model
                             # native region, just rename
-                            if (
-                                cr.is_single_constituent_region
-                                and not model_df.filter(
-                                    region=cr.constituent_regions[0]
-                                ).empty
-                            ):
+                            if cr.is_single_constituent_region:
                                 _processed_dfs.append(
                                     model_df.filter(
                                         region=cr.constituent_regions[0]
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -11,7 +11,8 @@ from pyam import IAMC_IDX, IamDataFrame, assert_iamframe_equal
 from conftest import TEST_DATA_DIR
 
 
-def test_region_processing_rename():
+@pytest.mark.parametrize("model_name", ["model_a", "model_c"])
+def test_region_processing_rename(model_name):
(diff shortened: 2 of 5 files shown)


nomenclature/processor/region.py
```diff
@@ -394,6 +407,20 @@ def apply(self, df: IamDataFrame, dsd: DataStructureDefinition) -> IamDataFrame:
...
+                                and not model_df.filter(
+                                    region=cr.constituent_regions[0]
+                                ).empty
+                            ):
+                                _processed_dfs.append(
+                                    model_df.filter(
+                                        region=cr.constituent_regions[0]
+                                    ).rename(region=cr.rename_dict)
+                                )
+                                continue
```
danielhuppmann, line 423: Please add docstring/comment with "else" for clarity

tests/test_core.py
```diff
@@ -241,20 +242,36 @@ def test_region_processing_weighted_aggregation(folder, exp_df, args, caplog):
         assert logmsg in caplog.text
 
 
-def test_region_processing_skip_aggregation():
-    test_df = IamDataFrame(
+@pytest.mark.parametrize(
+    "model_name, region_names",
+    [("m_a", ("region_A", "region_B")), ("m_b", ("region_A", "region_b"))],
```
danielhuppmann, line 247: For consistency and easier readability
```suggestion
    [("model_a", ("region_A", "region_B")), ("model_b", ("region_A", "region_b"))],
```

danielhuppmann (approved):
Looks good, a few minor suggestions inline

nomenclature/processor/region.py
```diff
@@ -394,6 +407,20 @@ def apply(self, df: IamDataFrame, dsd: DataStructureDefinition) -> IamDataFrame:
                             if var not in vars_default_args
                         }
                         for cr in self.mappings[model].common_regions:
+                            # If the common region is only comprised of a single model
+                            # native region, just rename
+                            if (
```
danielhuppmann, line 412: Thinking about this again, I think it would be preferable to only check for `is_single_constituent_region` because
1. first filtering to check if the region exists, then filtering again as part of the renaming is less performant than doing the renaming directly
2. the could would be more readable

If this creates verbose log output, you could use the `adjust_log_level` context
https://github.com/IAMconsortium/pyam/blob/de6b47bb1ea7e937ad32dc8a6c284beddca98685/pyam/logging.py#L9

phackstock: The reason I did it is directly related to iamconsortium/pyam#631, if the data frame that we would rename is empty, we get an error.
I totally see your point that this way is quite inefficient so what I could do alternatively is tackle the pyam issue first an then come back to this one.

nomenclature/processor/region.py
phackstock: added

nomenclature/processor/region.py
phackstock: Thanks for the quick PR, removed the second condition after fixing the version requirement in setup.cfg

phackstock:
Addressed all the comments. If it's all good I'd go ahead with the merge then.

PR #125, opened 2022-06-10
Fix indentation of yaml examples

danielhuppmann:
Fixing the yaml-examples so that they are consistent with correct yaml syntax


phackstock (approved):
Looks good to me. Ready to be merged.

Issue #127, opened 2022-06-26
Support checking consistency across the variable hierarchy


danielhuppmann:
There is a use case for checking consistency across the variable hierarchy (or for computing variables higher-up in the hierarchy automatically from its components).

Aggregation along the variable hierarchy and checking consistency is already supported by the pyam methods [`aggregate()`](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html#pyam.IamDataFrame.aggregate) and [`check_aggregate()`](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html#pyam.IamDataFrame.check_aggregate).

I propose to extend the variable-definitions for a nomenclature-compatible variable codelist by a two "standard" attributes:
- **components**: the list of variables that are components of the variable
- **check-aggregate**: takes a boolean on whether to perform the aggregate-check as part of the [`process()`](https://nomenclature-iamc.readthedocs.io/en/stable/api.html#nomenclature.process) function

An example of the configuration:
```yaml
- Emissions|CO2|Energy and Industrial Processes:
    description: CO2 emissions from the energy and industry sectors
    unit: Mt CO2
    components:
      - Emissions|CO2|Energy
      - Emissions|CO2|Emissions|CO2|Industrial Processes 
    check-aggregate: true
    <other attribute>: Some text, value, boolean or list (optional)
```

*Note*: the **pyam** methods `aggregate()` and `check_aggregate()` interpret `components=None` as "all variables one level below the variable to be aggregated" and automatically detects all these variables.

*Note: Computing aggregates will be tackled as a subsequent issue.*

@volker-krey @christophbertram @lolow

christophbertram:
It would be good to design this from the start with the functionality of adding more than one list of components, e.g. for Final Energy having both the components 
> Final Energy|Transport
> Final Energy|Buildings
> Final Energy|Industry
as well as the alternative components
> Final Energy|Electricity
> Final Energy|Liquids
> Final Energy|Gases
> Final Energy|Solids
> Final Energy|Other
The check-aggregate would potentially also be activated only for one list.

lolow:
Should it be done using 1 unique entry for the variable or using multiple:

unique entry:
```yaml
- Final Energy:
    description: Total final energy consumption by all end-use sectors and all fuels,
      excluding transmission/distribution losses
    unit: [EJ/yr, GWh]
    components: 1
      subcomponents:
        - Final Energy|Transport
[...]

PR #130, opened 2022-06-29
Add allowed attributes to variable schema

danielhuppmann:
This PR adds the to-be-allowed attributes for the aggregation-check along the variable tree to the variable-code list schema (see #127). This needs to be added before continuing to work on the actual implementation so that PRs on actual projects do not fail the nomenclature-validation.

### DIFF
--- a/nomenclature/validation_schemas/variable_schema.yaml
+++ b/nomenclature/validation_schemas/variable_schema.yaml
@@ -32,6 +32,14 @@ definitions:
                   type: [ string, number, boolean, "null" ]
               minProperties: 1
               maxProperties: 1
+          skip-region-aggregation:
+            type: boolean
+          check-aggregate:
+            type: boolean
+          components:
+            type: array
+            items:
+              type: string
         required:
           - unit
         additionalProperties:


phackstock:
Just took a quick look and I was wondering about the addition of `skip-region-aggregation`.
We've had this in the past and it has worked fine right? This should be covered by the allowed properties, here:

```json
additionalProperties:
type: [ string, number, boolean, "null" ]
```
the one limitation I would see with that it does not mention `array`. But I think if I get this right adding `array` to the list of allowed types for additional properties should be enough.

phackstock:
Sorry for closing the PR, it seems that GH changed the meaning of control+enter to close the PR ...
That really annoying now I have to click "Comment" again using the mouse...

danielhuppmann:
Right, I added **skip-region-aggregation** to be explicit about expected attributes and use the schema-handling to make sure that only the right types end up there.

Therefore, I would rather not simply add array to the generic attributes to have more control of the type of arrays - as a first step, only a list of strings is allowed for the **components**.

danielhuppmann:
> Alright, then it's fine with me to do it that way. Internally it ends up being mapped to a dictionary right? So no need to add any attributes to any classes?

Correct.

phackstock (approved):
As per the discussion above, good to merge from my side. Maybe with the addition to the docs once this feature is fully implemented.

PR #131, opened 2022-07-01
Enable aggregation-checks along the variable dimensions

### LINKED ISSUE #127: Support checking consistency across the variable hierarchy

danielhuppmann:
This PR implements a method `check_aggregate()` to validate consistency of scenario data along the variable hierarchy, i.e., that the values of *Final Energy* match the sum of all components. Closes #127 

The feature has to be activated by adding `check-aggregate: true` to all variables that should be validated.

The feature has three modes:
- default: use all sub-categories of a variable
- specify a list of components explicitly
- specify several lists of components to allow checking along different dimensions (sectors vs. fuels)

The behavior is as follows:
- `check_aggregate()` returns None if the validation succeeds or a pandas.DataFrame of all inconsistent data
- `process()` raises an error if the validation fails and writes the DataFrame to the log (for use as part of a scenario-explorer upload)

fyi @volker-krey @christophbertram @lolow @robertpietzcker @Renato-Rodrigues


nomenclature/codelist.py
```diff
@@ -100,6 +100,25 @@ def check_variable_region_aggregation_args(cls, values):
                     )
         return values
 
+    @root_validator(pre=False, skip_on_failure=True)
+    def cast_variable_components_args(cls, values):
+        """Cast "components" list of dicts to a codelist"""
+        if values["name"] == "variable":
+            items = [
```
phackstock, line 107: Alternatively, you could iterate over `values["mapping"].items()` once, and check for each loop iteration if the attributes contain `"components"` instead of creating the items variable and iterating over it again.
Might be more easily understood this way though and the performance difference should be negligible.

nomenclature/definition.py
```diff
@@ -60,6 +62,59 @@ def validate(self, df: IamDataFrame, dimensions: list = None) -> None:
...
         validate(self, df, dimensions=dimensions or self.dimensions)
 
+    def check_aggregate(self, df: IamDataFrame, **kwargs) -> None:
+        """Check for consistency of scenario data along the variable hierarchy
+
+        Parameters
+        ----------
+        df : :class:`pyam.IamDataFrame`
+            Scenario data to be checked for consistency along the variable hierarchy.
+        kwargs : Tolerance arguments for comparison of values
```
phackstock, line 72: I took a quick look at numpy.isclose and since there's only a total of three possible variables, maybe it would be more clear to the user to directly accept those instead of using `kwargs` here. 
Maybe `kwargs` is the way to go though as it preserves the numpy.isclose defaults. In this case perhaps listing the accepted variables in the docstring could be nice.

nomenclature/definition.py
```diff
@@ -60,6 +62,59 @@ def validate(self, df: IamDataFrame, dimensions: list = None) -> None:
...
+        ValueError
+            If the :any:`DataStructureDefinition` does not have a *variable* dimension.
+        """
+        if "variable" not in self.dimensions:
+            raise ValueError("Aggregation check requires 'variable' dimension.")
+
+        lst = []
+
+        with adjust_log_level(level="WARNING"):
+            for code, attr in self.variable.items():
```
phackstock, line 91: Would it be more efficient to iterate only over the variables which actually are present in the input dataframe instead of the whole variable list?

phackstock (approved):
Looks all pretty good to me. Some minor comments inline but even without those it looks good to merge from my side.
Two more safety features that I was thinking of while reviewing that might fit into this or a future PR:
* Check if all variables mentioned in `components` exist in the code list. Otherwise we could get unexpected errors or aggregates not lining up. This one might also be worth implementing for the `weight` attribute, i.e. ensure that all variables that are used as weights are also present as their own variable entries.
* Print a waning if a variable entry has `components` but no `check-aggregate: true`. This is to prevent from accidentally forgoing the comparison by not explicitly setting `check-aggregate: true`. If `components` are provided we can assume that the user's intent is to compare and issue them a warning in case they forgot to turn on the comparison.

nomenclature/definition.py
danielhuppmann: I think having the link to the numpy-docs is preferable rather than repeating specific information.

nomenclature/codelist.py
danielhuppmann: Good suggestion, implemented.

nomenclature/definition.py
danielhuppmann: Good suggestion, implemented.

nomenclature/codelist.py
danielhuppmann: I was worried that this would create a concurrent-edit-issue...

danielhuppmann:
Thanks for the review and suggestions, @phackstock. I created an issue for the internal-consistency check that all components must be defined as variabbles (#134), maybe you can implement that after merging the related PR.

About printing a warning if **components** are defined but the **aggregation-check** is not activated: I'd rather wait and see how this use case develops...

PR #133, opened 2022-07-05
Check that weight variables exist

### LINKED ISSUE #132: Check that weight variables exist
We want to verify that the variable values provided as a `weight` argument for weighted aggregation are present in the variable list as their own entries.
For example a variable defined like this:

```yaml
- Food Demand:
    unit: kcal/cap/day
    weight: Population
```

would need a entry for `Population`, e.g.:

```
- ...
- Population:
    unit: million
- ...
```

Currently, as far as I'm aware, this is not being checked.

phackstock:
Closes #132.

## Features added

The validation of the variable code list now includes a check that all variables mentioned as a `weight` also have their own entry in the list. I've written the validator so that it collects **all** missing weight variables and does not throw an error for the first one. This way multiple errors can be fixed in one go.
I added a point in the documentation to highlight that the weight is required. 
I also took the liberty to slightly refactor the codelist validators. The main change there is the switch from using `root_validator()` to `validator("mapping")`.
The reason for this change is that in the validators only modify and validate the `mapping` attribute. The `name` attribute is mainly used for better error reporting in this case. The way how to access multiple attributes in a non-root validator is adding `v` and `values` to the list of input variables (details can be found under the code example of "Validators" https://pydantic-docs.helpmanual.io/usage/validators/). As an example:

```
@validator("mapping")
def some_validator(cls, v, values):
    # v is the field to be validated "mapping" in our case
    # values is a dict, that contains all other previously validated fields
```

If you don't like the switch from the `root_validator` though I can also revert it.

### DIFF
--- a/doc/source/user_guide/codelist.rst
+++ b/doc/source/user_guide/codelist.rst
@@ -97,10 +97,9 @@ package. See the section :ref:`model_mapping` for more information.
   :meth:`aggregate_region() <pyam.IamDataFrame.aggregate_region>` will
   be passed to that method. Examples include *method* and *weight*.
 
-* The *weight* attribute specifies a second variable to be used as a weight for
-  computing the aggregation as a weighted average. The variable given in the *weight*
-  attribute **must** have its own entry in the variable list, otherwise an error will be
-  raised.
+* The *weight* attribute is optional. If provided, this variable will be used as a weight for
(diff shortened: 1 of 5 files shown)


danielhuppmann (approved):
A nice improvement, thanks! Just a few suggestions for shorter, more readable documentation and error messages.

PR #136, opened 2022-07-07
Cast `components`-attribute to dict instead of CodeList

### LINKED ISSUE #135: Components list violates CodeList definition
As I was trying to get the new component check tests working on #133 I found a somewhat sneaky error/inconsistency in the way that `CodeList` is using `CodeList` recursively for specifying a list of components as a dictionary. 

## Description

As it stands, in `cast_variable_components_args` (https://github.com/IAMconsortium/nomenclature/blob/main/nomenclature/codelist.py#L104), if we have a list of different aggregation dimensions, we cast them into a `CodeList`:
```python
...
# translate a list of single-key dictionaries to a simple dictionary
            for (name, attrs) in items:
                if not all([isstr(i) for i in attrs["components"]]):
                    values["mapping"][name]["components"] = CodeList(
                        name="components", mapping=attrs["components"]
                    )
```

This means that the value for the "components" attribute in a "variable" codelist can be of type `CodeList`.
This, however clashes with the type definition in the `CodeList` header for the `mapping` attribute (https://github.com/IAMconsortium/nomenclature/blob/main/nomenclature/codelist.py#L53):

```python
 mapping: Union[List, Dict[str, Dict[str, Union[str, float, int, list, None]]]] = {}
```
in the final `Union` which specifies what the final data types are allowed to be we would need an additional `CodeList` to make it work.

The reason why nothing broke despite `pydantic` enforcing types is because are checked **before** the validators are run (except when `pre=True`). The validators are then run under the assumption that they don't recast the type of the variable but only check if it conforms to whatever checks there are in place. 
As I found out today, however, there's nothing stopping you from changing the data type and therefore rendering the entire typing checking pointless. 
A minimum example to illustrate:

```python
from pydantic import BaseModel, root_validator

class A(BaseModel):

    a: int

    @root_validator()
    def change_a(cls, values):
        # convert to list 
        values["a"] = [values["a"]]
        return values

print(A(a=10).a) # prints out [10] instead of the expected 10!
```

This is the exact same thing that is currently happening in the `CodeList` model.
In my opinion this is bad because now `CodeList.variable["some variable"]["components"]` can not only contain a list or a dictionary as the definition would suggest but actually also another `CodeList`.

## Proposed fixes

I would see two fixes for this:
1. Add `CodeList` to the `Union` of allowed types in the final level of the `mapping` attribute. 
2. Change it so that `cast_variable_components_args` does not cast to a `CodeList` but rather directly to a dictionary as the original docstring in `cast_variable_components_args` suggests.

My preference would be option 2 because in option 1 we are opening the door for a recursive `CodeList` structure. This might be desired at some point in the future but I always have troubles with this kind of complexity. 
Option 2 on the other hand I find more simple to visualize and it should be good enough for our purposes. That is of course, unless, we want some kind of validation which `CodeList` conveniently provides for us. I've though about it quickly but apart from making sure that the keys in `"components"` are unique I could not think of anything that would not be covered by our yaml/json schema specification already. That being said if it's only about ensuring that the keys are unique I would just write a quick check before casting to a dict instead of using a whole new instance of `CodeList`.

In any case I would suggest that it might be a good idea to move any kind of re-casting for types to a validator (`root_validator` or normal `validator`) with `pre=True`. All other validators are then only allowed to verify the integrity of the data and not change it. This way we would ensure that the types specified in the class definition of `CodeList` are the types an instance of `CodeList` actually ends up having. 

This one got a bit more lengthy than expected, would love to hear your thoughts @danielhuppmann.

danielhuppmann:
Closes #135.

This PR (still) uses the CodeList-casting to parse the structure of the **components** attribute as a list of single-key-dicts to a dictionary, but then extracts the dictionary so that the actual **components** attribute is really just a simple dictionary. I think that this is an elegant solution to avoid nested CodeList structure while still re-using the parsing and validation of the CodeList-initialization. And I added a test...

### DIFF
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -115,7 +115,7 @@ class CodeList(BaseModel):
                 if not all([isstr(i) for i in attrs["components"]]):
                     values["mapping"][name]["components"] = CodeList(
                         name="components", mapping=attrs["components"]
-                    )
+                    ).mapping
 
         return values
 
--- a/nomenclature/definition.py
+++ b/nomenclature/definition.py
@@ -94,7 +94,7 @@ class DataStructureDefinition:
                     components = attr.get("components", None)
 
                     # check if multiple lists of components are given for a code
-                    if isinstance(components, CodeList):
+                    if isinstance(components, dict):
                         for name, _components in components.items():
                             error = df.check_aggregate(code, _components, **kwargs)
                             if error is not None:
--- a/tests/test_check_aggregate.py
+++ b/tests/test_check_aggregate.py
@@ -37,11 +37,20 @@ def expected_fail_return(name):
     return pd.DataFrame([[9.0, 10.0], [8.0, 7.0]], columns=columns, index=index)
 
 
-@pytest.mark.parametrize("components", ["components", "components_dict"])
-def test_check_aggregate_passing(components):
+@pytest.mark.parametrize(
+    "components, components_type",
+    [
+        ("components", list),
+        ("components_dict", dict),
+    ],
+)
+def test_check_aggregate_passing(components, components_type):
     """Assert that the aggregate-check passes with different types of components"""
 
     dsd = DataStructureDefinition(TEST_DATA_DIR / "check_aggregate" / components)
+
+    # check that components is returned as a basic type (not a codelist)
+    assert isinstance(dsd.variable["Final Energy"]["components"], components_type)
 
     # aggregation check returns None if no inconsistencies are found
     assert dsd.check_aggregate(TEST_DF) is None


phackstock (approved):
Thanks for the quick PR, looks good to me.
My personal preference would be still to not use `CodeList` within `CodeList` but I think that's a discussion to be had another time. In that discussion I would also propose the introduction of dedicated classes for "region" and "variable" `Codelist` entries. 
As we keep on adding more functionality, especially for variables, the generic `CodeList` class is becoming a bit bloated and hard to read in my opinion, having a dedicated class would be the more elegant solution.

danielhuppmann:
Agree that a refactoring is in order before adding yet more features... But not today.

PR #143, opened 2022-08-01
Add howto

### LINKED ISSUE #142: Add How-To section to the docs
As we are starting to roll out nomenclature to a number of projects, users are adding their model mappings.
One common pitfall that I've seen is that while a model mapping is added, the corresponding region definition is not. 
This causes the CI to fail and requires assistance from our side to fix. 
This could be a good start to a How-To/Quickstart guide so that users can help themselves.
Thanks @khaeru for the useful suggestion.

phackstock:
closes #142.

This is my first attempt at a "how to" detailing the common pitfall of adding a model mapping without the corresponding region definition files.
As the "how to" referenced the `nomenclature validate-project` cli command, I added some additional docstring explanation that does not seem to be rendered currently by sphinx-click. Additionally, the direct referencing of a cli command seems to not be possible at the moment (https://sphinx-click.readthedocs.io/en/latest/usage/#cross-referencing) so I liked the CLI page.

### DIFF
--- a/doc/source/user_guide.rst
+++ b/doc/source/user_guide.rst
@@ -47,4 +47,5 @@ the RegionProcessor and validated using DataStructureDefinition.
   user_guide/variable
   user_guide/region
   user_guide/model-mapping
+  user_guide/howto
   user_guide/local-usage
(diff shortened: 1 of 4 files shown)


phackstock:
Closing this PR as docs can only be rendered from a branch within the original repo.

PR #147, opened 2022-08-10
Features/add cli option

### LINKED ISSUE #71: Add "dimensions" arg to validation CLI
Projects/repositories will use non-default dimensions in their "definitions", so it would be helpful to add this as an optional argument to `nomenclature.testing.assert_valid_structure()` and also enable passing this via the `validate-project` CLI.

Related to #68.

luciecastella:
Closes #71 

This adds a features allowing users to optionally type in the dimensions to check non-default dimensions in the validation of a project. 
This is currently done with (edit) : --dimensions "['variable', 'region', 'foo']"


luciecastella:
For now it works with --dimensions=variable,region,foo
But I could also try multiple entries as Philip suggested, I do not know which one is the more elegant

danielhuppmann:
Tried to run this, but it seems that tests are now failing if a `dimensions` argument is not given (so this argument is now treated as required, not optional). Also, the test-definitions look fine but are not actually executed yet in any test.

About the approach: 
I see a risk of unintuitive behavior when users add a whitespace (which one would do when following pep8-style), so the following would fail
```
nomenclature validate-project . --dimensions=foo, variable
```

I still think that something like 
```
nomenclature validate-project . --dimensions="['foo', 'variable']"
```
would be preferable.
Maybe the following could be helpful? https://stackoverflow.com/a/47730333

danielhuppmann:
> Am I the only one who does not know that?

I didn't know that either, but if there are some examples (in the docs and the tests), it should be relatively straightforward for users to "automatically" use the right structure...

luciecastella:
Yes OK I will do that

luciecastella:
Hello! 
New problem: I have issues giving the new cli arguments in the CliRunner invoke. 
This is what I did
image
And result_valid ends up with an exit_code 2 even though the apparently same command line creates no error on my terminal. Any clue on how to write the invoke? I tried different things and looked online but I cannot figure out how it could be right

tests/test_cli.py
```diff
@@ -76,3 +76,20 @@ def print_helper(_path):
 
     with pytest.raises(NotADirectoryError, match=print_helper(path)):
         assert_valid_structure(TEST_DATA_DIR / "invalid_yaml")
+
+def test_cli_custom_dimensions_quotes():
+    """Check that CLI expected simple quotes (') inside double quotes (")"""
+    result_valid = runner.invoke(
+        cli, ["validate-project", 
+            str(TEST_DATA_DIR / "non-default_dimensions_passing"), 
+            "--dimensions \"['variable', 'region', 'foo']\""
```
phackstock, line 85: Depending on how click constructs the argument, I think this might fix your issue:
```suggestion
            "--dimensions", "\"['variable', 'region', 'foo']\""
```
i.e. having `--dimensions` and the value of the `dimensions` option as two different string entries in the list of commands.

phackstock: Unfortunately the click documentation is not really detailed (https://click.palletsprojects.com/en/8.1.x/testing/) but I would imagine that it works in analogy to how the python standard library does it when you invoke the command line from python (https://docs.python.org/3/library/subprocess.html#popen-constructor). 
In the example given in the link, in order to execute:
```console
/usr/bin/git commit -m "Fixes a bug."
```
from a python script you'd have to use:
```python 
Popen(["/usr/bin/git", "commit", "-m", "Fixes a bug."])
```
The important point is that `-m ` and `"Fixes a bug"` are two separate entries in the list. I think you'll have to do this in analogy to your `--dimensions` and `['variable', ...]`.

luciecastella: OK thank you! I does not work yet but I will try :)

nomenclature/testing.py
```diff
@@ -28,9 +29,13 @@ def assert_valid_yaml(path: Path):
 
 
 def assert_valid_structure(
-    path: Path, mappings: str = "mappings", definitions: str = "definitions"
+    path: Path,
+    mappings: str = "mappings",
+    definitions: str = "definitions",
+    dimensions: List[str] = ["region", "variable"],
```
phackstock, line 35: Since you added a new parameter, it would need an entry in the doc string below.

tests/data/non-default_dimensions_passing/definitions/region/regions.yaml
```diff
@@ -0,0 +1,8 @@
+- common:
+    - World
+- country:
+    - Austria:
+        iso2: AT
```
phackstock, line 5: Again those attributes are not necessary for the tests.

tests/test_cli.py
```diff
@@ -119,3 +119,48 @@ def test_cli_wrong_definitions_name():
...
+def test_cli_custom_dimensions():
+    """Check that CLI runs through with a non-default dimension"""
+
+    result_valid = runner.invoke(
+        cli,
+        [
+            "validate-project",
+            str(TEST_DATA_DIR / "non-default_dimensions_passing"),
+            "--dimensions",
+            "['variable', 'region', 'foo']",
```
phackstock, line 133: Using a folder called `"foo"` is fine in principle just one thing to note about the functionality of `nomenclature` in general.
The dimension that is validated is determined by the name of the **folder**, this means that in your case even though the folder `foo` contains a file called `scenarios.yaml` this would be used to validate a, usually non-existent, column called `foo`. 
For the tests it's fine to keep it this way though.

tests/test_cli.py
```diff
@@ -119,3 +119,48 @@ def test_cli_wrong_definitions_name():
...
+        ],
+    )
+    assert result_valid.exit_code == 0
+
+
+def test_cli_custom_dimensions_empty():
+    """Check that CLI raises an expected error with an empty non-default dimension"""
```
phackstock, line 140: This could be a bit more clear, something like:
```suggestion
    """Check that CLI raises an error when specifying a non-existent directory ('foo')"""
```

tests/test_cli.py
```diff
@@ -119,3 +119,48 @@ def test_cli_wrong_definitions_name():
...
+            str(TEST_DATA_DIR / "non-default_dimensions_failing"),
+            "--dimensions",
+            "['variable', 'region', 'foo']",
+        ],
+    )
+    assert result_valid.exit_code == 1
+
+
+def test_cli_custom_dimensions_fails():
+    """Check that CLI raises expected error when given a non-existing dimension"""
```
phackstock, line 155: What is the difference between this test and the previous one? The way I read it is that in both cases you specify a non-existing dimension. 
This is speculation now, but could it be that for the previous example you have created a folder called `foo/` and it's empty? In this case this would explain the difference and since git only tracks files and not empty folders the folder does not show up here on GitHub. If this is indeed the case and you want to test for an empty dimensions one way around this is adding an empty file called `.gitkeep` to the folder `foo/`. See here for details https://stackoverflow.com/questions/7229885/what-are-the-differences-between-gitignore-and-gitkeep.

phackstock (changes requested):
Looks pretty good overall. Some comments below in line.
One more general comment about testing the failure states of the cli is that maybe we want to test for more than just the exit code. We want to assure that the cli fails in the way intended and not just that it fails in any way. What do you think @danielhuppmann, should this be added here or should this be subject of a possible follow-up PR?

tests/test_cli.py
danielhuppmann: I think you raise a good point @phackstock - let us use names for "passing" tests that are similar to what is expected by users (ie, IAMC columns, region, variable, scenario) and use "foo" for failing tests, non-existing folders, etc.

luciecastella: Good to know thank you

nomenclature/testing.py
```diff
@@ -28,9 +29,13 @@ def assert_valid_yaml(path: Path):
 
 
 def assert_valid_structure(
-    path: Path, mappings: str = "mappings", definitions: str = "definitions"
+    path: Path,
```
danielhuppmann, line 32: The order of arguments is not consistent with the CLI `validate-project` - and I think it will be way more common to use the "dimensions" options than custom folder names, so please order as dimensions-first...

danielhuppmann:
> We want to assure that the cli fails in the way intended and not just that it fails in any way. What do you think @danielhuppmann, should this be added here or should this be subject of a possible follow-up PR?

Agree with you @phackstock in principle, but let's do that as a follow-up PR.

tests/test_cli.py
luciecastella: Thanks! I think I actually messed up the names so maybe that's my fault there was no folder, but yes I can do that anyways. And yes, your assumptions were correct, I changed it so that it is more comprehensible.

phackstock (approved):
Thanks for addressing all the comments @luciecastella.

nomenclature/cli.py
```diff
@@ -15,9 +25,24 @@ def cli_valid_yaml(path: Path):
 
 @cli.command("validate-project")
 @click.argument("path", type=click.Path(exists=True, path_type=Path))
-@click.option("--mappings", type=str, default="mappings")
-@click.option("--definitions", type=str, default="definitions")
-def cli_valid_project(path: Path, mappings: str, definitions: str):
+@click.option(
```
danielhuppmann, line 28: Please align the ordering of the arguments.

danielhuppmann (approved):
Thanks @luciecastella, nice work!

PR #149, opened 2022-08-11
Fix 2 typos : model-mapping

luciecastella:


phackstock (approved):
Thanks for spotting those @luciecastella :+1: 

phackstock:
Ah right, then I'll do it for now.

PR #150, opened 2022-08-11
Change path representation to posix style

### LINKED ISSUE #148: Plateform dependant tests' success
It seems that some tests (test_codelist_to_yalm and test_definition_with_custom_dimensions) do not work on Windows : the tests both check paths and the reference seems to have a different convention as the actual path on Windows. 
See images below

image

image

phackstock:
closes #148.

As per discussion in #148 changing the file path in `CodeList` entries to be represented as a posix style path.
For the first CI run I changed nothing and only set the OS to `windows-latest` to reproduce the error.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -24,7 +24,7 @@ jobs:
 
       fail-fast: false
 
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     name: py${{ matrix.python-version }}
 
     steps:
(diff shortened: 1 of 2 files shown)


phackstock:
As illustrated in https://github.com/IAMconsortium/nomenclature/runs/7782738143?check_suite_focus=true, the error is reproducible. Now for the fix.

phackstock:
Since the change seems to have fixed the tests on Windows, I'm reverting the tests back to `ubuntu-latest`.

danielhuppmann (approved):
Nice, thanks @phackstock!

luciecastella (approved):

PR #151, opened 2022-08-11
Add nightly run with multiple OS

phackstock:
As discussed yesterday in #148 running integration tests for multiple operating systems for every PR is overkill.

I still think that running them once per day as a nightly run on the main branch can be useful to catch operating system dependent bugs early. I slightly modified `pytest.yaml` so that it now also runs every day at 3am UTC. In addition, only for these scheduled nightly runs, we also run them on `macos-latest` and `windows-latest`.

@danielhuppmann if you think it's not a good idea, I can also close the PR.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -10,7 +10,7 @@ on:
   pull_request:
     branches: [ '**' ]
   schedule:
-    cron: '0 3 * * *'
+    - cron: '0 3 * * *'
 jobs:
   pytest:
 
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -9,8 +9,6 @@ on:
     branches: [ 'main' ]
   pull_request:
     branches: [ '**' ]
-  schedule:
-    - cron: '0 3 * * 7'
 jobs:
   pytest:
 
@@ -22,7 +20,6 @@ jobs:
         os: [ubuntu, macos, windows]
       fail-fast: false
     # for non-nightly run only ubuntu, for nightly run all three OS
-    if: ${{ (matrix.os == 'ubuntu') || (github.event_name == 'schedule') }}
     runs-on: ${{ matrix.os }}-latest
     name: 'OS: ${{matrix.os}} py${{ matrix.python-version }}'
     steps:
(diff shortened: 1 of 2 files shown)


danielhuppmann:
Thanks @phackstock! I would actually flip the test strategy around:
- do the nightly tests only on ubuntu (and only once a week)
- do the tests on PRs on all OSs

This is safer (because merging to the `main` branch is tested on all OSs) and it is also less resource-intensive (because nightly tests will likely run more often than PRs).

phackstock:
Alright, sounds good as well. Should we do all three versions of python in every case? So running nine sets of tests for every PR and three sets for the nightly ones?
Also it seems that unfortunately the if clause I put into the workflow is evaluated before the matrix expression and so it's not working right now. I'll see that I fix it.

danielhuppmann:
Maybe easier to have separate workflows for the PRs and the nightly tests... Also easier for new users to understand what's happening, instead of having to work through some if-statements in the workflow.

phackstock:
Yes, you're right. I was trying to be fancy and have it all in one so that any future changes are applied to all tests by default.
Seems like what I'm trying to do is not possible by default on GitHub actions (https://github.com/orgs/community/discussions/26253). Looks like there is a workaround but it's even more complicated than what I have now so I'll just create a separate nightly file. 
Coming back to the question from above, do you think we should do 3 (python versions) x 3 (OS types) = 9 test runs for PRs?

danielhuppmann:
> do you think we should do 3 (python versions) x 3 (OS types) = 9 test runs for PRs?

Yes, let's go for that...

phackstock:
Perfect, I modified the standard tests to run for all OS and added a nightly test run which runs on Sunday for ubuntu with all three python versions.

.github/workflows/pytest.yml
```diff
@@ -9,24 +9,19 @@ on:
...
-        - '3.8'
-
+        python-version: ['3.8', '3.9', '3.10']        
+        os: [ubuntu, macos, windows]
       fail-fast: false
-
-    runs-on: ubuntu-latest
-    name: py${{ matrix.python-version }}
-
+    # for non-nightly run only ubuntu, for nightly run all three OS
```
danielhuppmann, line 22: This is not correct (any more)...

phackstock: Ah sorry for missing that, removed it.

danielhuppmann (approved):

phackstock:
You're welcome @danielhuppmann :) always nice to tighten up our testing and thanks to @luciecastella for starting this issue.

PR #153, opened 2022-08-16
Features/add cli folders

### LINKED ISSUE #68: Add definitions and mappings folder options to the cli
Following the discussion in #64, it emerged that the CLI currently looks for folders called "mappings" and "definitions" in `cli.cli_valid_project`.
While these two names are a good standard they should be able to be modified as we allow the same when instantiating `RegionProcessor` and `DataStructureDefinition`.

luciecastella:
Closes #68 

The user now can type: validate-project . --definitions=def --mappings=map 
For non-default folder names


nomenclature/testing.py
```diff
@@ -27,7 +27,7 @@ def assert_valid_yaml(path: Path):
         )
 
 
-def assert_valid_structure(path: Path):
+def assert_valid_structure(path: Path, mappings, definitions):
```
danielhuppmann, line 30: Please make these arguments optional (with current usage as default) so that it is possible to use the function directly (without the CLI) as before.

tests/data/non-default_folders/def/region/regions.yaml
```diff
@@ -0,0 +1,8 @@
+- common:
+    - World
+- country:
+    - Austria:
+        iso2: AT
```
phackstock, line 5: While technically correct, these lines aren't strictly needed for the unit tests to work so I would remove them and just leave the `Austria` entry.

nomenclature/testing.py
```diff
@@ -36,11 +36,14 @@ def assert_valid_structure(path: Path):
...
       - If a `mappings` folder exists, it must be a valid :class:`RegionProcessor`
     """
-    definition = nomenclature.DataStructureDefinition(path / "definitions")
-    if (path / "mappings").is_dir():
+
+    definition = nomenclature.DataStructureDefinition(path / str(definitions))
+    if (path / str(mappings)).is_dir():
         nomenclature.RegionProcessor.from_directory(
-            path / "mappings"
+            path / str(mappings)
```
phackstock, line 43: Since we expect a string input anyway I don't think this is needed
```suggestion
            path / mappings
```
should be enough

nomenclature/testing.py
```diff
@@ -27,7 +27,7 @@ def assert_valid_yaml(path: Path):
         )
 
 
-def assert_valid_structure(path: Path):
+def assert_valid_structure(path: Path, mappings="mappings", definitions="definitions"):
```
phackstock, line 30: Some type hints would be nice:
```suggestion
def assert_valid_structure(path: Path, mappings: str ="mappings", definitions: str="definitions") -> None:
```

nomenclature/cli.py
```diff
@@ -15,7 +15,9 @@ def cli_valid_yaml(path: Path):
 
 @cli.command("validate-project")
 @click.argument("path", type=click.Path(exists=True, path_type=Path))
-def cli_valid_project(path: Path):
+@click.option("--mappings", default="mappings")
```
phackstock, line 18: Although technically redundant it's always nice to tell the users the exact data type you expect. 
```suggestion
@click.option("--mappings", type=str, default="mappings")
```

nomenclature/cli.py
```diff
@@ -15,7 +15,9 @@ def cli_valid_yaml(path: Path):
...
+@click.option("--definitions", default="definitions")
```
phackstock, line 19:
```suggestion
@click.option("--definitions", type=str, default="definitions")
```

nomenclature/cli.py
```diff
@@ -15,7 +15,9 @@ def cli_valid_yaml(path: Path):
...
+def cli_valid_project(path: Path, mappings, definitions):
```
phackstock, line 20:
```suggestion
def cli_valid_project(path: Path, mappings: str, definitions: str):
```

nomenclature/testing.py
```diff
@@ -36,11 +36,14 @@ def assert_valid_structure(path: Path):
         :class:`DataStructureDefinition`
       - If a `mappings` folder exists, it must be a valid :class:`RegionProcessor`
     """
-    definition = nomenclature.DataStructureDefinition(path / "definitions")
-    if (path / "mappings").is_dir():
+
+    definition = nomenclature.DataStructureDefinition(path / str(definitions))
+    if (path / str(mappings)).is_dir():
```
phackstock, line 41: I think the `str()` conversion is not strictly needed
```suggestion
    if (path / mappings).is_dir():
```

nomenclature/testing.py
```diff
@@ -36,11 +36,14 @@ def assert_valid_structure(path: Path):
         :class:`DataStructureDefinition`
       - If a `mappings` folder exists, it must be a valid :class:`RegionProcessor`
```
phackstock, line 37: Since there's two new arguments for the function a docstring explaining what they do would be great.
Since the current doc string is not in the standard format, I could go ahead and re-write it to numpydoc style if that's ok for you.

nomenclature/testing.py
```diff
@@ -36,11 +36,14 @@ def assert_valid_structure(path: Path):
...
-    definition = nomenclature.DataStructureDefinition(path / "definitions")
-    if (path / "mappings").is_dir():
+
+    definition = nomenclature.DataStructureDefinition(path / str(definitions))
+    if (path / str(mappings)).is_dir():
         nomenclature.RegionProcessor.from_directory(
-            path / "mappings"
+            path / str(mappings)
         ).validate_mappings(definition)
+    elif mappings != "mappings":
```
phackstock, line 45: Not sure I fully understand the logic of this `elif`. Let's say the `mappings` variable is indeed `"mappings"` but such a folder does not exist.
Then `if (path / str(mappings)).is_dir()` would be false but so would be `elif mappings != "mappings"` which means we would not get the error that we would expect. 
Perhaps, this would be better?
```suggestion
    else: # if we don't find the given directory just throw an error
```

nomenclature/cli.py
```diff
@@ -29,4 +31,4 @@ def cli_valid_project(path: Path):
        in a model mapping are defined in the region codelist.
```
phackstock, line 31: Adding an example here for non-standard use that will be rendered on https://nomenclature-iamc.readthedocs.io/en/stable/cli.html would be great.
Something like
```python
"""
...
    Example
    -----------
    $ nomenclature validate-project . --mappings mapping --definitions definitions

"""
```

tests/test_cli.py
```diff
@@ -75,4 +75,44 @@ def print_helper(_path):
...
+
+def test_cli_wrong_definitions_name():
+    """Check that CLI expects an error when wrong non-default directory
+    is given"""
+    result_valid = runner.invoke(
+        cli,
+        [
+            "validate-project",
+            str(TEST_DATA_DIR / "structure_validation"),
+            "--definitions=def",
```
phackstock, line 115: This test fails but I don't think it does for the reasons intended. The way it currently is, I think the cli is called incorrectly. This way should be the correct one and the cli will fail correctly then.
```suggestion
            "--definitions", "def",
```
What would be an option, although probably overkill in this case is not only assert the `exit_code` from the cli call but also, for the failing test cases the content of the error message. But this would be subject for another PR I would say.

phackstock (changes requested):
Very solid start @luciecastella :+1:.
A few comments inline below

tests/test_cli.py
luciecastella: Good point, did not see that. How come it runs fine with test_cli_non_default_folders() then?

nomenclature/testing.py
luciecastella: OK for me. I can also try to do it if you prefer.

phackstock: Thanks, would be great if you could do that.

phackstock:
@danielhuppmann after discussing with @luciecastella we would need your input on the logic of the fail condition of `assert_valid_structure`. Previously the way it worked was that if a folder called `"mappings"` existed, it would be checked if its contents are all valid model mappings. If it did not exist nothing happened.
Now, since we allow the user to optionally specify a directory other than the default `"mappings"`, it's a little more complicated. The way it's currently proposed is that if the directory `path / mappings` does not exist **and** `mappings` is **not** `"mappings"` (the default value), we raise an error. The potential problem I'd see with that is if the user provides the string `"mappings"` and it does not exist we currently treat that as fine which would be undefined behavior. 
What I would propose is to set the default value for `mappings` to `None` and check this way if `"mappings"` was user provided or not. 
As I'm typing this I realize that we should probably just go ahead with the merge in this PR and I'll fix it quickly in a follow-up. Does that sound good to you?

danielhuppmann:
> What I would propose is to set the default value for mappings to None and check this way if "mappings" was user provided or not.

Yes.

> As I'm typing this I realize that we should probably just go ahead with the merge in this PR and I'll fix it quickly in a follow-up. Does that sound good to you?

Yes.

Thank you.

phackstock (approved):
Looks all good to me, good work @luciecastella. Ready to be merged from my side.

nomenclature/cli.py
```diff
@@ -27,6 +29,11 @@ def cli_valid_project(path: Path):
     3. Test that all model mappings in `mappings/` can be correctly parsed as a
        :class:`RegionProcessor` object. This includes a check that all regions mentioned
        in a model mapping are defined in the region codelist.
+
+    Example
+    -------
+    $ nomenclature validate-project . --mappings map --definitions def
```
danielhuppmann, line 35: Maybe change to reflect that these are folder names and turn order to follow docs structure...
```suggestion
    $ nomenclature validate-project .  --definitions <def-folder> --mappings <mapping-folder>
```

nomenclature/testing.py
```diff
@@ -27,20 +27,40 @@ def assert_valid_yaml(path: Path):
         )
 
 
-def assert_valid_structure(path: Path):
+def assert_valid_structure(
+    path: Path, mappings: str = "mappings", definitions: str = "definitions"
+) -> None:
     """Assert that `path` can be initialized as a :class:`DataStructureDefinition`
+    with optionnaly non-default folder names
```
danielhuppmann, line 34: There is a typo in this line, but it's anyway better practice to keep the first sentence of a docstring to one line. And optional stuff doesn't need to be mentioned in the first line.
```suggestion
```

tests/test_cli.py
```diff
@@ -75,4 +75,47 @@ def print_helper(_path):
...
+    """Check that CLI expects an error when non-default "definitions" and
+    "mappings" directory names are not given"""
+    result_valid = runner.invoke(
+        cli, ["validate-project", str(TEST_DATA_DIR / "non-default_folders")]
+    )
+    assert result_valid.exit_code == 1
+
+
+def test_cli_wrong_definitions_name():
+    """Check that CLI expects an error when wrong non-default directory
```
danielhuppmann, line 110: it's not obvious what is meant by "wrong" in this line.

Also, minor edit suggestion
```suggestion
    """Check that CLI raises an error when wrong non-default directory
```

danielhuppmann (approved):
Looks good overall, just a few minor comments on the docstrings.

Good to be merged after these edits are implemented.

luciecastella:
Thank you, I have now implemented the changes. But I still cannot merge it myself @phackstock @danielhuppmann

phackstock:
I'll go ahead with the merge.

PR #156, opened 2022-08-18
Check if user provided "mapping" attribute in assert_valid_structure

### LINKED ISSUE #155: Check user input for assert_valid_structure
In `testing.assert_valid_structure` the default for `mappings` is currently `"mappings"`. 
If a folder of the name `"mappings"` does not exist there is currently no error raised. Reason being that there's no way to know if this was the default (in which case no error should be raised) or a user provided value, in which case there should be an error.

phackstock:
closes #155.

Quick little PR to address the logic improvements discussed in #153.

## Features implemented

`testing.assert_valid_structure` now has the following default `mapping: str = None`. This means that we can now distinguish if the user provided input or we are using our default. This allows for the following logic:
* If `mapping is None`, check if a folder called `"mappings/"` (the previous default value) exists. If so check the mappings, if not exit **without** an error. This is the desired behavior since mappings are not strictly required. 
* If the user provided a value for `mapping`. Check if a folder of that name exists. If so check the mappings, if not, raise a `FileNotFound` error. The logic here being that if the user provided a value for `mappings` it needs to exist.

### DIFF
--- a/nomenclature/testing.py
+++ b/nomenclature/testing.py
@@ -31,7 +31,7 @@ def assert_valid_yaml(path: Path):
 def assert_valid_structure(
     path: Path,
     dimensions: List[str] = ["region", "variable"],
-    mappings: str = "mappings",
+    mappings: str = None,
     definitions: str = "definitions",
 ) -> None:
     """Assert that `path` can be initialized as a :class:`DataStructureDefinition`
@@ -41,11 +41,11 @@ def assert_valid_structure(
     path : Path
         directory path to the file of interest
     dimensions : List[str]
-        Optionnal list of dimensions to be checked
+        Optional list of dimensions to be checked
     mappings : str
-        Optionnal non-default name for the mappings folder
+        Optional non-default name for the mappings folder, default "mappings"
     definitions : str
-        Optionnal non-default name for the definitions folder
+        Optional non-default name for the definitions folder
 
     Notes
     -----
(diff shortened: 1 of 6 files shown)


nomenclature/testing.py
```diff
@@ -37,9 +37,9 @@ def assert_valid_structure(
     path : Path
         directory path to the file of interest
     mappings : str
-        Optionnal non-default name for the mappings folder
+        Optional non-default name for the mappings folder
```
danielhuppmann, line 40: Suggest to add a note here that the default folder name is "mappings" because this is not clear from the signature.

phackstock: Done.

phackstock:
Not sure why stickler is complaining. The error it raises is W503 which is "line break before binary operator". There was a change in 2016 when the opposite was deemed the way to go. Details can be found here: https://stackoverflow.com/questions/57074300/what-is-the-recommended-way-to-break-long-if-statement-w504-line-break-after-b. 
We've had this issue before and for this reason I've added W503 to the ignore list of `.stickler.yml` so in principle the error that stickler is reporting should be ignored. Not sure why it's not...
=

phackstock:
Found the reason for the stickler error, this is the command that's run (from [here](https://github.com/IAMconsortium/nomenclature/pull/156/checks?check_run_id=7897476166)):

```console
INFO: Running container: flake8 --ignore ['W605', 'W503'] --max-line-length 88 --max-complexity 20 --isolated /src/nomenclature/cli.py /src/nomenclature/testing.py /src/tests/test_cli.py
```

For some reason it seems that flake8 is run with the `--isolated` flag for three files. This flag disregards all the previous config. 
Not sure why this is the reason though and why this those three files specifically.

phackstock:
Now stickler works as expected the error was in `.stickler.yml`, the ignore arguments should be given as a string and not a list:
```yaml
ignore: W503, W605
```
instead of 
```yaml
ignore:
  - W503
  - W605
```

The `--isolated` option is actually correct and desired as it runs stickler only on the files changed in the current PR.

PR #159, opened 2022-08-18
Docs/numpystyle use

### LINKED ISSUE #158: Use numpystyle correctly in docstrings
I noticed that in #153, @luciecastella added docstrings structured as

```
Parameters
 ----------
 mappings : str
     Optionnal non-default name for the mappings folder
```
whereas the correct numpy-style would be
```
    Parameters
     ----------
     mappings : str, optional
         Name for the mappings folder
```
See https://numpydoc.readthedocs.io/en/latest/format.html#parameters

Also, it's not necessary to explain that something is non-default - that is (usually) clear from the function signature.

Please fix after #156 is merged

luciecastella:
Closes #158

Resolves the numpystyle issue with proper docstrings for the optional parameters.

### DIFF
--- a/nomenclature/testing.py
+++ b/nomenclature/testing.py
@@ -40,12 +40,12 @@ def assert_valid_structure(
     ----------
     path : Path
         directory path to the file of interest
-    dimensions : List[str]
-        Optional list of dimensions to be checked
-    mappings : str
-        Optional non-default name for the mappings folder, default "mappings"
-    definitions : str
-        Optional non-default name for the definitions folder
+    dimensions : List[str], optional
+        List of dimensions to be checked
+    mappings : str, optional
+        Name for the mappings folder
+    definitions : str, optional
+        Name for the definitions folder
 
     Notes
     -----
--- a/nomenclature/testing.py
+++ b/nomenclature/testing.py
@@ -43,9 +43,9 @@ def assert_valid_structure(
     dimensions : List[str], optional
         List of dimensions to be checked
     mappings : str, optional
-        Name for the mappings folder
+        Name for the mappings folder, defaults to "mappings"
     definitions : str, optional
-        Name for the definitions folder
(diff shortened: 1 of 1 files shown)


nomenclature/testing.py
```diff
@@ -40,12 +40,12 @@ def assert_valid_structure(
...
-    dimensions : List[str]
-        Optional list of dimensions to be checked
-    mappings : str
-        Optional non-default name for the mappings folder, default "mappings"
-    definitions : str
-        Optional non-default name for the definitions folder
+    dimensions : List[str], optional
+        List of dimensions to be checked
+    mappings : str, optional
+        Name for the mappings folder
```
danielhuppmann, line 46: The name of the standard folder is not clear from the function signature (because it says `None`), so better to repeat it here for clarity.
```suggestion
        Name for the mappings folder, defaults to "mappings"
```

nomenclature/testing.py
```diff
@@ -40,12 +40,12 @@ def assert_valid_structure(
...
+    definitions : str, optional
+        Name for the definitions folder
```
danielhuppmann, line 48: This is repeating from the signature, but I'd add it for consistency with the "mappings" description...
```suggestion
        Name for the definitions folder, defaults to "definitions"
```

danielhuppmann (changes requested):
Thanks, two suggestions inline

danielhuppmann (approved):
Thanks, looks good to me!

phackstock (approved):
Looks good to me as well.
A tip for the future, there is a vs code plugin called [autoDocstring](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring), that automatically creates the structure of the docstring for you. I also knows type hints and fills them out. Only thing you have to change is the style which is google by default I think.

PR #160, opened 2022-08-19
Features/default command cli

### LINKED ISSUE #157: Add validation-CLI option to check 'all' dimensions
Building on #147.

We have several projects where we also use the "scenario" dimension for validation, eg https://github.com/iiasa/ariadne-intern-workflow, and we sometimes run into the problem that PRs that change the scenarios-codelist do not trigger the validation but then fail in production (ie when users upload a scenario data file).

We could now change the GitHub Actions workflow in that repository to

```
      run: nomenclature validate-project . --dimensions "['variable', 'region', 'scenario']"
```

However, this would still require a project manager to be aware that this change is necessary when adding validation for new dimensions.

I think it would be better to have a generic option, e.g., 

```
      run: nomenclature validate-project . --dimensions all
```
where all folders in *definitions* are checked automatically.

The GitHub Actions workflow in the template repository https://github.com/iiasa/scenario-explorer-workflow-template should then also be updated accordingly, so that by default, all dimensions are checked in all new projects.

luciecastella:
Closes #157 

The program now checks all (even non-default: 'variable' and 'region') dimensions if nothing is specified to the function or in the command line.


nomenclature/testing.py
```diff
@@ -30,7 +31,7 @@ def assert_valid_yaml(path: Path):
 
 def assert_valid_structure(
     path: Path,
-    dimensions: List[str] = ["region", "variable"],
+    dimensions: List[str] = "[]",
```
phackstock, line 34: Same as in `cli.py`, I would avoid `"[]"` as a default. In this special instance it also contradicts the type hint as this is a string and not a list of strings. 
I would go with `None` as the default. For the type hints you can then use `Optional[List[str]]` to indicate to python type checkers that the input can be `None.` Using the same `Optional` for `mappings` might be a good idea as well.

nomenclature/testing.py
```diff
@@ -55,6 +56,8 @@ def assert_valid_structure(
         - If a `mappings` folder exists, it must be a valid :class:`RegionProcessor`
 
     """
+    if dimensions == []:
```
phackstock, line 59: Together with my comment above I would go with the following:
```suggestion
    if dimensions is None:
```

nomenclature/testing.py
```diff
@@ -55,6 +56,8 @@ def assert_valid_structure(
...
+        dimensions = os.listdir(f"{str(path)}/{definitions}")
```
phackstock, line 60: This one's a bit dangerous as it lists _all_ folders and _files_ in the definitions directory. Also you are mixing functions from the `os` module and from `pathlib`. The latter is the newer way to do things and almost everything that the `os` module implements in terms of file operations can be achieved with `pathlib`. Additionally, you don't have to convert to string anymore. Here's a list of how the `pathlib` module implements the `os` module's functionality, https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module.
```suggestion
        dimensions = [x for x in (path / definitions).iterdir() if x.is_dir()]
```

tests/test_cli.py
```diff
@@ -136,34 +136,45 @@ def test_cli_custom_dimensions():
...
+            str(TEST_DATA_DIR / "non-default_dimensions_failing"),
             "--dimensions",
-            "['variable', 'region', 'foo']",
+            "['variable', 'region']",
         ],
     )
+    assert result_valid.exit_code == 0
+
+
+def test_cli_empty_dimensions_fails():
```
phackstock, line 171: Since this is a newly created test for this PR, it would be great to not only test for the exit code but for the type of error and error message as well. You could take some inspiration from here: https://github.com/IAMconsortium/nomenclature/blob/main/tests/test_cli.py#L200

tests/test_cli.py
```diff
@@ -136,34 +136,45 @@ def test_cli_custom_dimensions():
...
+def test_cli_custom_dimensions_fails():
+    """Check that CLI raises an error when specifying a non-existing
+    directory ('foo')"""
 
     result_valid = runner.invoke(
         cli,
         [
             "validate-project",
-            str(TEST_DATA_DIR / "non-default_dimensions_failing"),
+            str(TEST_DATA_DIR / "non-default_dimensions_passing"),
```
phackstock, line 147: I think with the way the tests have evolved you might want to rename the two directories `non-default_dimensions_passing` and `non-default_dimensions_failing` as this test uses the first directory and is supposed to fail while the next one uses the second and is supposed to run. Maybe something more descriptive could be an idea like `non-default_dimensions_one_empty` or something like this.

phackstock (changes requested):
Looks good overall @luciecastella.
A few comments in line.

nomenclature/testing.py
luciecastella: @phackstock The os version allowed me to directly get the name of the directories. With the option you suggest, we get wholes paths and I have to change what is then sent to the next variable. But OK, I can try to work with that.

phackstock: Ah yes sorry that was a little imprecise on my side. This should do the job: 
```suggestion
       dimensions = [x.stem for x in (path / definitions).iterdir() if x.is_dir()]
```
Reference: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem

nomenclature/cli.py
```diff
@@ -11,6 +11,8 @@
 class PythonLiteralOption(click.Option):
     def type_cast_value(self, ctx, value):
         try:
+            if value is None:
```
phackstock, line 14: In the interest of keeping the `try...except` clause as small as possible you could move the `if value is None` check above so that this is done beforehand.

phackstock (changes requested):
Looks very good to me, almost finished. One small comment in line.
The only thing that's left to do from my point of view is documentation. The fact that the `dimensions` argument in `cli_valid_project` and `assert_valid_structure` defaults to `None` which means "validate everything that you find" should be communicated. I would put it in the docstrings of both functions. Then it should be good.

nomenclature/cli.py
```diff
@@ -50,6 +50,27 @@ def cli_valid_project(
 ):
     """Assert that `path` is a valid project nomenclature
 
+    Parameters
+    ----------
+    path : Path
+        directory path to the file of interest
+    dimensions : List[str], optional
+        List of dimensions to be checked, default to None which implies that all
+        directories in `dimensions` will be checked
```
phackstock, line 59:
```suggestion
        directories in `definitions` will be checked
```

nomenclature/testing.py
```diff
@@ -41,9 +41,11 @@ def assert_valid_structure(
     path : Path
         directory path to the file of interest
     dimensions : List[str], optional
-        List of dimensions to be checked
+        List of dimensions to be checked, default to None which implies that all
+        directories in `dimensions` will be checked
```
phackstock, line 45:
```suggestion
        directories in `definitions` will be checked
```

phackstock (changes requested):
Two tiny suggestions otherwise good to be merged from my side, good work @luciecastella.

nomenclature/cli.py
```diff
@@ -41,10 +43,34 @@ def cli_valid_yaml(path: Path):
...
+    dimensions: Optional[List[str]],
+    mappings: Optional[str],
+    definitions: str,
 ):
     """Assert that `path` is a valid project nomenclature
 
+    Parameters
+    ----------
+    path : Path
+        directory path to the file of interest
```
danielhuppmann, line 56:
```suggestion
        Project directory to be validated
```

nomenclature/cli.py
```diff
@@ -41,10 +43,34 @@ def cli_valid_yaml(path: Path):
     default="definitions",
 )
 def cli_valid_project(
-    path: Path, dimensions: List[str], mappings: str, definitions: str
+    path: Path,
+    dimensions: Optional[List[str]],
+    mappings: Optional[str],
+    definitions: str,
```
danielhuppmann, line 49:
```suggestion
    definitions: Optional[str],
```

nomenclature/cli.py
luciecastella: oupsi I keep inverting dimensions and definitions, sorry for that. thank you @phackstock !!

nomenclature/cli.py
```diff
@@ -41,10 +43,34 @@ def cli_valid_yaml(path: Path):
...
 ):
     """Assert that `path` is a valid project nomenclature
 
+    Parameters
+    ----------
+    path : Path
+        directory path to the file of interest
+    dimensions : List[str], optional
+        List of dimensions to be checked, default to None which implies that all
+        directories in `dimensions` will be checked
```
danielhuppmann: Alternative, more concise formulation
```suggestion
        Dimensions to be checked, defaults to all sub-folders of `dimensions`
```

nomenclature/cli.py
```diff
@@ -41,10 +43,34 @@ def cli_valid_yaml(path: Path):
...
+    mappings : str, optional
+        Name for the mappings folder, defaults to None which implies that if the
+        `mappings` directory is not found, there are no mappings to check
+    definitions : str, optional
+        Name for the definitions folder, defaults to "definitions"
```
danielhuppmann, line 64:
```suggestion
        Name of the definitions folder, defaults to "definitions"
```

nomenclature/cli.py
```diff
@@ -41,10 +43,34 @@ def cli_valid_yaml(path: Path):
...
+    Parameters
+    ----------
+    path : Path
+        directory path to the file of interest
+    dimensions : List[str], optional
+        List of dimensions to be checked, default to None which implies that all
+        directories in `dimensions` will be checked
+    mappings : str, optional
+        Name for the mappings folder, defaults to None which implies that if the
+        `mappings` directory is not found, there are no mappings to check
```
danielhuppmann, line 62:
```suggestion
        Name of the mappings folder, defaults `mappings` (if this folder exists)
```

nomenclature/testing.py
```diff
@@ -41,9 +41,11 @@ def assert_valid_structure(
     path : Path
         directory path to the file of interest
```
danielhuppmann, line 42: Please add the same changes as I suggested above.

danielhuppmann (approved):
Thanks, looks good on the workflow, but I have some additional suggestions to make the docstrings more concise

nomenclature/cli.py
phackstock: I don't think `Optional` would be correct here. In the python typing sense `Optional` specifies a parameter that can be `None` not necessarily an optional parameter (meaning there is a default value). Here, the function would fail if `None` was used as an input.

danielhuppmann: Ok, I see your point but I find it confusing that a "required" (or non-optional) argument comes after optional ones. I suggest to either change the order (for both the CLI and the `testing.assert_valid_structure()`) or set the default for *definitions* also to `None` (and replace it in the code).

phackstock: Fair point, I would vote for changing the order then. When invoking the cli with the parameter names the order does not matter anyway. I would leave the string default in the function definition since it's more concise.

luciecastella: Fine to me. We loose the order that @danielhuppmann recommended last week with 'dimensions' first because it will be used more, but OK to me.

luciecastella:
I added a check for non-existing definition folder because I suddenly failed one of the tests I did not fail before changing the order of the arguments. 
I think all your reviews are applied, I think it can be merged if you think so too @phackstock @danielhuppmann

nomenclature/testing.py
```diff
@@ -30,22 +30,22 @@ def assert_valid_yaml(path: Path):
...
-    dimensions : List[str], optional
-        List of dimensions to be checked
-    mappings : str, optional
-        Name for the mappings folder, defaults to "mappings"
+        Project directory to be validated
     definitions : str, optional
-        Name for the definitions folder, defaults to "definitions"
+        Name of the definitions folder, defaults to "definitions"
+    mappings : str, optional
+        Name of the mappings folder, defaults `mappings` (if this folder exist)
```
danielhuppmann, line 46: Sorry, my bad, I was confused in my formatting in the suggestion earlier - we should use "" for a proper name and `` for the argument...
```suggestion
        Name of the mappings folder, defaults to "mappings" (if this folder exist)
```

nomenclature/cli.py
```diff
@@ -26,25 +28,47 @@ def cli_valid_yaml(path: Path):
...
     """Assert that `path` is a valid project nomenclature
 
+    Parameters
+    ----------
+    path : Path
+        Project directory to be validated
+    definitions : str, optional
+        Name of the definitions folder, defaults to "definitions"
+    mappings : str, optional
+        Name of the mappings folder, defaults `mappings` (if this folder exist)
```
danielhuppmann, line 60: As below
```suggestion
        Name of the mappings folder, defaults to "mappings" (if this folder exists)
```

phackstock (approved):
Looks all good to me.

PR #161, opened 2022-08-22
Add common vs native region explanation

### LINKED ISSUE #145: Documentation of common region vs. model-native region
Per this comment by @luciecastella in https://github.com/IAMconsortium/nomenclature/issues/144#issuecomment-1210296974_, it would be useful to have an own section explaining the difference between a "common" region vs. a "model_native" region. Could be an own new sub-page of the user guide or in the Region section of the "codelist" page.

phackstock:
closes #145.

As discussed, this PR adds a short section detailing the differences between model native and common regions. I have kept the explanation very minimal. Hope this work for clearing up confusion and giving people a quick explanation.

### DIFF
--- a/doc/source/user_guide/model-mapping.rst
+++ b/doc/source/user_guide/model-mapping.rst
@@ -21,9 +21,14 @@ Difference between model-native and common regions
 Model native regions are regions that are reported *directly* from an integrated
 assessment model. 
 
-Common regions, on the other hand, are calculated by *aggregating model native*
-regions. Common regions are used to compare results from models with different
-native regional resolution. 
+Common regions, on the other hand, are calculated by *aggregating* a model's *native*
+regions. Common regions are used to compare results from models with different native
+regional resolution. 
+
+As an example, let's say we have two models *model_a* and *model_b*. model_a reports
+native regions *Western* and *Eastern Europe* while *model_b* reports *Northern* and
+*Southern Europe*. In order to compare the results of the two models we might introduce
+a common region called *Europe*. 
 
 
 Model mapping format specification
--- a/doc/source/user_guide/model-mapping.rst
+++ b/doc/source/user_guide/model-mapping.rst
@@ -13,6 +13,18 @@ purposes:
 2. Allowing for renaming of model native regions.
 3. Defining how model native regions should be aggregated to common regions.
 
+.. _native-vs-common-region:
+
+Difference between model-native and common regions
+---------------------------------------------------
+
+Model native regions are regions that are reported *directly* from an integrated
+assessment model. 
+
+Common regions, on the other hand, are calculated by *aggregating model native*
+regions. Common regions are used to compare results from models with different
+native regional resolution. 
+
 
 Model mapping format specification
 ----------------------------------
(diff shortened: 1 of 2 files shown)


danielhuppmann (commented):
I think that having the explanation at the end of a very long and technical page is not helpful, because non-expert users will not look there. I suggest to either put it at the top of this page or add it to the region-codelist page.

Also, please set your auto-format tool to 88 characters per line to not have spurious changes in this PR.

phackstock:
I've reset the formatting and moved the explanation towards the top of the page model mapping. 
I think the explanation is better here since the region codelist page does not really touch on the topic at the moment.

doc/source/user_guide/model-mapping.rst
```diff
@@ -11,8 +11,19 @@ purposes:
...
+
+.. _native-vs-common-region:
+
+Difference between model-native and common regions
+---------------------------------------------------
+
+Model native regions are regions that are reported *directly* from an integrated
+assessment model. 
+
+Common regions, on the other hand, are calculated by *aggregating model native*
```
luciecastella, line 24: Aggregating regions from one model or different models? Maybe this could be added. 

Also, maybe an example would be useful?

luciecastella (changes requested):

doc/source/user_guide/model-mapping.rst
phackstock: For comparison you aggregate only regions from one model otherwise you'd be mixing results. I'll try to make it more clear. 
I've thought about adding an example but left it out as I wanted the explanation to be as short as possible. Seems like I lost clarity on the way. I'll add an example to make things more clear.

phackstock:
Added an example for the difference between native and common regions. I kept it minimal again. Please let me know if it makes sense to you @luciecastella or if I should elaborate further.

luciecastella (approved):
Great! Yes I think like this it is very clear, thank you :)

phackstock:
@danielhuppmann, does the explanation look good to you as well?

doc/source/user_guide/model-registration.rst
```diff
@@ -44,6 +44,8 @@ For this guide we will consider a model mapping as an example:
 ``model_a`` reports three regions natively, ``region_1``, ``region_2`` and ``region_3``.
 These three are to be renamed to the ``{model}|{region}`` pattern as is common place in
 many projects. For the ``common_regions``, there are ``World`` and ``Common region 1``.
+*For more details on the model native and common regions refer to*
```
danielhuppmann, line 47: Suggest to add an empty line and use a hyphen (model-native) in such cases to improve readability.
```suggestion

*For more details on the model-native and common regions refer to*
```

doc/source/user_guide/model-mapping.rst
```diff
@@ -11,8 +11,24 @@ purposes:
...
-   regions.
+3. Defining how model native regions should be aggregated to common regions.
+
+.. _native-vs-common-region:
+
+Difference between model-native and common regions
+---------------------------------------------------
+
+Model native regions are regions that are reported *directly* from an integrated
+assessment model.
```
danielhuppmann, line 22: For generality (to avoid offending energy-systems modelers)...
```suggestion
assessment or energy system model.
```

danielhuppmann (approved):
A few suggestions inline, then good to be merged!

PR #168, opened 2022-08-26
Feature/no sub folder

### LINKED ISSUE #165: Variable and Region definition not mandatory anymore
I realised that now that the program checks for all the dimensions to be found when nothing is specified, it will just validate the ones it founds. Meaning we could totally have just a scenario folder with no variable and region definition and it would pass the validation. This was not the case before as they were in the default value. Should we change that and require variables and regions when nothing is specified?

luciecastella:
Closes #165 

Throws an error when `definitions` is empty if no dimension is specified.


nomenclature/testing.py
```diff
@@ -61,6 +61,11 @@ def assert_valid_structure(
         )
     if dimensions is None:
         dimensions = [x.stem for x in (path / definitions).iterdir() if x.is_dir()]
+        if not dimensions:
+            raise FileNotFoundError(
+                f"`Definitions` directory is empty : {path / definitions}"
+                '. Specify --dimensions "[]" if this is intended.'
```
phackstock, line 67: Not sure about this suggestion. 
If `"[]"` is provided, `DataStructureDefinition` is empty and as a result mappings cannot be validated either so `assert_valid_structure` would be pointless. If there are no definitions, you should not run `assert_valid_structure` at all.
I think I would just leave the suggestion out. 
```suggestion
```

phackstock (commented):
Looks good to me. One small comment below.
You could also add a short description to the docstring section of this function describing the newly introduced error scenario.

nomenclature/testing.py
luciecastella: True...

luciecastella: @phackstock Regarding this comment, this would actually mean that if people run validate-project with `--dimensions "[]"`, it would be useless. Should we somehow forbid this command?

tests/test_cli.py
```diff
@@ -214,3 +214,17 @@ def test_cli_missing_mappings_fails():
     assert cli_result.exit_code == 1
     assert isinstance(cli_result.exception, FileNotFoundError)
     assert "Mappings directory not found" in str(cli_result.exception)
+
+
+def test_cli_no_subfolder():
+    """Assert that an error is raised when the `definitions` directory is empty if
+    no dimension is specified"""
```
luciecastella, line 221: This "if no dimension is specified" maybe does not make sense then?

nomenclature/testing.py
luciecastella: I could change the code so that if we type `--dimensions "[]"` it still looks for all the available dimensions...

nomenclature/testing.py
```diff
@@ -61,6 +62,10 @@ def assert_valid_structure(
         )
     if dimensions is None:
         dimensions = [x.stem for x in (path / definitions).iterdir() if x.is_dir()]
+        if not dimensions:
+            raise FileNotFoundError(
+                f"`Definitions` directory is empty : {path / definitions}"
```
danielhuppmann, line 67:
```suggestion
                f"`definitions` directory is empty: {path / definitions}"
```

nomenclature/testing.py
```diff
@@ -61,6 +62,10 @@ def assert_valid_structure(
         )
     if dimensions is None:
         dimensions = [x.stem for x in (path / definitions).iterdir() if x.is_dir()]
+        if not dimensions:
```
danielhuppmann, line 65: You could simply move the new lines outside of the if-statement above, then the empty-check would also apply if a user calls the validation with ` --dimensions "[]"`.

danielhuppmann (commented):
Two suggestion inline

nomenclature/testing.py
danielhuppmann: Agree with @phackstock, it should not be possible to validate with an empty definitions folder, so --dimensions "[]" should also raise an error. See my inline suggestion.

nomenclature/testing.py
luciecastella: OK thank you! That was indeed what I wanted to do but I was unsure we wanted this behaviour or not

luciecastella: @danielhuppmann Can I actually separate the case where the sub-folders do not exist and the case when "[]" is given (but sub-folders exist)? I feel like the error message should not be the same...

danielhuppmann: You are absolutely right, these are two separate errors...

luciecastella: I took the liberty of doing so, tell me if I should change it

nomenclature/testing.py
```diff
@@ -52,15 +52,26 @@ def assert_valid_structure(
...
+    if dimensions is None:  # if "dimensions" were not specified
         dimensions = [x.stem for x in (path / definitions).iterdir() if x.is_dir()]
+        if not dimensions:
+            raise FileNotFoundError(
+                f"`definitions` directory is empty: {path / definitions}"
+            )
+
+    if dimensions == []:  # if "dimensions" were specified as "[]"
+        raise FileNotFoundError(
+            f"Invalid dimensions: {dimensions}, there is nothing to validate."
```
danielhuppmann, line 73: I think that this will be shown as 
> Invalid dimensions: , there is nothing to validate.
which is not quite helpful. Maybe better
> No dimensions to validate.

nomenclature/testing.py
```diff
@@ -52,15 +52,26 @@ def assert_valid_structure(
...
+
+    if dimensions is None:  # if "dimensions" were not specified
         dimensions = [x.stem for x in (path / definitions).iterdir() if x.is_dir()]
+        if not dimensions:
+            raise FileNotFoundError(
+                f"`definitions` directory is empty: {path / definitions}"
+            )
+
+    if dimensions == []:  # if "dimensions" were specified as "[]"
+        raise FileNotFoundError(
```
danielhuppmann, line 72: I think that this should rather be a ValueError.

nomenclature/testing.py
```diff
@@ -52,15 +52,26 @@ def assert_valid_structure(
...
-    if dimensions is None:
+
+    if dimensions is None:  # if "dimensions" were not specified
         dimensions = [x.stem for x in (path / definitions).iterdir() if x.is_dir()]
+        if not dimensions:
+            raise FileNotFoundError(
+                f"`definitions` directory is empty: {path / definitions}"
+            )
+
+    if dimensions == []:  # if "dimensions" were specified as "[]"
```
danielhuppmann, line 71: I think that it would be more intuitive to do this check first (if user provided an empty list, raise an error; then continue to see whether there are folders to validate...)

danielhuppmann (commented):
A few more suggestions...

danielhuppmann (approved):
One more comment inline, then good to be merged from my side!

phackstock (approved):
Looks good to me, I'll go ahead with the merge then.

Issue #164, opened 2022-08-24
Raise error if code has trailing whitespace


danielhuppmann:
Per the PR https://github.com/openENTRANCE/openentrance/pull/222 by @Renato-Rodrigues: it would be useful to add a validation that any code (variable, region, scenario name) cannot end on a whitespace.

luciecastella:
Do you know which error did it trigger? because trailing whitespaces are automatically erased and codes in the codelist have none, even if the files has some.

danielhuppmann:
Ah, interesting to know that the pyyaml automatically removes trailing whitespaces (just tested this myself). I think that in this case, there is no need to protect against them from the point of view of the nomenclature package.

Renato-Rodrigues:
It was not a whitespace. It was some other hidden formatting character.

![image](https://user-images.githubusercontent.com/33692508/186621628-bed3ee4b-e198-4dce-8063-524624128279.png)

(https://github.com/openENTRANCE/openentrance/pull/222)

luciecastella:
OK interesting, I'll see if there is a way to check that. Thank you!

danielhuppmann:
Reopening as this might be a useful validation check after all...

luciecastella:
@danielhuppmann after some discussion with @phackstock , we concluded that we should check for all the lines of the yaml files, and therefore make it a function in `testing.py`. Now the question would be: do we want to check all yaml files to make sure there are no special character anywhere, or only in the dimensions specified by the user? (related to Issue #165)

danielhuppmann:
Good question @luciecastella...

My initial idea would have been to simply define a list of illegal characters/expressions and then check that a code does not have any of those. I think that this would be enough for what I had in mind, and you would have the benefit of not having to worry about which sub-folders/dimensions to check.

You could do the more elaborate check in `testing.assert_valid_yaml()`, copying the signature and behavior of `testing.assert_valid_structure()`. If you're feeling ambitious...

danielhuppmann:
One more follow-up: some users create the yaml files from Excel, see the tests for this feature as illustration [here](https://github.com/IAMconsortium/nomenclature/blob/a5a8789c6f0b1e28dec9a838e2cd6346a18f39b0/tests/test_definition.py#L71). Now, it's quite easy to overlook trailing whitespaces (not special characters, just simple spaces) in Excel, and the result in yaml is something like
```yaml
- 'Primary Energy|Coal ':
    unit: EJ/yr
```
[...]

PR #169, opened 2022-08-26
Fix/check hidden character

luciecastella:
First half of #164 resolution. 

The error message in case of special character is: 
ValueError: Unexpected special character(s) in: scenarios.yaml, line 3, col 12. variables.yaml, line 5, col 53.

Tell me if it could be written in a more understandable way.


tests/test_testing.py
```diff
@@ -26,3 +26,10 @@ def test_assert_yaml_fails(caplog):
     obs = log[2].replace("\n", "")  # strip newlines from observed log message
     exp = r"Error parsing file while scanning a simple key.*, line 4, column 1"
     assert re.match(exp, obs)
+
+
+def test_special_character():
```
phackstock, line 31: I would call this "hidden character".
```suggestion
def test_hidden_character():
```

tests/test_testing.py
```diff
@@ -26,3 +26,10 @@ def test_assert_yaml_fails(caplog):
...
+    """Check that a non-printable character in any yaml file will raise an error"""
+    match = "variables.yaml, line 5, col 53."
+    with pytest.raises(ValueError, match=match):
+        assert_valid_yaml(TEST_DATA_DIR / "special_character")
```
phackstock, line 35: I would also rename the folder where the test data are located
```suggestion
        assert_valid_yaml(TEST_DATA_DIR / "hidden_character")
```

tests/test_testing.py
```diff
@@ -26,3 +26,10 @@ def test_assert_yaml_fails(caplog):
     obs = log[2].replace("\n", "")  # strip newlines from observed log message
     exp = r"Error parsing file while scanning a simple key.*, line 4, column 1"
     assert re.match(exp, obs)
+
+
+def test_special_character():
+    """Check that a non-printable character in any yaml file will raise an error"""
+    match = "variables.yaml, line 5, col 53."
```
phackstock, line 33: If you only test for the invisible character in variables.yaml, which totally fine, you should leave `data/special_character/definitions/scenario/scenarios.yaml` out as it is not used in the test.

tests/data/special_character/definitions/variable/tag_fuel.yaml
```diff
@@ -0,0 +1,3 @@
+- Fuel:
```
phackstock, line 1: Don't think you need the tag codelist for testing for hidden characters.

tests/data/special_character/definitions/scenario/scenarios.yaml
```diff
@@ -0,0 +1,4 @@
+- Scenario1
```
phackstock, line 1: If this codelist is not used in the test you can remove this file altogether.

tests/data/special_character/definitions/variable/variables.yaml
```diff
@@ -0,0 +1,9 @@
+- Primary Energy:
```
phackstock, line 1: I would shorten this file to include maybe only the variable with the hidden character. Also I don't think the tags are needed.

phackstock (changes requested):
Solid start to the PR, some comments below.

phackstock (approved):
Looks all good to me.

tests/data/hidden_character/definitions/scenario/scenarios.yaml
```diff
@@ -0,0 +1,4 @@
+- Scenario1
+- Scenario2 
+- Scenario3 
+- Scenario4
```
danielhuppmann, line 4: Please add an empty line at end of file

nomenclature/testing.py
```diff
@@ -21,6 +24,18 @@ def assert_valid_yaml(path: Path):
             error = True
             logger.error(f"Error parsing file {e}")
 
+        with open(file, "r", encoding="utf-8") as all_lines:
+            # check if any special character is found in the file
+            for index, line in enumerate(all_lines.readlines()):
+                for col, char in enumerate(line):
+                    if char not in printable:
+                        special_characters += (
+                            f"{file.name}, line {index + 1}, col {col + 1}. "
```
danielhuppmann, line 33: Print error message with pointers to invalid characters as list.
```suggestion
                            f"\n - {file.name}, line {index + 1}, col {col + 1}. "
```

danielhuppmann (approved):
Two minor suggestions, then good to be merged!

PR #170, opened 2022-08-26
Features/end whitespace

luciecastella:
Closes second half of #164 

Raises error if a code ends with a whitespace


nomenclature/codelist.py
```diff
@@ -145,6 +145,17 @@ def check_stray_tag(cls, v):
                 )
         return v
 
+    @validator("mapping")
+    def check_end_whitespace(cls, v):
+        """Check that no code ends with a whitespace"""
+        print(v)
```
phackstock, line 151: This can be removed.
```suggestion

```

nomenclature/codelist.py
```diff
@@ -145,6 +145,17 @@ def check_stray_tag(cls, v):
...
+        for code in v:
+            if code.endswith(" "):
+                raise ValueError(
+                    f"Unexpected white space at the end of a code: {code}."
```
phackstock, line 155: Since this error is specifically about white space it might not be a bad idea to put the `{code}` value in quotes:
```suggestion
                    f"Unexpected white space at the end of a code: '{code}'."
```
This way it should be more obvious to the user that there is in fact a white space character at the end of the code.

tests/test_codelist.py
```diff
@@ -90,3 +90,13 @@ def test_stray_tag_fails():
         CodeList.from_directory(
             "variable", TEST_DATA_DIR / "stray_tag" / "definitions" / "variable"
         )
+
+
+def test_end_whitespace_fails():
+    """Check that typos in a tag raises expected error"""
+
+    match = "Unexpected white space at the end of a code: scenario2"
```
phackstock, line 98: Since you are testing that the white space is correctly caught I would include it in the error message as the faulty scenario is called `'scenario2 '` and not just `'scenario2'`. If you include the suggestion from above you could test the error message like this:
```suggestion
    match = "Unexpected white space at the end of a code: 'scenario2 '"
```

phackstock (changes requested):
Looks pretty good to me.
Three small comments in line.

nomenclature/codelist.py
```diff
@@ -149,6 +149,16 @@ def check_stray_tag(cls, v):
                 )
         return v
 
+    @validator("mapping")
+    def check_end_whitespace(cls, v):
+        """Check that no code ends with a whitespace"""
+        for code in v:
+            if code.endswith(" "):
+                raise ValueError(
+                    f"Unexpected white space at the end of a code: '{code}'."
```
danielhuppmann, line 158: If you add `values` to the function call, you can make this error message more helpful
```suggestion
                    f"Unexpected whitespace at the end of a {values['name']} code: '{code}'."
```

danielhuppmann (approved):
Looks good, thanks @luciecastella

phackstock (approved):
Looks good to me as well.

PR #172, opened 2022-08-29
Codelist/variable subclass

luciecastella:
Linked to Issue #137 

Add the VariableCodeList subclass for a clearer code.

Issue #137, opened 2022-07-07
Refactor the CodeList class to a generic VariableCodeList class


danielhuppmann:
The **CodeList** class is becoming bloated and hard to read because there are several exceptions and special treatment for variables. We could create a subclass **VariableCodeList** to separate general and dimension-specific features.

phackstock:
I was starting to take a look at this issues yesterday and wanted to share some thoughts before getting started:

* Maybe we can drop the usage of the the `Code` class. The reason for suggesting this is that, to me somewhat counterintuitively, a `CodeList` does not actually contain `Code` instances. The `Code` class is merely used for parsing and for replacing the tags. The former use could in my opinion be replaced by a per-item validator in the CodeList class. The tag class is a little trickier but only keeping this one should already make the code easier to read and understand.
* What should the `CodeList` (or the proposed derived `VariableCodeList`) contain? 
  * For me it would make sense to create a dedicated `VariableCode` class. In this case a `VariableCodeList` would be a list of `VariableCode` instances. 
  * Not sure if we should extend this to a `RegionCode` for the generic `CodeList` class or not.
  * Additionally for the generic `CodeList` should we stick with the current data type or go the same route as for the `VariableCodeList` and have `CodeList` actually contain a list of `Code` instances?

danielhuppmann:
Per our discussion, there are several steps that should be taken (as separate PRs):
- [ ] Create a sub-class `VariabeCodeList` and move variable-specific code there
- [ ] Create a sub-class `RegionCodeList` and move region-specific code there (in particular with the hierarchy-inversion when reading the yaml files)
- [ ] Investierte strategies to remove the `Code` class (per suggestions above)

phackstock:
Closing as #174 and #177 are merged.

PR #173, opened 2022-08-29
Codelist/region subclass

luciecastella:
Linked to Issue #137 
Follows PR #172


luciecastella:
@phackstock @danielhuppmann 
Question: Should the functions `from_directory()` and `read_excel` output a RegionCodeList instead of a CodeList for regions? The classes do not have any difference so it would add complexity for the same result, but I don't know if it should be done anyway?

luciecastella:
Fixing the commits indicating the merge commit to the main : a new pull request will be created.

PR #178, opened 2022-09-01
Fix RegionAggregationMapping.common_region_names docstring

phackstock:
Noticed yesterday that the docstring of the `RegionAggregationMapping.common_region_names` was wrong.


danielhuppmann (approved):
Looks better...

PR #179, opened 2022-09-05
Quickfix to make yaml-check less stringent

danielhuppmann:
I just noticed that the newly added yaml-validation against illegal (hidden) characters is too stringent - in the openENTRANCE project, we have (valid) entries in the yaml files that are not "printable", like region names with special characters, see [this example](https://github.com/openENTRANCE/openentrance/blob/bc6a4fe490610246a9fe5cb7d47c95702ec54fd9/definitions/region/nuts1.yaml#L487).

This PR adapts the validation to only check against a list of specific illegal characters and changes the error type to AssertionError for consistency).

(We may need to add other illegal characters over time)

### DIFF
--- a/nomenclature/testing.py
+++ b/nomenclature/testing.py
@@ -2,11 +2,12 @@ import yaml
 import logging
 from pathlib import Path
 from typing import List, Optional
-from string import printable
 
 import nomenclature
 
 logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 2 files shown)


phackstock (approved):
Looks good to me.

luciecastella (approved):
Yes looks great thank you!

PR #185, opened 2022-10-12
Add a `to_csv()` method

danielhuppmann:
Per a suggestion by @gunnar-pik at the NAVIGATE consortium meeting to have codelists as csv and then run a simple diff on them, this PR adds a `to_csv()` method (and also `to_pandas()` as a utility).

FYI @khaeru

### DIFF
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -302,7 +302,7 @@ class CodeList(BaseModel):
         return codelist
 
     def to_csv(self, path=None, sort_by_code=False, **kwargs):
-        """Write the codelist to an Excel spreadsheet
+        """Write the codelist to a comma-separated values (csv) file
 
         Parameters
         ----------
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -280,12 +280,12 @@ class CodeList(BaseModel):
         else:
             return stream
 
-    def to_pandas(self, sorted: bool=False) -> pd.DataFrame:
+    def to_pandas(self, sort_by_code: bool = False) -> pd.DataFrame:
         """Export the CodeList to a :class:`pandas.DataFrame`
 
         Parameters
         ----------
-        sorted : bool, optional
+        sort_by_code : bool, optional
             Sort the codelist before exporting to csv.
         """
         codelist = (
@@ -294,14 +294,14 @@ class CodeList(BaseModel):
             .rename(columns={"index": self.name})
             .drop(columns="file")
         )
-        if sorted:
+        if sort_by_code:
             codelist.sort_values(by=self.name, inplace=True)
         codelist.rename(
             columns={c: str(c).capitalize() for c in codelist.columns}, inplace=True
         )
         return codelist
 
-    def to_csv(self, path=None, sort_by_code=False, **kwargs):
+    def to_csv(self, path=None, sort_by_code: bool = False, **kwargs):
         """Write the codelist to a comma-separated values (csv) file
 
         Parameters
@@ -314,10 +314,16 @@ class CodeList(BaseModel):
             Sort the codelist before exporting to csv.
         **kwargs
             Passed to :meth:`pandas.DataFrame.to_csv`.
+
+        Returns
+        -------
+        None or csv-formatted string (if *path* is None)
         """
         return self.to_pandas(sort_by_code).to_csv(path, **kwargs)
 
-    def to_excel(self, excel_writer, sheet_name=None, sort_by_code=False, **kwargs):
+    def to_excel(
+        self, excel_writer, sheet_name=None, sort_by_code: bool = False, **kwargs
+    ):
         """Write the codelist to an Excel spreadsheet
 
         Parameters
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -280,7 +280,7 @@ class CodeList(BaseModel):
         else:
             return stream
 
-    def to_pandas(self, sorted=False):
+    def to_pandas(self, sorted: bool=False) -> pd.DataFrame:
         """Export the CodeList to a :class:`pandas.DataFrame`
 
         Parameters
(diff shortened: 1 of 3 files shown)


nomenclature/codelist.py
```diff
@@ -280,34 +280,70 @@ def to_yaml(self, path=None):
...
+            If *None*, the result is returned as a csv-formatted string.
+            See :meth:`pandas.DataFrame.to_csv` for details.
+        sort_by_code : bool, optional
+            Sort the codelist before exporting to csv.
+        **kwargs
+            Passed to :meth:`pandas.DataFrame.to_csv`.
+        """
+        return self.to_pandas(sort_by_code).to_csv(path, **kwargs)
+
+    def to_excel(self, excel_writer, sheet_name=None, sort_by_code=False, **kwargs):
```
phackstock, line 320: Maybe this could be reduced a bit more to just take a path and write out the codelist similar to `to_csv`. I'm not sure we need the option to pass and ExcelWriter object and the call to pyam to write the sheet.
But if you want all the fine grained control that's totally fine as well.

nomenclature/codelist.py
```diff
@@ -280,34 +280,70 @@ def to_yaml(self, path=None):
...
-        # format and write the codelist
         codelist = (
             pd.DataFrame.from_dict(self.mapping, orient="index")
             .reset_index()
             .rename(columns={"index": self.name})
             .drop(columns="file")
         )
+        if sorted:
+            codelist.sort_values(by=self.name, inplace=True)
         codelist.rename(
```
phackstock, line 299: Is there a specific reason why we need to capitalize the columns? I was just looking at the unit test and was initially confused why it turned "bool" into "Bool".

phackstock (commented):
Looks good to me overall, some minor comments below.
Just for my understanding, is csv supposed to be supported as an input format in the future as well or is this just for more efficient comparison purposes?

luciecastella (approved):
All good to me

nomenclature/codelist.py
danielhuppmann: First, my philosophy here is to follow pyam, and pyam follows pandas - in the spirit of "least surprise".

Also, if you want to write several codelists from one **DataStructureDefintion** to an Excel file, you could do:

```python
writer = pd.ExcelWriter()
for dim in dsd.dimensions:
    dsd[dim].to_excel(writer)
writer.close()
```

Thanks to your suggestion, I realized that **DataStructureDefintion** also has a `to_excel()` method - I've now streamlined that and use the **CodeList** method directly.

nomenclature/codelist.py
```diff
@@ -280,34 +280,70 @@ def to_yaml(self, path=None):
...
             columns={c: str(c).capitalize() for c in codelist.columns}, inplace=True
         )
-        write_sheet(excel_writer, sheet_name, codelist)
+        return codelist
+
+    def to_csv(self, path=None, sort_by_code=False, **kwargs):
```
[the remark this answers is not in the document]
danielhuppmann: Done

nomenclature/codelist.py
```diff
@@ -280,34 +280,70 @@ def to_yaml(self, path=None):
...
+        """Export the CodeList to a :class:`pandas.DataFrame`
 
         Parameters
         ----------
-        excel_writer : str, path or ExcelWriter object
-            Any valid string path, :class:`pathlib.Path` or :class:`pandas.ExcelWriter`
-        sheet_name : str
-            Name of sheet that will contain the codelist
+        sorted : bool, optional
+            Sort the codelist before exporting to csv.
```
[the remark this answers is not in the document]
danielhuppmann: Done

nomenclature/codelist.py
danielhuppmann: This is a legacy/convenience design decision because the Excel templates used to be capitalized but in the yaml files, we went the more pythonic way of all-lowercase...

phackstock: Ah right, let's not break convention then.

danielhuppmann:
> Just for my understanding, is csv supposed to be supported as an input format in the future as well or is this just for more efficient comparison purposes?

I would not add csv as a supported file-type for importing/parsing right now - the idea was really just to allow users to have an efficient way to text-diff (even though **git blame** might be more efficient)...

nomenclature/codelist.py
phackstock: > First, my philosophy here is to follow pyam, and pyam follows pandas - in the spirit of "least surprise".

Ah yes, good point, keeping the APIs aligned is always a good idea. 

> Thanks to your suggestion, I realized that DataStructureDefintion also has a to_excel() method - I've now streamlined that and use the CodeList method directly.

That's in a different PR though, right? I just checked and I could not find any changes to `definitions.py`.

danielhuppmann: Forgot to push the commit, sorry.

nomenclature/codelist.py
```diff
@@ -280,34 +280,76 @@ def to_yaml(self, path=None):
...
+        sort_by_code : bool, optional
+            Sort the codelist before exporting to csv.
+        **kwargs
+            Passed to :meth:`pandas.DataFrame.to_csv`.
+
+        Returns
+        -------
+        None or csv-formatted string (if *path* is None)
+        """
+        return self.to_pandas(sort_by_code).to_csv(path, **kwargs)
```
phackstock, line 322: I would set the default for `index` in `to_csv` to `False`. Reason being that if a new entry is added that would push all the other entries down, changing their index and produce a misleading diff where it looks like every entry below the new one was changed when actually it was just the index number.
Also as I just checked that `index=False` is the default for `write_sheet` in pyam. Probably also a good idea for that reason to align the behavior.

phackstock (commented):
Looks good to me, just one small comment below.

nomenclature/codelist.py
danielhuppmann: Good catch!

phackstock (approved):
Very nice, good to be merged from my side.

PR #188, opened 2022-11-07
Feature/code usage

phackstock:
Opened from #186 as the conflict resolution broke some tests and some commits will be necessary before the PR can be merged. As I have read but now write access to Lucie's fork I'm continuing this from my own fork.


phackstock:
Finished the clean up, now it should be mostly all good. 

Here's a quick overview of the changes implemented and open questions. 

## Updates

* `CodeList` classes and all derived classes `RegionCodeList` and `VariableCodeList` now contain just a dictionary which holds the `Code` or `VariableCode` objects. 
* The `Code` class was also overhauled to contain two types of attributes. Named attributes such as `name` and `description` and any arbitrary attributes which are saved as a dictionary under `attributes`. 
* As discussed in #189 **both** named and additional attributes can be accessed and set as attributes, e.g. `Code.name` or `Code.iso2`. 
* While making this attribute access and setting work I also had to take a look at the tag replacement routines and streamlined them a bit as well. 
* I generalized the `Code` class so that the only difference between `Code` and `VariableCode` are the additional named attributes for region aggregation. 

## Open questions/possible issues

* While looking at the tag replacement I noticed that there might be an issue in the future. The way that the tag replacement logic is described in the docs is that: 

> When importing a tagged codelist, any occurrence of {Tag} in the name of a code will be replaced by every element in the Tag dictionary. The {Tag} will also be replaced in any of the code attributes.

This is of course fine for the name and the description of a `Code` as these are strings. Already for the `unit` property this could cause an error since a unit can be a list of strings. Looking at further region processing related attributes as well as the fact that we allow additional attributes to have different structures, lists, dicts, etc... further complicates things. 
I've taken a quick look into openentrance to see how tags are used in practice and it seems to me that apart from description they do not contain any other attributes anyway. Should we restrict the tag replacing to the `description` attribute or should we keep it general and risk applying string replace to a list?

* Something that we touched on in #189 is restricting the allowed names for attributes to valid python identifiers. I have not implemented that yet. The issue is that we currently have three named attributes for region aggregation that would violate this: _skip-region-aggregation_, _region-aggregation_ and _check-aggregate_. Right now I kept the workaround that Lucie put in place where we replace the dashes with underscores to get valid identifiers. My preference would be to implement the check that all attributes in the yaml files need to be valid Python identifiers. The downside of that would be that I would need to update all currently used codelists to conform to this standard so that merging this PR does not break the processing. 

------

Sorry for number of changes and wall of text, I really tried keeping the updates as minimal as possible. 
Looking forward to your input @danielhuppmann.

nomenclature/code.py
```diff
@@ -19,47 +35,102 @@ def from_dict(cls, mapping):
...
-        for _key, _value in target.attributes.items():
-            if _key in attrs:
-                attrs[_key] = attrs[_key].replace(f"{{{tag}}}", _value)
 
-        _code = Code(name=key, attributes=attrs)
-        _code_list.append(_code)
+class VariableCode(Code):
 
-    return _code_list
+    unit: Optional[str]
```
danielhuppmann, line 130: This should be required! Even if the variable is dimensionless, the current logic is that a unit attribute has to be explicitly stated in the yaml, e.g.
```yaml
- Some Variable:
    description: some text
    unit: 
```

danielhuppmann (commented):
One import issue in the `VariableCode` is that units are required, plus a suggestion on shorter docstrings. Otherwise, it looks good to me.

On the specific questions:

1. Tag-replacement does not work on lists: True, but not critical - please make an issue. And I would rather keep them generic, not limit them to certain attributes.
2. Name attributes with "-": I suggest to do the following:
  - merge this PR
  - then implement the check for valid Python attributes except the three region-attributes (with the renaming in place)
  - then have a discussion if we want to maintain the current syntax/workaround or change all repos

nomenclature/codelist.py
```diff
@@ -41,43 +41,21 @@ def read_validation_schema(i):
...
 
-    Parameters
+    Attributes
     ----------
     name : str
         Name of the CodeList
-    mapping : dict, list
-        Dictionary or list of Code items
+    mapping : dict
+        Dictionary linking a Code to its name
```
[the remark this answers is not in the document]
phackstock: Ah yes I didn't go over the docstrings again. Thanks for catching that.

nomenclature/code.py
phackstock: Made it mandatory as described here: https://pydantic-docs.helpmanual.io/usage/models/#required-optional-fields.

phackstock:
Made the unit attribute mandatory and added a test for it.
Added alias values for the three hyphenated region aggregation attributes which makes `Code.from_dict` a lot cleaner. Added a test for that as well.
Another side benefit (although for another PR) is that now that all the `VariableCode` keywords are fixed we don't need the validation schema for the variable anymore.

nomenclature/code.py
```diff
@@ -1,65 +1,145 @@
...
+
+    def replace_tag(self, tag: str, target):
+        """Return a new instance with tag applied
+
+        Parameters
+        ----------
+        tag : str
+            Name of the tag
+        target : _type_
+            Code that is inserted
```
danielhuppmann, line 85:
```suggestion
            Code attributes to be modified by the tag
```

nomenclature/code.py
```diff
@@ -1,65 +1,145 @@
...
+
+        Returns
+        -------
+        _type_
+            new instance with occurrences of "{tag}" replaced by target
```
danielhuppmann, line 90:
```suggestion
            New instance with occurrences of "{tag}" replaced by target
```

nomenclature/code.py
```diff
@@ -1,65 +1,145 @@
...
+
+    @property
+    def contains_tags(self) -> bool:
+        return re.search("{.*}", self.name) is not None
+
+    @property
+    def tags(self):
+        return re.findall("(?<={).*?(?=})", self.name)
+
+    def replace_tag(self, tag: str, target):
```
danielhuppmann, line 77: target is always a Code, right? (Used to be a dict?)

tests/alias_test.py
```diff
@@ -0,0 +1,15 @@
...
+
+class TestModel(pydantic.BaseModel):
+    some_attribute: str = pydantic.Field(alias="some-attribute")
+
+    class Config:
+        allow_population_by_field_name = True
+
+
+tm1 = TestModel(some_attribute="test1")
+print(tm1)
```
danielhuppmann, line 12: Remove prints

danielhuppmann (approved):
A few minor comments on the docstrings, then good to go as far as I can tell...

Thanks @phackstock!

tests/alias_test.py
phackstock: Ah sorry for that, this was just a quick test for setting attributes with alias. This was not meant for upload, I'll correct it.

nomenclature/code.py
phackstock: Yes target is always a `Code`. Since this is in the class definition of `Code` itself, just putting `target: Code` did not work. Luckily there's PEP484 to the rescue https://peps.python.org/pep-0484/#forward-references. Added the type hints now.

PR #195, opened 2022-11-11
Simplify Code.attributes type hint

### LINKED ISSUE #191: Simplify Code.attributes type hint
Now that `Code.attributes` is only used to hold additional dict style named attributes and no longer values related to region processing we should be able to simplify the type hint from:
https://github.com/IAMconsortium/nomenclature/blob/70cbe0b3651a486d3fabdc52e3af2e0ed8be0347/nomenclature/code.py#L11-L28

```python
attributes: Dict = {}
```

phackstock:
closes #191.

Mostly self explanatory, since `Code.attributes` does no longer need to support more complex region aggregation related data structures we can simplify the type hint.

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -9,7 +9,7 @@ class Code(BaseModel):
 
     name: str
     description: Optional[str]
-    attributes: Dict[str, Any] = {}
+    extra_attributes: Dict[str, Any] = {}
 
     @classmethod
     def from_dict(cls, mapping) -> "Code":
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -9,7 +9,7 @@ class Code(BaseModel):
 
     name: str
     description: Optional[str]
-    attributes: Dict = {}
+    attributes: Dict[str, Any] = {}
 
     @classmethod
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Just to be sure, this typehint has no functional implications, right?

phackstock:
> Just to be sure, this typehint has no functional implications, right?

In the pure python sense this does not have any implications. By using pydantic it _can_ have some functional implications. In the order of definition of options for data types in Union, pydantic will attempt type conversion.

As an example if we have:

```python
from pydantic import BaseModel

class A(BaseModel):
    a: Union[int, str]

instance_A = A(a = "3") 
print(type(A.a)) >>> "int"
```

pydantic converts the string "3" into the integer 3. 

This might have actually had a functional use in the past where we had attributes such as `check-aggregate` in the `attributes` object. Now that this is only used to hold additional, non-nomenclature functional information, it _should_ not have any functional implications.

danielhuppmann (approved):
Thanks @phackstock, looks good!

phackstock:
Seems like resolving the merge conflicts blew up something, I'll investigate.

PR #196, opened 2022-11-11
Check names of additional attribute keys

### LINKED ISSUE #190: Restrict attribute values
As discussed in #188 we want to restrict the names of the additional attributes to only be valid python identifiers so that they can be accessed like this:

```python
Code.some_additional_attribute
```

therefore a check is needed for additional attributes preventing attributes like `some-additional-attribute`, `2-more-attributes`, etc...

phackstock:
Closes #190.

As discussed in #188, this PR adds a validator for the keys of `Code.attribute` and raises an error if any of them are not valid python identifiers or shadow keywords. 
Examples that are no longer allowed would be `"contains-hyphen"` or `"True".`

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -12,7 +12,7 @@ class Code(BaseModel):
     description: Optional[str]
     extra_attributes: Dict[str, Any] = {}
 
-    @validator("attributes")
+    @validator("extra_attributes")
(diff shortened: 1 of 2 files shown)


danielhuppmann (approved):
Looks good, thanks!

PR #199, opened 2022-11-15
Apply tag replacing to named attributes as well

phackstock:
As changed in 70cbe0b3651a486d3fabdc52e3af2e0ed8be0347, the tag replacing would only be applied to `name`, `description` and additional attributes. 
This PR applies it to all string attributes found.
This should also fix openENTRANCE/openentrance#234.

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -93,15 +93,19 @@ class Code(BaseModel):
         mapping = {
             key: value for key, value in self.dict().items() if key != "attributes"
         }
+        # replace name and description
         mapping["name"] = mapping["name"].replace("{" + tag + "}", target.name)
         mapping["description"] = mapping["description"].replace(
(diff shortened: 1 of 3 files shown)


danielhuppmann (approved):
Fine to be merged to get openENTRANCE operational, but I think we should discuss whether the `attributes`-attribute is a smart choice because it can become confusing...

phackstock:
Fair point, I'm not 100% happy with it either. At some point I was thinking something along the lines of `additional_attributes` but did not end up changing it as the name was a bit long. If you have a preference for a new name I can change that in a quick follow-up PR.

danielhuppmann:
Maybe `custom_attributes` or `extra_attributes`? It's still a bit long, but shorter (and slightly more intuitive) than "additional"...

PR #201, opened 2022-11-16
Improve tag replacement

phackstock:
This PR implements an improved tag replacement logic. 
The new logic is:
* Find all occurrences of `"{tag}"`
* If the tag has the same attribute as the base code, replace the value of the attribute
* If the tag does not have the same attribute, replace with the name of the tag. 

Also, since I added the property `Code.flattened_dict` I also made sure to use it in `CodeList.codelist_repr`.

### DIFF
--- a/doc/source/user_guide/codelist.rst
+++ b/doc/source/user_guide/codelist.rst
@@ -50,7 +50,7 @@ to be used as tag. The files defining the tags must be named like ``tag_*.yaml``
      - Some Key:
          description: a short description of the key
 
-When importing a tagged codelist, code containing ``{Tag}`` in its name of a code will
+When importing a tagged codelist, any code containing ``{Tag}`` in its name will
 be replaced by the name of the corresponding tag. In addition any code attributes
 containing ``{Tag}`` will also be replaced. If the corresponding tag contains the attribute, this will be used for substitution, otherwise the tag name is used.
 
--- a/doc/source/user_guide/codelist.rst
+++ b/doc/source/user_guide/codelist.rst
@@ -50,10 +50,11 @@ to be used as tag. The files defining the tags must be named like ``tag_*.yaml``
      - Some Key:
          description: a short description of the key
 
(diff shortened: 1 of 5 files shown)


doc/source/user_guide/codelist.rst
```diff
@@ -50,10 +50,11 @@ to be used as tag. The files defining the tags must be named like ``tag_*.yaml``
      - Some Key:
          description: a short description of the key
 
-When importing a tagged codelist, any occurrence of ``{Tag}`` in the name of a code will
-be replaced by the corresponding element in the Tag dictionary. If the tag does not contain an element with ``{Tag}``, the tag name is used.
+When importing a tagged codelist, code containing ``{Tag}`` in its name of a code will
```
danielhuppmann, line 53:
```suggestion
When importing a tagged codelist, any code containing ``{Tag}`` in its name will
```

tests/data/tagged_codelist/tag_sector.yaml
```diff
@@ -6,4 +6,3 @@
       weight: Energy
```
danielhuppmann, line 6: Are the the "value"-entries actually used anywhere?

danielhuppmann (approved):
Looks good, just two minor suggestions!

tests/data/tagged_codelist/tag_sector.yaml
phackstock: What do you mean by the "value" entries?

danielhuppmann: There are two attributes "value_bool" and "value_number" just above the commented line...

phackstock: Ah right, sorry didn't see those. I don't think they're used, I'll double check. Should I remove them if they're not used?

phackstock: Removed it, didn't break the tests on my machine. If they also run through here I'll go ahead with the merge.

PR #202, opened 2022-11-23
Correct test pypi installation test

phackstock:
This is a tiny correction to DEVELOPING.rst. As the tests are not bundled with the package you cannot check that the tests are working.


danielhuppmann (approved):
Good catch! Maybe create an issue so that this can be tackled at some point?

phackstock:
> Maybe create an issue so that this can be tackled at some point?

Do you mean we should bundle the tests with the package? I was under the impression that that's not standard practice.

danielhuppmann:
Ok, then let's not do this...

phackstock:
Alright, sounds good. I just double checked and it does not seem like pyam includes the tests either.

PR #204, opened 2022-11-30
Check against duplicate native region names (and rename-targets)

### LINKED ISSUE #203: RegionProcessor "native-region-rename" drops mappings
While toiling away at the GAINS-linkage workflow, I noticed some unexpected behavior of the nomenclature package.

For GAINS, we need to map model-native regions in a 1-to-n-manner: the values for the MESSAGE region WEU have to be copied/renamed to the GAINS regions Austria, Germany, etc., so the resulting dataframe must have ~181 regions from the MESSAGE-native 11.

My initial plan was to simply use the native-region mappings, see https://github.com/iiasa/engage-gains-mapping/blob/main/mappings/MESSAGE_v1.1.yaml

# Unexpected behavior

The attribute `RegionAggregationMapping.rename_mapping` (see [here](https://github.com/IAMconsortium/nomenclature/blob/57b08f759d7a5ec7e2129355196164d96810ab6d/nomenclature/processor/region.py#L270) only has a single mapping for each native-region.

# Solution 1:

Add a dedicated validation that the section "native_regions" must have unique keys (to avoid misunderstandings like mine in the future), and use the "common-region" section for the GAINS-mapping.

# Solution 2:

Refactor the `rename_mapping` attribute such that it is either a mapping "str->str" if unique or a mapping "str -> list[str]" otherwise (and also refactor the `RegionProcessor.apply` method)...

Any thoughts, @phackstock @peterkolp?

danielhuppmann:
Adds a check against duplicate native-region names in a region-processing mapping.

Closes #203


phackstock (approved):
Looks good, thanks.

PR #205, opened 2022-12-05
Require data

### LINKED ISSUE #126: Support *required* variables and regions
Per a discussion with @znicholls and @lewisjared.

The validation currently only checks that variables/regions/... of a scenario (as an **IamDataFrame**) are defined in the **DataStructureDefinition**. We have use cases for the opposite direction: whether data for a list of required variables/regions exists in the scenario.

Related: pyam has a method [`require_variable()`](https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html#pyam.IamDataFrame.require_variable) - but this only checks that data for that variable exist for any region/year...

There are several possible "levels":

1. Require that (at least some data for) all variables and regions exist in the **IamDataFrame**.
2. Add an attribute `required` to the variable attributes that have special treatment:
   ```yaml
   - Allowed variable name:
       description: A short explanation or definition
       unit: A unit
       required: true
       <other attribute>: Some text, value, boolean or list (optional)

3. Allow a dictionaryas value for the `required` attribute for a more detailed configuration of "required", e.g., for specific regions or years...

phackstock:
closes #126.

This PR implements machinery to support requiring data.

## Changes implemented

* `RequireDataValidator` is the main class for the require data feature.
* It contains a ~~two lists~~ a list of `RequiredData` objects. ~~One for required data and one, optional one, for optional data. The difference being that as~~ absence of required data will raise an error in apply ~~while for optional data only a warning is issued.~~ 
 
---
_these were open questions from when the PR was started keeping them for reference_
## ~~Open questions~~

* The way that I would implement it consists of three classes: `RequiredTS` for a single required timeseries, `RequiredTSConfig` which bundles together a name attribute and a list of `RequiredTS` and `RequiredTSProcessor` which bundles together a list of all `RequiredTSConfig`s and performs the validation on an incoming `IamDataFrame`.  Does that sound like a good outline of the proposed classes? I don't like the term _workflow_ too much since to me it sounds ambiguous as we also call an entire project specific repository, for example, _engage-workflow_. This is why I went with `RequiredTS`. 
* How should this `RequiredTSProcessor` be included in the nomenclature structure? In principle I like the idea of using something like `nomenclature.process(df, dsd, processor=[region_processor, required_timeseries_processor])` but then the success or failure could depend on the order of applying the processors. E.g. if a required time series is only created via region aggregation it would fail if we apply the required_timeseries_processor before the region_processor. Conversely, aggregating first and then checking the time series might work perfectly fine. I would think that in principle we always want to aggregate first and then check for required timeseries. It might be easier to add another name input variable to `nomenclature.process`, something like `required_ts_processor`. On the other hand, if we plan on adding more different processors in the future it might be nice to use a list (post-processing options come to mind as an option for this).
* Is there a pyam function that can be used for asserting that a specific time series is present? I checked (`pyam.IamDataFrame.require_variable`)[https://pyam-iamc.readthedocs.io/en/stable/api/iamdataframe.html?highlight=require_variable#pyam.IamDataFrame.require_variable], but it seems that this functions checks that the variable in question exists for _any_ of the years provided. I think for the required timeseries feature we would want to assert that the variable exists for _all_ years. Is that assumption correct? If so, should we try to implement a feature like that in pyam or perhaps modify `require_variable` in pyam?
* Speaking of pyam's `require_variable` function, it seems that it currently does not feature inputs for _region_ and _scenario_ values of the required variable. For _region_ I would think so, for _scenario_ I'm not sure as I currently don't see a use case where we would require a variable to be present in one scenario but not in the other. 
* Since pyam's `require_variable` features a _unit_ option I was thinking if we should add that as well to our required time series feature. Perhaps as an optional argument. Maybe that's not necessary though as we already check units using the `VariableCodeList`. I was just pondering the case where a variable has multiple allowed units but we would need one specific one for a post-processing. Maybe that case does not exist though as we would restrict the unit accordingly in the variable code list.

@danielhuppmann would love to hear your thoughts on these questions.


tests/data/requiredTS/requiredTS.yaml
```diff
@@ -0,0 +1,9 @@
+name: MAGICC
+required_timeseries:
+  - variable: Emissions|CO2
+    region: World
+    years: [2020, 2025, 2050, 2075, 2100]
```
danielhuppmann, line 5: I'd use decadal timesteps for the tests because that is more intuitive for ou user base
```suggestion
    years: [2020, 2030, 2040, 2050]
```

nomenclature/processor/requiredTS.py
```diff
@@ -0,0 +1,27 @@
+from pathlib import Path
+from typing import List, Optional, Union
+
+import yaml
+from pydantic import BaseModel
+
+
+class RequiredTS(BaseModel):
```
danielhuppmann, line 8: I'm not a fan of using non-standard abbreviations - maybe call this `RequiredData` (because this is the `data` attribute in an IamDataFrame)?

nomenclature/processor/requiredTS.py
```diff
@@ -0,0 +1,27 @@
...
+
+    variable: Union[str, List[str]]
+    region: Optional[Union[str, List[str]]]
+    years: Optional[Union[int, List[int]]]
+    scenario: Optional[Union[str, List[str]]]
```
danielhuppmann, line 13: The scenario is part of the index, not the IAMC data columns.
```suggestion
```

danielhuppmann (commented):
Thanks for this first step, @phackstock. A few comments inline, general comments below.

1. Class structure

The key question is how to check for required variables *and then call another module*, e.g., MAGICC or GAINS.

Option 1:

Have attributes `execute` to specify another module and `raise` to specify whether to raise an error or (only) print a warning.

Option 2:

```python
try:
    _df = df.process(df, dsd, processor=required_processor)
    magicc_results = run_magicc(_df) 
    df.append(magicc_results)
except Exception as e:
    logger.warning(e.message)
```

I think option 2 is more versatile and pythonic - so I'd rather use a class name like **RequiredValidator** to be more generic (we may require certain meta indicators).

2. Passing a list of processors

I think that passing the order of processors in the list is a feature, not a problem.

3. A pyam function for validation

Indeed, the pyam method `require_variable()` does not yet support checking for a combination of variables, regions and years. There is a loosely related issue for the `categorize()` method, see https://github.com/IAMconsortium/pyam/issues/61.

4. Checking for units

I can imagine a use case where we allow multiple units for the generic validation but can only run a specific post-processing with a subset of those units. So I would implement the "required" attributes completely generic, including a *unit* option.

nomenclature/processor/requiredData.py
```diff
@@ -0,0 +1,140 @@
...
+    variable: List[str]
+    region: Optional[List[str]]
+    year: Optional[List[int]]
+    unit: Optional[str]
+
+    @validator("variable", "region", "year", pre=True)
+    def single_input_to_list(cls, v):
+        return v if isinstance(v, list) else [v]
+
+    def validate_with_definition(self, dsd: DataStructureDefinition) -> None:
```
danielhuppmann, line 29: This is a duplication of the function in the RegionProcessor, right? I guess you thought about streamlining and had a good reason for not doing it, right?

danielhuppmann (commented):
Please also add validation of the required-data yaml files to https://github.com/IAMconsortium/nomenclature/blob/01c1e42496eed3dd33792e133262f27aa8bdfd1d/nomenclature/testing.py#L49

nomenclature/processor/requiredData.py
phackstock: It's less of a duplication but more implementing functionality that's also needed here. The logic is that `requiredData` and as a result `requiredDataValidator` should not require anything that's not covered by the codelists as it would otherwise be impossible to satisfy. E.g. if in `requiredData` we require the variable _Final Energy_ to be present for the region _World_ both must exist in the project's codelists.
`requiredDataValidator` also has `validate_with_definition` which loops over each `requiredData` element and validates it.

danielhuppmann: Maybe move this as a validation-method into the codelist and have the processors pass their lists to that method?

phackstock: Ah sure, that works too. Should I do the same thing for the `RegionProcessor` as well?

danielhuppmann: Yes, I think that would simplify the package... Thanks!

phackstock: Sounds good, I'll do that.

nomenclature/processor/requiredData.py
```diff
@@ -0,0 +1,140 @@
...
+                logger.error(
+                    f"Required data {data} from file {get_relative_path(self.file)} "
+                    f"missing for:\n{missing_index}"
+                )
+        if error:
+            raise RequiredDataMissingError(
+                "Required data missing. Please check the log for details."
+            )
+
+        # check for optional data and issue warning if missing
```
danielhuppmann, line 108: I think it's confusing to have "optional data" within a validation for required data - and why print a warning if optional data isn't there?

danielhuppmann (commented):
Reminder - please also add validation of the required-data yaml files to https://github.com/IAMconsortium/nomenclature/blob/01c1e42496eed3dd33792e133262f27aa8bdfd1d/nomenclature/testing.py#L49

nomenclature/processor/requiredData.py
phackstock: Fair point, the reason I put it there is that I was thinking along the lines of a use case where we have a core set of data for a given post-processing plus an additional set of optional data. Having these optional data is desired as it makes the post-processing more accurate and requires less infilling, like the ar6 climate assessment workflow does it. 
Now that I think about it again it is probably best though to cover this logic in a `MAGICCPostProcessor` class and leave `requiredData` more simple.

danielhuppmann (commented):
This comment wasn't included in the review yesterday for some reason

nomenclature/error/requiredData.py
```diff
@@ -0,0 +1,2 @@
+class RequiredDataMissingError(ValueError):
```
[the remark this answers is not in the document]
phackstock: According to PEP8 for class names this style cap words is recommended https://peps.python.org/pep-0008/#class-names. For example as in `ValueError.`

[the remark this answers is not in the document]
danielhuppmann: But the file name should be mimicking the module name, not the class name - and that should be all lower-case... https://peps.python.org/pep-0008/#package-and-module-names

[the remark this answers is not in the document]
phackstock: Ah sorry I over read the file part. Updated the files accordingly.

phackstock:
### A few updates to the PR:

* dropped `optional_data` from `RequiredDataValidator`. Functionality to recommend having data beyond the minimum requirement to make post-processing results more accurate is most likely best covered in a dedicated post-processor class.
* renamed modules from `requiredData` to `requireddata`
* added validation of `RequiredDataValidator` to `assert_valid_structure`, along with a test and cli support.
* use the new `DataStructureDefinition.invalid_items` in validating `RegionProcessor` as well.
* changed `RequiredDataValidator.apply` to return `None` in case of success, before it returned the data frame, to be debated though if `None` or the data itself is better

### Comments

For validating all `RequiredDataValidator` yaml files in `assert_valid_structure` I opted to collect all errors to report all issues at once. Since one yaml file maps to one `RequiredDataValidator`, as opposed to one `RegionProcessor` reading **all** region mappings I opted for the following structure:

```python
    errors: List[str] = []
    for file in (requiredDatadir).iterdir():
        try:
            RequiredDataValidator.from_file(file).validate_with_definition(dsd)
        except pydantic.ValidationError as pve:
            errors.append(str(pve))
    if errors:
        all_errors = "\n".join(errors)
        raise ValueError(f"Found error(s) in required data files: {all_errors}")
```

this might be nice to use in other places of the code where we have want to achieve similar behavior, collect all errors and then raise a single one. One place that comes to mind is `RegionAggregationMapping.validate_with_definition` which I implemented somewhat awkwardly by (mis)using the `ValidationError` from pydantic.

nomenclature/cli.py
```diff
@@ -59,6 +66,8 @@ def cli_valid_project(
         Name of the definitions folder, defaults to "definitions"
     mappings : str, optional
         Name of the mappings folder, defaults to "mappings" (if this folder exists)
+    requiredData: str, optional
```
danielhuppmann, line 69:
```suggestion
    requireddata: str, optional
```

danielhuppmann (commented):
I‘d have slight preference for „required_data“ instead of „requiredata“.

> * changed `RequiredDataValidator.apply` to return `None` in case of success, before it returned the data frame, to be debated though if `None` or the data itself is better

I think that returning the full IamDataFrame would be better so that we can more easily chain up several validators and processors.

phackstock:
> I‘d have slight preference for „required_data“ instead of „requiredata“.

Agreed, that's a lot more readable. I applied the renaming consistently. They input option for the CLI is now called `required-data` instead of `requireddata` and the default option for the directory where the required data files live is now `required_data` (from previously `requiredData`).

> I think that returning the full IamDataFrame would be better so that we can more easily chain up several validators and processors.

Updated as well.

danielhuppmann (approved):

PR #206, opened 2022-12-12
Clean up imports

phackstock:
As I was working on #205 I made some changes to the imports as `processor` is now getting to a place where it makes sense to make it into a proper sub module.
In order to keep #205 minimal I took out all the clean up of the imports into this PR.

Changes include order of imports, order within imports and making inner module imports consistent.


danielhuppmann (approved):
Looks good to me!

Issue #215, opened 2023-01-11
Unify processor interface


phackstock:
We want to implement a modular processor pipeline design for the `process` function. As of right now it takes a single `RegionProcessor` object as an optional input. 
This should change to accept a list of `Processor` objects which can be `RegionProcessor` or `RequiredDataValidator` (for now). 
These `Processor` objects will implement a common interface method called `apply(df: pyam.IamDataFrame)`. In the future, post-processing options such as climate models (MAGICC, FAIR, etc...) which also inherit from `Processor` are planned.

In order to get there a number of changes are required:

* `RegionProcessor.apply(df, dsd)` currently takes a `DataStructureDefinition` argument as input as this is needed for region aggregation while `RequiredDataValidator.apply(df)` only uses a `pyam.IamDataFrame`. This can be mitigated by adding the `DataStructureDefinition` as a class attribute of `RegionProcessor`.
* Once the interfaces of `.apply()` are aligned a common parent class `Processor` can be introduced. 
* The `process` function needs to be adapted to work with a number of `Processor` objects.

PR #216, opened 2023-01-11
Refactor RegionProcessor apply interface

phackstock:
Addresses first point in #215.

As per discussion yesterday this PR is the first step towards allowing `process` to take a list of `Processor` instances. There are more steps required but I opted to go for multiple PRs to keep them on the smaller side. 

### Implemented changes

* `RegionProcessor.apply()` now only takes a `pyam.IamDataFrame` instance as an input as opposed to also needing a `DataStructureDefinition`.
* Instead the `DataStructureDefinition` is now handed to the `RegionProcessor` at creation as part of the `from_directory()` classmethod. I opted to go for fully instantiated `DataStructureDefinition` instead of a path to the definitions directory so that any errors that may occur related to the `DataStructureDefinition` not being valid, occur before the `RegionProcessor` can be instantiated. I.e. I opted for:

```python
dsd = DataStructureDefinition("definitions/") # error occurs here
rp = RegionProcessor("mappings/", dsd)
```
instead of:

```python
# error related to DataStructureDefinition would still occur in RegionProcessor
# which is less straightforward to read
rp = RegionProcessor("mappings/", "definitions/") 
```

* I chose to not include the entire `DataStructureDefinition` in the `RegionProcessor` class but only the required RegionCodeList and VariableCodeList to keep the class to the required minimum of information needed.
* Since the `RegionProcessor` object now contains codelists the `validate_with_definition` method no longer needs a `DataStructureDefinition` as input. Therefore I made it into a proper pydantic validator that is always executed at object creation. 

### Open questions/considerations

* I included a check for the provided `DataStructureDefinition` in `RegionProcessor.from_directory` for the required variable and region codelists. An error is raised if either is missing. Technically, this is a slight deviation from the previous implementation. Previously if no region codelist was given not error was thrown and region validation was simply skipped. I could of course change it back to the old style but I think the case where we have a `RegionProcessor` but no region codelist either does not exist or should be avoided anyway.

* Since `RegionProcessor.validate_with_definition` is now a pydantic validator it is called upon object creation. This means that a faulty model mapping will cause an upload to crash. This will be the case even if the faulty model mapping applies to a different model than is in the upload. This can also be changed back but since we have integration tests in place anyway this case should not occur. I can also revert it back to the old style though.

* Merging this PR into `main` would crash all current project workflows using region processing as the `RegionProcessor` now needs to be instantiated with a `DataStructureDefinition`. This is of course not a problem, I could simply update them all but this PR is part of a set of planned PRs that will change the interface of at least the `process` function as well. So in order to avoid updating all workflows after merging this PR and then doing essentially the same thing again I was wondering if it might be a good idea to create a temporary `dev` branch where I'd merge all PRs that concern the implementation of the new processing pipeline and once it's all done I only have to update all the workflows once.

Looking forward to your input @danielhuppmann.

### DIFF
--- a/nomenclature/testing.py
+++ b/nomenclature/testing.py
@@ -56,11 +56,9 @@ def _check_mappings(
     dsd = DataStructureDefinition(path / definitions, dimensions)
     if mappings is None:
         if (path / "mappings").is_dir():
-            RegionProcessor.from_directory(path / "mappings").validate_with_definition(
-                dsd
-            )
+            RegionProcessor.from_directory(path / "mappings", dsd)
     elif (path / mappings).is_dir():
-        RegionProcessor.from_directory(path / mappings).validate_with_definition(dsd)
+        RegionProcessor.from_directory(path / mappings, dsd)
     else:
         raise FileNotFoundError(f"Mappings directory not found: {path / mappings}")
 
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -39,9 +39,9 @@ def test_region_processing_rename(model_name):
 
     obs = process(
         test_df,
-        DataStructureDefinition(TEST_DATA_DIR / "region_processing/dsd"),
+        dsd := DataStructureDefinition(TEST_DATA_DIR / "region_processing/dsd"),
         processor=RegionProcessor.from_directory(
-            TEST_DATA_DIR / "region_processing/rename_only"
+            TEST_DATA_DIR / "region_processing/rename_only", dsd
         ),
     )
 
@@ -67,8 +67,8 @@ def test_region_processing_empty_raises(rp_dir):
     with pytest.raises(ValueError, match=("'model_a', 'model_b'.*empty dataset")):
         process(
             test_df,
-            DataStructureDefinition(TEST_DATA_DIR / "region_processing/dsd"),
-            processor=RegionProcessor.from_directory(TEST_DATA_DIR / rp_dir),
+            dsd := DataStructureDefinition(TEST_DATA_DIR / "region_processing/dsd"),
+            processor=RegionProcessor.from_directory(TEST_DATA_DIR / rp_dir, dsd),
         )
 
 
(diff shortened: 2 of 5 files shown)


tests/test_core.py
```diff
@@ -39,9 +39,9 @@ def test_region_processing_rename(model_name):
 
     obs = process(
         test_df,
-        DataStructureDefinition(TEST_DATA_DIR / "region_processing/dsd"),
+        dsd := DataStructureDefinition(TEST_DATA_DIR / "region_processing/dsd"),
         processor=RegionProcessor.from_directory(
-            TEST_DATA_DIR / "region_processing/rename_only"
+            TEST_DATA_DIR / "region_processing/rename_only", dsd
         ),
     )
```
danielhuppmann, line 46: A more readable approach would be
```python
dsd = DataStructureDefinition(TEST_DATA_DIR / "region_processing/dsd"),
region_processor = RegionProcessor.from_directory(
             TEST_DATA_DIR / "region_processing/rename_only", dsd
)
process(test_df, dsd, region_processor)
```

nomenclature/testing.py
```diff
@@ -56,11 +56,9 @@ def _check_mappings(
     dsd = DataStructureDefinition(path / definitions, dimensions)
     if mappings is None:
```
danielhuppmann, line 57: Maybe streamline to 
```python
mappings = mappings or "mappings"
if (path / mappings).is_dir():
         RegionProcessor.from_directory(path / mappings, dsd)
...
```

danielhuppmann (approved):
A few suggestions inline for easier readability, but looks great overall!

nomenclature/testing.py
phackstock: I just implemented your suggestion and to my surprise it broke a unit test `test_cli_missing_mappings_runs`. 
After a little bit of digging if found the that this slightly changes the logic. 
Previously if no mappings directory was given and none exists no error would be raised. Now it would raise an error as after the first line mappings would be "mappings" even if none was provided.
Even though I like your suggestion in terms of readability I'd keep it as is to preserve the logic.

tests/test_core.py
phackstock: Ah yes thanks, that's much cleaner.

nomenclature/testing.py
danielhuppmann: Ha, thanks, very helpful!

PR #218, opened 2023-01-12
Serialize and read complex attributes

### LINKED ISSUE #217: VariableCode ValidationError when generating variable template
Trying to edit/add to the `engage_internal` variable template, but info that is already there, regarding the `Region_aggregation` and `Weight` is causing an issue e.g. Documented here in https://github.com/iiasa/engage-internal-workflow/pull/62

Removing both of those two columns resolves the issue.

Files included in the PR should recreate the issue. @danielhuppmann had a try already but says is issue in nomenclature for more complex attributes

https://github.com/iiasa/engage-internal-workflow/blob/5e76f80be4e4bef402721080322e450871560c69/definitions/Tutorial_edit_existing_variable_codelist.ipynb

N.B the excel file included in the above PR, already has the edits / additional variables that I would like to include. But they are not what are causing the issue (e.g. the Region agg and weight info for `Price|Carbon` was already there)

### LINKED ISSUE #220: Reading complex attributes from Excel
This issues addresses the first problem mentioned in #217.
Any codelist entry that is more complex than a mapping to a single value, e.g. `region-aggregation` in:

```yaml
- Price|Carbon:
    definition: Price of carbon (as reported by the model)
    unit: US$2010/t CO2
    region-aggregation:
        - Price|Carbon (mean):
            method: mean
```

will be exported to Excel using `CodeList.to_excel` in the following way:

| Variable      |                 Description                | Unit          | Region-aggregation                             |   |   |
|---------------|:------------------------------------------:|---------------|------------------------------------------------|---|:-:|
| Price\|Carbon | Price of carbon (as reported by the model) | US$2010/t CO2 | [{'Price\|Carbon (mean)': {'method': 'mean'}}] |   |   |

when reading it back from the Excel file into a `CodeList` we get an error as the string under _Region-aggregation_ cannot be interpreted as the structure above.

This should be fixable by using json serialization. Any attribute value that is a list or a dict, should be serialized as json during `CodeList.to_pandas`. As it is read it again using a pydantic pre validator on those fields the values can be read again correctly.
FYI @byersiiasa

phackstock:
Closes #220, #219, #217, #222, #223.

## Update in the comments below (original text of the PR kept for the record)

I took a look at reproducing the error pointed out by @byersiiasa. The issue is that in `VariableCode` we have two, potentially three attributes (`region_aggregation`, `components` and potentially `unit`) that are more complex than a single value: 

https://github.com/IAMconsortium/nomenclature/blob/387cb8aa9f3f62d0186251f830d475919f437618/nomenclature/code.py#L126-L139

For now I have added a failing unit test in `test_create_yaml_from_xlsx` to mark the issue. The crux of the problem is that expressing complex data structures is Excel is not really possible without things getting messy. In the example of `region_aggregation` the data structure we would need to somehow encode in Excel is `List[Dict[str, Dict]]`. 
To me the only sensible fix for this would be to explicitly exclude those complex columns from xlsx to yaml conversion. 
Alternatively we could think of allowing users to write yaml or json as a string into an Excel spreadsheet but at that point I think it's easier to just use yaml directly

Thinking about the issue a bit more, this topic is probably not restricted to `create_yaml_from_xlsx` but to the various `CodeList` related I/O functions we have namely `CodeList.to_pandas`, `CodeList.to_csv` and `CodeList.to_excel`. Although I haven't checked explicitly how complex attributes are encoded there it's probably worth taking a look.


phackstock:
This PR implements a number of fixes that are all connected.

### Changes implemented

* When calling `CodeList.to_pandas()` any `Code` attribute that is a list or a dictionary is json serialized before it is written into the pandas DataFrame.
* `VariableCode` now has a validator called `deserialize_json` for the attributes _unit_, _region_aggregation_ and _components_. This validator is called before the attributes are type checked. If the provided value is a string, json-deserialization is attempted. If the value is not a string it is just used normally. This is to allow for values to be provided as the proper python data type or a json string.
* Added a test `test_set_attributes_with_json` to verify this behavior. Importantly it tests that this deserialization plays well with _unit_ which can be a string or a list of strings.
* As testing a "round trip" (read variable code list from yaml, safe to Excel and read again from Excel) as @byersiiasa was trying to do was not really possible without fixing the nan issue (#219) I added `codelist = codelist.replace(np.nan, None)` to `CodeList.read_excel()`. 
When trying to make this work I ran into a pandas bug that was only recently fixed in version 1.5.2 (see the second point in the "Fixed regressions" section https://pandas.pydata.org/docs/whatsnew/v1.5.2.html#fixed-regressions). For this reason I added the pandas version constraint to `setup.cfg`. 
Since I also needed `np.nan` I added numpy as an explicit dependency as well.
* In `CodeList.read_excel()` I restricted the columns that are read from the file in the beginning of the function to `[col]+attrs` as per definition no other columns are needed.
* One change that retracted again was adding file information into `CodeList.read_excel()`. I would think it is useful to have this though as any errors raised when creating a CodeList require the file attribute to be present. If this is not there and there is a (different) error in the CodeList we run into an error while trying to report an error. The reason why I left it out for now is that it broke `test_create_yaml_from_xlsx`. Reason being that `CodeList.to_yaml` includes the file attribute when writing the codelist to a file which breaks the comparison with the reference file. Maybe `to_yaml` should also exclude the _file_ attribute though as `to_excel` and `to_csv` do.

@byersiiasa this should fix your issue.

byersiiasa:
wow - thanks @phackstock that's a huge amount of of digging you have done, much appreciated! Will try to check this today.

phackstock:
Thanks to @byersiiasa we've found another bug in `Codelist.to_yaml`.
Reading a yaml file into a `CodeList` object and dumping it out to yaml again does not work for nested structures.
As an example taking a variable code list with a _region-aggregation_ attributes

```yaml
- Price|Carbon:
    definition: Price of carbon (as reported by the model)
    unit: US$2010/t CO2
    region-aggregation:
        - Price|Carbon (mean):
            method: mean
```

reading it in and writing it out to yaml again:

```python
VariableCodeList.from_directory("variable", "definitions").to_yaml()
```

gives the following:

```yaml
Price|Carbon:
    description: Price of carbon (as reported by the model)
    unit: US$2010/t CO2
    region_aggregation:
    - Price|Carbon (mean):
        method: mean
```

The indentation under region_aggregation is missing which breaks reading the yaml file.

It seems that this is a known issue with pyyaml (https://github.com/yaml/pyyaml/issues/234 and https://stackoverflow.com/questions/25108581/python-yaml-dump-bad-indentation) so it shouldn't be too hard to fix.
For now I've added a breaking unit test that tests the above described behavior.

phackstock:
Fixed another two small issues #222 and #223.

byersiiasa:
Can confirm that for the workflow I tried for engage_internal - this worked ;)

PR #221, opened 2023-01-13
Add simple tutorial to demonstrate -- read codelist, export to excel,…

byersiiasa:
Add simple tutorial to demonstrate
 -- read codelist, export to excel,…edit, 
-- then read and write back to yaml.

Suggested by @phackstock  that this could be as ipynb and render on the docs... if not we can change to .rst also.

Also suggested to put comment in the variables docs... (will do)

### DIFF
--- a/doc/source/user_guide/tutorial_edit_existing_variable_codelist.ipynb
+++ b/doc/source/user_guide/tutorial_edit_existing_variable_codelist.ipynb
@@ -1,0 +1,210 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "15ad0472",
+   "metadata": {},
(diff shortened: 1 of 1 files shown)


byersiiasa:
Ok - edited the variable.rst page too.
Good for a first review I hope

byersiiasa:
@mwerning I haven't checked if this still works, but if it does, is there still interest @phackstock @dc-almeida in merging?

byersiiasa:
(needs addition to documentation I think)

danielhuppmann:
Thanks @byersiiasa - this tutorial should still work, but I'm very skeptical whether we want to highlight this approach as part of the documentation...

PR #226, opened 2023-03-01
Hierarchy Filter (first pull request)

GretchenSchowalter:
I've added the hierarchy filter to the codelist.py file, and two different tests to the test_codelist.py. 

Short explanation: The filter requires two arguments: self and the particular hierarchy you want to filter for. The first step in the function is the creation of an empty list called countries. Then iterating through the values of the self.mapping.values(), we then can compare if the hierarchy for each region matches the hierarchy argument. If so, then the name of that country is appended to the countries[] list. Once the computer has iterated through self.mapping.values(), it then checks if the countries[] list is empty. If not, then filter has worked and it returns the list of names. If the countries[] list is still empty, then either the hierarchy argument inputted is not used for the particular model, or the hierarchy argument has a typo.

I have just a couple questions: 
1) Right now the hierarchy_filter(hierarchy: str) returns a List[str] with just the names of each of the regions that compose that particular hierarchy. Would it be better if the filter instead returned List[Code]? 

2) Would it be helpful to add another level of filtration? For example, print(hierarchy_filter(R5)) prints OECD & EU, Asia, Reforming economies, Latin America, Middle East, and Other. Then one step further, print(hierarchy_filter(OECD & EU)) prints Australia, Austria, Belgium,..., United States.


danielhuppmann:
Thanks for the initial PR, @GretchenSchowalter! To directly answer your first question, I think that ideally, the filter-option should return a `RegionCodeList` instance (downselected to the relevant items). This way, a user can easily "chain" multiple operations.

GretchenSchowalter:
Okay! Thanks for the feedback! I've updated the code, so I will make another pull request soon.

danielhuppmann:
No need to close the PR and start a new one, @GretchenSchowalter - simply add new commits to this branch…

nomenclature/codelist.py
```diff
@@ -542,3 +543,28 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
...
+        -------
+        RegionCodeList
+            Returns a list of the component regions of the inputted hierarchy.
+        """
+
+        mapping = {k: v for (k, v) in self.mapping.items() if v.hierarchy == hierarchy}
+        if mapping:
+            return RegionCodeList(name=f"{hierarchy}", mapping=mapping)
+        else:
+            raise ValueError(f"No hierarchy found for {hierarchy}.")
```
danielhuppmann, line 570:
```suggestion
            raise ValueError(f"Hierarchy level does not exist: {hierarchy}")
```

nomenclature/codelist.py
```diff
@@ -542,3 +543,28 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
             mapping[code.name] = code
 
         return cls(name=name, mapping=mapping)
+
+    def hierarchy_filter(self, hierarchy: str) -> "RegionCodeList":
```
danielhuppmann, line 547: Suggest to rename that for simplicity (and adding other filter options later).
```suggestion
    def filter(self, hierarchy: str) -> "RegionCodeList":
```

danielhuppmann (commented):
Two additional minor comments...

nomenclature/codelist.py
```diff
@@ -542,3 +543,28 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
...
+
+        Returns
+        -------
+        RegionCodeList
+            Returns a list of the component regions of the inputted hierarchy.
+        """
+
+        mapping = {k: v for (k, v) in self.mapping.items() if v.hierarchy == hierarchy}
+        if mapping:
+            return RegionCodeList(name=f"{hierarchy}", mapping=mapping)
```
phackstock, line 568: `hierachy` is already a string and you're not concatenating it with anything else so no need for an f-string here:
```suggestion
            return RegionCodeList(name=hierarchy, mapping=mapping)
```

On a related note I'm not sure if we should "rename" the resulting RegionCodeList. This `RegionCodeList(name=self.name, mapping=mapping)` might be fine as well but that's more of a design question.

nomenclature/codelist.py
```diff
@@ -542,3 +543,28 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
...
+        ValueError
+            If User inputs a typo or a hierarchy not compatible with the given model.
+
+        Returns
+        -------
+        RegionCodeList
+            Returns a list of the component regions of the inputted hierarchy.
+        """
+
+        mapping = {k: v for (k, v) in self.mapping.items() if v.hierarchy == hierarchy}
```
phackstock, line 566: Nice work on using a dict comprehension here :+1:. 
Minor technicality, you don't need the parentheses around the `k, v`:
```suggestion
        mapping = {k: v for k, v in self.mapping.items() if v.hierarchy == hierarchy}
```

phackstock (changes requested):
Looks quite good to me, nice work @GretchenSchowalter.
Just two minor comments below.

nomenclature/codelist.py
```diff
@@ -542,3 +543,28 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
...
+        if mapping:
+            return RegionCodeList(name=f"{hierarchy}", mapping=mapping)
+        else:
+            raise ValueError(f"Hierarchy level does not exist: {hierarchy}")
```
phackstock, line 570: Maybe you could also provide the available hierarchy levels here.

tests/test_codelist.py
```diff
@@ -1,203 +1,231 @@
...
+    assert obs == exp
+    
+    
+def test_RegionCodeList_hierarchy_filter_ValueError():
+    """Test that verifies the filter gives error when user inputs an unrecognizeable
+    hierarchy"""
+    
+    # read RegionCodeList
+    rcl = RegionCodeList.from_directory("Region", TEST_DATA_DIR / "region_codelist")
+    match = "Hierarchy level does not exist: R77."
```
phackstock, line 229: I think this test is currently failing because `match` is a regular expression where "." means _match anything_.
If you leave out the dot it should be all good. If you actually want to match for the . character and not for anything you'd need to escape it. If I recall correctly that should work like this: "\." or "\\.".

phackstock (approved):
Looks good to me, thanks @GretchenSchowalter. 
The one open question is if we should add this new feature formally to the docs or if this is more of an internal one anyway. @danielhuppmann, what are your feelings on that?

nomenclature/codelist.py
```diff
@@ -542,3 +541,47 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
...
+        Returns
+        -------
+        RegionCodeList
+            Returns a list of the component regions of the inputted hierarchy.
+        """
+
+        mapping = {k: v for k, v in self.mapping.items() if v.hierarchy == hierarchy}
+        if mapping:
+            return RegionCodeList(name=self.name, mapping=mapping)
+        else:
```
danielhuppmann, line 565: This is a nice error message and elegant code, but it is a lot of code and will be hard to maintain going forward. Better to simply raise: 
```python
f"Filtered RegionCodeList is empty: hierarchy={hierarchy}\nUse `RegionCodeList.hierarchy` for available items."
```

And then add a placeholder
```python
def hierarchy(self):
    raise NotImplementedError("This method is not yet implemented.")
```

And you can add this method as a future PR.

danielhuppmann (changes requested):
Thanks for moving forward, @GretchenSchowalter - a few more suggestions inline.

nomenclature/codelist.py
```diff
@@ -542,3 +541,41 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
             mapping[code.name] = code
 
         return cls(name=name, mapping=mapping)
+
```
phackstock, line 544: `hierarchy` would need to be a property if it should be called like this `RegionCodeList.hierarchy` as mentioned in the error message in line 579:

```suggestion
    @property
```

As it currently stands you'd have to use `RegionCodeList.hierarchy()`.

danielhuppmann: I think that a property is more appropriate given the rest of the API...

danielhuppmann: e.g., you can do 
```python
RegionCodeList["Some Region"].hierarchy
```
gives the hierarchy for a particular region
```python
RegionCodeList.hierarchy
```
gives the list of all hierarchy levels

danielhuppmann (approved):
Looks good to me! Thanks @GretchenSchowalter, and congrats for your first successful PR!

PR #229, opened 2023-03-08
Remove unnecessary test module

phackstock:
Found a file that should not have been in the package. 
Removing it.


danielhuppmann (approved):
Thanks!

PR #232, opened 2023-03-16
Adding to hierarchy method in codelist.py

GretchenSchowalter:
I added short bit of code that updates the hierarchy method in codelist.py which currently returns a NotImplementedError.

I've created a sorted list from a set of all the possible hierarchies for one .yaml file. This code simply returns this list of strings to the user.


nomenclature/codelist.py
```diff
@@ -543,8 +543,21 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
         return cls(name=name, mapping=mapping)
 
     @property
-    def hierarchy(self):
-        raise NotImplementedError("This method is not yet implemented.")
+    def hierarchy(self) -> List[str]:
+        """Return List of strings of available hierarchies to filter
```
danielhuppmann, line 547: It's obvious that the returned object is a list of strings from the signature, no need to repeat. Also, the method is not only relevant for filtering, so no to write that
```suggestion
        """Return the hierarchies defined in the RegionCodeList
```

nomenclature/codelist.py
```diff
@@ -543,8 +543,21 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
...
+        List[str]
+
+        """
+
+        # creates a sorted list of strings from a set with
+        # all available hierarchy options
+        avail_ops: List[str] = sorted(
+            list({v.hierarchy for v in self.mapping.values()})
+        )
+        return avail_ops
```
danielhuppmann, line 560: Make it one line... Also, list comprehension is easier to read, I think...
```suggestion
        return sorted([v.hierarchy for v in self.mapping.values()])
```

danielhuppmann (commented):
Thanks! But please check one thing - at the moment, the `region_codelist` used for the test has exactly one region per hierarchy. I believe that your current implementation would have duplicates if there are multiple regions in a hierarchy. Please extend the test-region-codelist (e.g., by adding an "Another Country"), and extend the code to remove duplicates from the returned list.

tests/test_codelist.py
```diff
@@ -203,29 +203,38 @@ def test_to_yaml_from_directory(tmp_path):
     )
 
 
-def test_RegionCodeList_hierarchy_filter():
+def test_RegionCodeList_filter():
```
phackstock, line 206: Do you need the extra attributes here if you're just testing for the hierarchy filtering?
I don't think you're actually using them to test anything so I would cut them out here and of the region.yaml file.
`hierarchy` and `file` you can just add when you're creating the Code objects making for a shorter test.

danielhuppmann: Disagree, the "file" attribute is added automatically at initialization of the **RegionCodeList** instance from a file - so you have to add it to ensure equality between expected and observed objects.

phackstock (approved):
Looks all good to me thanks @GretchenSchowalter :+1:.

Issue #233, opened 2023-03-27
Filter CodeLists by any attribute


danielhuppmann:
Building on #226, we should now extend the `filter()` method and make it usable for any **CodeList** instance (not just **RegionCodeList**).

The use case that I have in mind is filtering a list of variables to have only those that are *required*, e.g. that have an attribute like
```yaml
- Final Energy|Electricity:
    description: Final energy consumption of electricity (including on-site solar
      PV), excluding transmission/distribution losses
    unit: EJ/yr
    required: true
```

See a real-world use case [here](https://github.com/iiasa/engage-gains-mapping/blob/ccb1cfd237a021ac0df1f37dae67a8fef5a4df42/definitions/variable/variable.yaml#L39).

phackstock:
I guess the interface for this would look something like this then:

```python

class CodeList:

    def filter(self, attribute, value) -> CodeList:
        matched = [code.attribute == value for code in self]
        # return a new CodeList(matched)
```

The above example wouldn't work exactly like this but is this what you had in mind @danielhuppmann?

danielhuppmann:
Well, I'd rather follow the pyam `filter()` signature

```python
class CodeList:

    def filter(self, **kwargs) -> CodeList:
        keep = np.ones(len(self), dtype=bool)
        for attribute, value in kwargs:
            keep = np.logical_and(keep, [code.attribute == value for code in self])
            # return a new CodeList(keep)
```

This would also be compatible with the existing signature (if a user wrote `filter(hierarchy="common")`, not `filter("common")`).

phackstock:
Ok, good point, that is cleaner and more powerful.

phackstock:
Closing as this was addressed by #236

PR #236, opened 2023-04-12
General Filter: Addresses the "Filter CodeLists by any attribute" issue #233

GretchenSchowalter:
This is the more generalized filter which will filter by any attribute. It uses the "keep" array to keep track of whether the elements satisfy the provided filter attributes & values. It raises an AttributeError when a provided attribute is not in any code element of the CodeList. There is a logger.warning() if a user "over-filters" by providing so many attributes that no code elements satisfy all the filter criteria.

Right now, the filter appends (with None (what I set as the default value), and then eventually to not 'value') to the extra_attributes: {} of a code element if the provided attribute doesn't exist for that code element. Right now the code does not go back and delete this "added" extra_attribute, but should I go back and do that to sort of clean up after the filter?


GretchenSchowalter:
Hey, Daniel! Quick update. Philip and I chatted and came up with several changes to make this code shorter and more readable (and just better!). I will be working on making those changes now (he also addressed my question above... the filter should not alter the actual codes getting filtered, which makes total sense), so stay tuned for another commit coming in the next couple of days.

GretchenSchowalter:
Hello Daniell and Philip! Another update: I just pushed another commit. Hopefully, the code is simpler now, although I realize that I still have a problem using the setattr() and getattr(). I tried only using the getattr(code, attribute, None) like this to set the default value of an attribute to None, but this has two problems: It sets that attribute to None even if that code does have that attribute because it hasn't been set yet, and then it still modifies code(s) that don't have that attribute by adding it and setting it to None. So I opted to use the setattr() as well, but now I still have the issue of modifying the original code(s) to get the filter to work. Fixing this (by deleting these set attributes) is my next step! If there is another suggestion as to how to filter without using the setatt() or getattr() and modifying the codes (and then going back to delete the "added on" parts), let me know! I also realize that this is an issue if any of the extra_attributes are set to be None before the filtering even begins. So if there are any suggestions on this as well, I am all ears!

GretchenSchowalter:
Another quick update: To solve this issue, I just made a copy of self, and then deleted this copy when the filtering is complete. Everything still runs smoothly. Was this an alright workaround for this? Or do you think there is a more storage-efficient way to solve this problem of altering the "self" when filtering?

nomenclature/codelist.py
```diff
@@ -388,6 +389,65 @@ def codelist_repr(self, json_serialized=False) -> Dict:
...
+        -------
+        :class:'CodeList'
+
+        """
+
+        keep: List(Code) = []
+        for attribute in kwargs:
+            # Set the syntax as: code.<attribute>
+            for code in self.mapping.values():
+                if attribute in code.extra_attributes:
```
phackstock, line 416: I think you should be able to leave you this and the following three lines:
```suggestion
```
which means you should be good to leave out the entire `for code in self.mapping.values()`` loop.

danielhuppmann:
I don't understand the setattr-strategy.

I would have expected an approach like
```python
def _match_attribute(code, key, value):
    if attribute in code.extra_attributes:
        if code[attribute] == value
            return True
    return False

keep = np.ones(len(self), dtype=bool)
for key, value in kwargs:
    keep = np.logical_and(keep, [_match_attrribute(code, key, value) in self.mapping.values])

return CodeList(
     name=self.name,
     mapping=dict([code.name, code for code, _keep in zip(self.mapping.values, keep) if _keep]),
 )
```

phackstock (changes requested):
Looks like a nice step in the right direction.
Couple of comments that should solve your attribute setting problem and clean up the code a bit.
After that we should be good to go.

GretchenSchowalter:
Hello, Daniel and Philip! I just pushed the updated general filter. Right now it follows Daniel's strategy. I also came up with another (possibly shorter) way to do it (see below). Which one would you prefer? 
```
    def filter(self, **kwargs) -> "CodeList":        
	filtered_codelist = CodeList(
            name=self.name,
            # Use a dictionary comprehension to filter the Code
            # objects based on the attribute-value pairs in kwargs
            mapping={
                code.name: code
                for code in self.mapping.values()
                if all(
                    hasattr(code, attribute) and getattr(code, attribute, None) == value
                    for attribute, value in kwargs.items()
                )
            },
        )

        if filtered_codelist.mapping:
            # If the resulting CodeList is not empty, return it
            return filtered_codelist

        # If the resulting CodeList is empty,
        # log a warning and return the original CodeList
        logging.warning("Formatted data is empty!")
        return filtered_codelist
```

danielhuppmann:
When two options work, shorter is preferable. And if you combine the implemented approach (with the internal `_match_attribute()` function) with your alternative proposal, you could have an even more concise and also more readable result.

Also, if the method is as short as yours, it's not necessary to add a docstring for every line.

phackstock (changes requested):
Two small changes other than that good to go from my side.
Just for illustrative purposes, an alternative way to achieve the same filtering. 

```python
filtered_codelist = CodeList(
            name=self.name,
            mapping={
                code.name: code
                for code in self.mapping.values()
                if all(
                    hasattr(code, attribute) and getattr(code, attribute) == value
                    for attribute, value in kwargs.items()
                )
            },
        )

        if not filtered_codelist.mapping:
            logging.warning("Formatted data is empty!")
        return filtered_codelist

```

phackstock (approved):
Looks good to me. Thanks @GretchenSchowalter.

tests/data/general_filtering/basic_codelist.yaml
```diff
@@ -0,0 +1,10 @@
+- Big Variable:
+    definition: Some basic variable
+    required: True
+- Another Variable:
+    definition: some details
+    required_A: True
+    required_B: This is true
+- Second Big Variable:
+    definition: a very basic variable
+    required: False
```
danielhuppmann, line 10: Please add an empty line at the end of the file

tests/data/general_filtering/basic_codelist.yaml
```diff
@@ -0,0 +1,10 @@
+- Big Variable:
+    definition: Some basic variable
+    required: True
+- Another Variable:
+    definition: some details
+    required_A: True
```
danielhuppmann, line 6: Please use other attribute names, `required_A` and `required_B` doesn't really fit, maybe `some_attribute`...

nomenclature/codelist.py
```diff
@@ -388,6 +389,49 @@ def codelist_repr(self, json_serialized=False) -> Dict:
...
+        ----------
+        **kwargs
+            a variable number of keyword arguments representing
+            attribute-value pairs.
+
+        Returns
+        -------
+        Instance of 'CodeList'
+            A new CodeList object with only the Code objects that
+            match all the attribute-value pairs.
```
danielhuppmann, line 405: This can be made shorter...
```suggestion
        CodeList
```

nomenclature/codelist.py
```diff
@@ -388,6 +389,49 @@ def codelist_repr(self, json_serialized=False) -> Dict:
...
+
+        Raises
+        ------
+        logging.WARNING
+            A warning if the new CodeList is empty.
+
+
```
danielhuppmann, line 412: Too many empty lines, and I think it's strange that a warning is mentioned in the "Raises" (error) section, so I'd remove that section.
```suggestion
```

nomenclature/codelist.py
```diff
@@ -388,6 +389,49 @@ def codelist_repr(self, json_serialized=False) -> Dict:
...
         return nice_dict
 
+    def filter(self, **kwargs) -> "CodeList":
+        """Filter a CodeList by any attribute-value pairs.
+
+        Parameters
+        ----------
+        **kwargs
+            a variable number of keyword arguments representing
+            attribute-value pairs.
```
danielhuppmann, line 399: The term "variable" can be very confusing here with the class **VariableCodeList**.
```suggestion
            Attribute-value mappings to be used for filtering.
```

danielhuppmann (commented):
Almost there, a few final clean-up suggestions.

danielhuppmann:
Thanks @GretchenSchowalter!

PR #237, opened 2023-04-13
Export aggregation difference to excel

phackstock:
Started by a feature request from @orichters (here: https://github.com/iiasa/ngfs-phase-4-internal-workflow/discussions/6) I have implemented a first version of exporting the differences found in region aggregation to excel.
To be precise, this tackles the case where a model natively reports a region, e.g. _World_ that is also re-calculated as part of our region aggregation. So far only the top rows of the difference data frame are printed as part of the log file (e.g. https://data.ece.iiasa.ac.at/ngfs-phase-4-internal/#/uploads/jobs/21/log):

```console
2023-04-11 16:55:14 WARNING  Difference between original and aggregated data:
                                                                                                              original  aggregated
model                 scenario  region variable                                        unit            year                       
REMIND-MAgPIE 3.1-4.6 d_delfrag World  Agricultural Demand|Crops|Energy                million t DM/yr 2005  25.324800   25.325200
                                       Agricultural Demand|Crops|Energy|2nd generation million t DM/yr 2005   0.666700    0.667200
                                       Agricultural Demand|Livestock|Other             million t DM/yr 2015  24.059500   24.059800
                                       Agricultural Production|Energy|Crops            million t DM/yr 2005   0.680300    0.680400
                                       Capacity Additions|Electricity|Biomass          GW/yr           2060   0.000038    0.000038
...                                                                                                                ...         ...
                      o_lowdem  World  Yield|Sugarcrops                                t DM/ha/yr      2060  42.070800   46.233420
                                                                                                       2070  43.330200   48.201466
                                                                                                       2080  44.277000   49.474088
                                                                                                       2090  45.033500   49.908712
                                                                                                       2100  43.943600   50.042956
```

I have added an `export_difference` option to `RegionProcessor.apply()` that writes the data to a file called _difference.xlsx_.

So far, the workflow I envision, should work as follows:

1. A user uploads data to a Scenario Explorer and the processing finds differences in aggregated and model native data.
2. There is a warning issued with the top few rows of the differences. 
3. After that warning there is another message:
```console
If you want to get the differences, run the following command on your local machine:
`nomenclature run-region-processing your_data.xlsx --export-differences`
from the processing workflow directory. 
The differences will be exported to `difference.xlsx`.
```
4. The user then clones the corresponding workflow directory locally and installs the requirements.
5. Using `nomenclature run-region-processing your_data.xlsx --export-differences`, they then get the differences into a file called _difference.xlsx_ for more close inspection.

@orichters and @danielhuppmann, does this sound like a good workflow to you?

## Update

For the data frame of the differences between original and aggregated data I now added a _relative (absolute) difference_ column, calculated as `100*abs((original-aggregated)/original)`. 
I also sort the data frame by the newly created column to show the user the biggest differences first.
Finally, I changed the relative tolerance from 1e-5 to 1% to reduce the rate of false positives.

### DIFF
--- a/doc/source/user_guide/model-mapping.rst
+++ b/doc/source/user_guide/model-mapping.rst
@@ -120,9 +120,39 @@ are compared and combined according to the following logic:
    *region_b*. The **provided results** take **precedence** over the aggregated ones.
    Additionally, the aggregation is computed and compared to the provided results. If
    there are discrepancies, a warning is written to the logs.
-   
+
    .. note::
 
       Please note that in case of differences no error is raised. Therefore it is
       necessary to check the logs to find out if there were any differences. This is
       intentional since some differences might be expected.
+
+
+How to get aggregation differences locally
+------------------------------------------
+
+In order to get the full differences perform the following steps:
+
+1. Make sure you have `pyam-iamc >= 1.7.0`, `nomenclature-iamc>=0.9.2` and
+`pandas >= 1.5.2` installed.
+2. Clone the workflow directory of your project
+3. Navigate to the workflow directory
+4. Using a jupyter notebook or python script run the following:
+
+.. code:: python
+
+  from pyam import IamDataFrame
+  from nomenclature import DataStructureDefinition, RegionProcessor
+
+  data = IamDataFrame("/path/to/your/input/data.xlsx")
+
+  dsd = DataStructureDefinition("definitions")
+  processor = RegionProcessor.from_directory("mappings", dsd)
+
+  # get the differences as a pandas dataframe
+  # the value for the relative tolerances can be adjusted, defaults to 0.01
+  differences = processor.check_region_aggregation(data, rtol_difference=0.01)
+  differences.to_excel("differences.xlsx")
+
+For details on the check_region_aggregation function please refer to
+:func:`RegionProcessor.check_region_aggregation`.
--- a/doc/source/user_guide/model-mapping.rst
+++ b/doc/source/user_guide/model-mapping.rst
@@ -1,5 +1,7 @@
 .. _model_mapping:
 
+.. currentmodule:: nomenclature
+
 Region processing using model mappings
 ======================================
 
@@ -134,8 +136,7 @@ Computing differences between original and aggregated data
 In order to get the differences between the original data (e.g., results reported by the model)
 and the data aggregated according to the region mapping, perform the following steps:
 
-1. Make sure you have `pyam-iamc >= 1.7.0`, `nomenclature-iamc>=0.10.0` and
-`pandas >= 1.5.2` installed.
+1. Make sure you have ``pyam-iamc >= 1.7.0`` and ``nomenclature-iamc>=0.10.0`` installed.
 2. Clone the workflow directory of your project
 3. Navigate to the workflow directory
 4. Using a jupyter notebook or python script run the following:
@@ -153,7 +154,19 @@ and the data aggregated according to the region mapping, perform the following s
   # get the differences as a pandas dataframe
   # the value for the relative tolerances can be adjusted, defaults to 0.01
   processed_data, differences = processor.check_region_aggregation(data, rtol_difference=0.01)
+  # save the result of the region processing
+  processed_data.to_excel("results.xlsx")
+  # and the differences
   differences.to_excel("differences.xlsx")
 
 For details on this feature, please refer to
-:func:`RegionProcessor.check_region_aggregation`.
+:py:meth:`RegionProcessor.check_region_aggregation`.
+
+Alternatively you can also use the nomenclature cli:
+
+.. code-block:: bash
+
+  $ nomenclature check-region-aggregation /path/to/your/input/data.xlsx
+  -w workflow_directory --processed_data results.xlsx --differences differences.xlsx
+
+For cli details please refer to :ref:`cli`.
--- a/doc/source/user_guide/model-mapping.rst
+++ b/doc/source/user_guide/model-mapping.rst
@@ -128,10 +128,11 @@ are compared and combined according to the following logic:
       intentional since some differences might be expected.
 
 
-How to get aggregation differences locally
-------------------------------------------
+Computing differences between original and aggregated data
+----------------------------------------------------------
 
-In order to get the full differences perform the following steps:
+In order to get the differences between the original data (e.g., results reported by the model)
+and the data aggregated according to the region mapping, perform the following steps:
 
 1. Make sure you have `pyam-iamc >= 1.7.0`, `nomenclature-iamc>=0.10.0` and
 `pandas >= 1.5.2` installed.
@@ -151,8 +152,8 @@ In order to get the full differences perform the following steps:
 
   # get the differences as a pandas dataframe
   # the value for the relative tolerances can be adjusted, defaults to 0.01
-  differences = processor.check_region_aggregation(data, rtol_difference=0.01)
+  processed_data, differences = processor.check_region_aggregation(data, rtol_difference=0.01)
   differences.to_excel("differences.xlsx")
 
-For details on this feature please refer to
+For details on this feature, please refer to
 :func:`RegionProcessor.check_region_aggregation`.
--- a/doc/source/user_guide/model-mapping.rst
+++ b/doc/source/user_guide/model-mapping.rst
@@ -133,7 +133,7 @@ How to get aggregation differences locally
 
 In order to get the full differences perform the following steps:
 
-1. Make sure you have `pyam-iamc >= 1.7.0`, `nomenclature-iamc>=0.9.2` and
+1. Make sure you have `pyam-iamc >= 1.7.0`, `nomenclature-iamc>=0.10.0` and
 `pandas >= 1.5.2` installed.
 2. Clone the workflow directory of your project
 3. Navigate to the workflow directory
@@ -154,5 +154,5 @@ In order to get the full differences perform the following steps:
   differences = processor.check_region_aggregation(data, rtol_difference=0.01)
   differences.to_excel("differences.xlsx")
 
-For details on the check_region_aggregation function please refer to
+For details on this feature please refer to
 :func:`RegionProcessor.check_region_aggregation`.
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -98,7 +98,7 @@ def cli_valid_project(
     assert_valid_structure(path, definitions, mappings, required_data, dimensions)
 
 
-@cli.command("run-region-processing")
+@cli.command("check-region-aggregation")
 @click.argument("input_data_file", type=click.Path(exists=True, path_type=Path))
 @click.option(
     "-w",
@@ -112,7 +112,7 @@ def cli_valid_project(
     "--processed-data", type=click.Path(path_type=Path), default="results.xlsx"
 )
 @click.option("--differences", type=click.Path(path_type=Path), default=None)
-def run_region_processing(
+def check_region_aggregation(
     input_data_file: Path,
     workflow_directory: Path,
     definitions: str,
@@ -148,7 +148,7 @@ def run_region_processing(
     results of the aggregation will be exported to results.xlsx and the differences to
     differences.xlsx.
 
-    $ nomenclature run-region-processing input_data.xlsx -w workflow_directory
+    $ nomenclature check-region-processing input_data.xlsx -w workflow_directory
                         --processed_data results.xlsx --differences differences.xlsx
 
     """
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -153,14 +153,11 @@ def run_region_processing(
                         --processed_data results.xlsx --differences differences.xlsx
 
     """
-    region_processor = RegionProcessor.from_directory(
+    results_df, differences_df = RegionProcessor.from_directory(
         workflow_directory / mappings,
         DataStructureDefinition(workflow_directory / definitions),
-    )
-    data = IamDataFrame(input_data_file)
+    ).check_region_aggregation(IamDataFrame(input_data_file))
     if processed_data:
-        region_processor.apply(data).to_excel(processed_data)
+        results_df.to_excel(processed_data)
     if differences:
-        region_processor.check_region_aggregation(data).reset_index().to_excel(
-            differences, index=False
-        )
+        differences_df.reset_index().to_excel(differences, index=False)
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -114,7 +114,6 @@ def run_region_processing(
     output: Optional[bool],
     export_differences: Optional[bool],
 ):
-
     result, difference = process(
         IamDataFrame(input_data_file),
         dsd := DataStructureDefinition(path / definitions),
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -123,4 +123,4 @@ def run_region_processing(
     if output:
         result.to_excel(output)
     if export_differences:
-        difference.to_excel(export_differences)
+        difference.reset_index().to_excel(export_differences, index=False)
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -103,8 +103,8 @@ def cli_valid_project(
 @click.option("-p", "path", type=click.Path(exists=True, path_type=Path), default=".")
 @click.option("-d", "--definitions", type=str, default="definitions")
 @click.option("-m", "--mappings", type=str, default="mappings")
-@click.option("--output", type=click.Path(path_type=Path), default="results.xlsx")
-@click.option("--export-differences", type=click.Path(path_type=Path), default=None)
+@click.option("--processed-data", type=click.Path(path_type=Path), default="results.xlsx")
+@click.option("--differences", type=click.Path(path_type=Path), default=None)
 def run_region_processing(
     input_data_file: Path,
     path: Path,
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -5,6 +5,7 @@ from typing import List, Optional
 import click
 
 from pyam import IamDataFrame
+from nomenclature.core import process
 from nomenclature.definition import DataStructureDefinition
 from nomenclature.processor import RegionProcessor
 from nomenclature.testing import assert_valid_structure, assert_valid_yaml
@@ -100,27 +101,27 @@ def cli_valid_project(
 
 @cli.command("run-region-processing")
 @click.argument("input_data_file", type=click.Path(exists=True, path_type=Path))
-@click.option(
-    "-d",
-    "--definitions",
-    type=click.Path(exists=True, path_type=Path),
-    default="definitions",
-)
-@click.option(
-    "-m", "--mappings", type=click.Path(exists=True, path_type=Path), default="mappings"
-)
-@click.option("--export-processing-result", is_flag=True, default=True)
-@click.option("--export-differences", is_flag=True, default=False)
+@click.option("-p", "path", type=click.Path(exists=True, path_type=Path), default=".")
+@click.option("-d", "--definitions", type=str, default="definitions")
+@click.option("-m", "--mappings", type=str, default="mappings")
+@click.option("--output", type=click.Path(path_type=Path), default="results.xlsx")
+@click.option("--export-differences", type=click.Path(path_type=Path), default=None)
 def run_region_processing(
     input_data_file: Path,
-    definitions: Path,
-    mappings: Path,
-    export_processing_result: bool,
-    export_differences: bool,
+    path: Path,
+    definitions: str,
+    mappings: str,
+    output: Optional[bool],
+    export_differences: Optional[bool],
 ):
 
-    result = RegionProcessor.from_directory(
-        mappings, DataStructureDefinition(definitions)
-    ).apply(IamDataFrame(input_data_file), export_difference=export_differences)
-    if export_processing_result:
-        result.to_excel("result.xlsx")
+    result, difference = process(
+        IamDataFrame(input_data_file),
+        dsd := DataStructureDefinition(path / definitions),
+        ["variable"],
+        RegionProcessor.from_directory(path / mappings, dsd),
+    )
+    if output:
+        result.to_excel(output)
+    if export_differences:
+        difference.to_excel(export_differences)
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -120,8 +120,7 @@ def run_region_processing(
     processed_data: Optional[Path],
     differences: Optional[Path],
 ):
-    """Perform region processing an optionally return differences between aggregated
-    and model native data
+    """Perform region processing and compare aggregated and original data
 
     Parameters
     ----------
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -4,6 +4,9 @@ from typing import List, Optional
 
 import click
 
+from pyam import IamDataFrame
+from nomenclature.definition import DataStructureDefinition
+from nomenclature.processor import RegionProcessor
 from nomenclature.testing import assert_valid_structure, assert_valid_yaml
 
 cli = click.Group()
@@ -93,3 +96,31 @@ def cli_valid_project(
     """
     assert_valid_yaml(path)
     assert_valid_structure(path, definitions, mappings, required_data, dimensions)
+
+
+@cli.command("run-region-processing")
+@click.argument("input_data_file", type=click.Path(exists=True, path_type=Path))
+@click.option(
+    "-d",
+    "--definitions",
+    type=click.Path(exists=True, path_type=Path),
+    default="definitions",
+)
+@click.option(
+    "-m", "--mappings", type=click.Path(exists=True, path_type=Path), default="mappings"
+)
+@click.option("--export-processing-result", is_flag=True, default=True)
+@click.option("--export-differences", is_flag=True, default=False)
+def run_region_processing(
+    input_data_file: Path,
+    definitions: Path,
+    mappings: Path,
+    export_processing_result: bool,
+    export_differences: bool,
+):
+
+    result = RegionProcessor.from_directory(
+        mappings, DataStructureDefinition(definitions)
+    ).apply(IamDataFrame(input_data_file), export_difference=export_differences)
+    if export_processing_result:
+        result.to_excel("result.xlsx")
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -5,7 +5,6 @@ from typing import List, Optional
 import click
 
 from pyam import IamDataFrame
-from nomenclature.core import process
 from nomenclature.definition import DataStructureDefinition
 from nomenclature.processor import RegionProcessor
 from nomenclature.testing import assert_valid_structure, assert_valid_yaml
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -111,15 +111,17 @@ def run_region_processing(
     path: Path,
     definitions: str,
     mappings: str,
-    output: Optional[bool],
-    export_differences: Optional[bool],
+    output: Optional[Path],
+    export_differences: Optional[Path],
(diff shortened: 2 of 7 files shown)


orichters:
Workflow seems fine from my side. "clones the corresponding workflow directory locally" refers in my case to `ngfs-phase-4-internal-workflow`? Where is the explanation how I can "installs the requirements"? The rest sounds feasible :)

phackstock:
> "clones the corresponding workflow directory locally" refers in my case to ngfs-phase-4-internal-workflow?
Exactly, as this is just a first quick draft. I'll work on making the instructions better.
> Where is the explanation how I can "installs the requirements"?
You would run `python3 -m pip install -r requirements.txt` to ensure that you can run the workflow locally and have the corresponding nomenclature command line option `run-region-processing` installed.

nomenclature/cli.py
```diff
@@ -93,3 +96,31 @@ def cli_valid_project(
     """
     assert_valid_yaml(path)
     assert_valid_structure(path, definitions, mappings, required_data, dimensions)
+
+
+@cli.command("run-region-processing")
+@click.argument("input_data_file", type=click.Path(exists=True, path_type=Path))
```
danielhuppmann, line 102: Suggest to use shorter arg names, similar to pyam...
```suggestion
@click.argument("data", type=click.Path(exists=True, path_type=Path))
```

nomenclature/processor/region.py
```diff
@@ -563,11 +573,31 @@ def _compare_and_merge(original: pd.Series, aggregated: pd.Series) -> IamDataFra
 
     # drop rows that are not in conflict
     compare = compare.dropna()
-    compare = compare[~np.isclose(compare["original"], compare["aggregated"])]
-
+    compare = compare[
+        ~np.isclose(compare["original"], compare["aggregated"], rtol=0.01)
```
danielhuppmann, line 577: Please make `rtol` a keyword-argument.

nomenclature/processor/region.py
```diff
@@ -563,11 +573,31 @@ def _compare_and_merge(original: pd.Series, aggregated: pd.Series) -> IamDataFra
...
+    ]
+    compare["relative difference (%)"] = 100 * np.abs(
```
danielhuppmann, line 579: Please don't do number-formatting as hard-code - if at all, do that when printing the dataframe to the log.

nomenclature/cli.py
```diff
@@ -93,3 +96,31 @@ def cli_valid_project(
...
+@click.option(
```
danielhuppmann, line 103: Suggest to add `path` as optional keyword argument for consistency with `validate-project()`.

nomenclature/cli.py
```diff
@@ -93,3 +96,31 @@ def cli_valid_project(
...
+    "-d",
+    "--definitions",
+    type=click.Path(exists=True, path_type=Path),
+    default="definitions",
+)
+@click.option(
+    "-m", "--mappings", type=click.Path(exists=True, path_type=Path), default="mappings"
+)
+@click.option("--export-processing-result", is_flag=True, default=True)
```
danielhuppmann, line 112: Suggest to simplify to `output`, default to "results.xlsx", not written to file if `None`.

nomenclature/cli.py
```diff
@@ -93,3 +96,31 @@ def cli_valid_project(
...
+@click.option("--export-differences", is_flag=True, default=False)
```
danielhuppmann, line 113: Maybe make that a string or `None`, where the string is the name of the output file (default `None`, no file is exported.

nomenclature/cli.py
```diff
@@ -93,3 +96,31 @@ def cli_valid_project(
...
+def run_region_processing(
+    input_data_file: Path,
+    definitions: Path,
+    mappings: Path,
+    export_processing_result: bool,
+    export_differences: bool,
+):
+
+    result = RegionProcessor.from_directory(
```
danielhuppmann, line 122: I suggest to add a validation against the variable code-list just to be safe.

tests/test_core.py
```diff
@@ -413,3 +414,56 @@ def test_partial_aggregation(input_data, exp_data, warning, caplog):
...
+        "scenario",
+        "region",
+        "variable",
+        "unit",
+        "year",
+        "original",
+        "aggregated",
+        "relative difference (%)",
+    ]
+    exp = pd.DataFrame(difference, columns=columns)
```
danielhuppmann, line 466: Why is that necessary?

danielhuppmann (commented):
Thanks @phackstock, a few suggestions below.

danielhuppmann:
One more comment: please make sure to `reset_index()` before writing `compare` to xlsx, otherwise it is very cumbersome to do some analysis later in Excel (e.g., filtering hides the model/etc index values).

tests/test_core.py
phackstock: It's to check that the differences are exported correctly.

nomenclature/cli.py
phackstock: You mean a validation of the input data right? The `RegionProcessor` already gets validated.

nomenclature/processor/region.py
phackstock: Do you mean leaving out the `100` to make it a percent value?

tests/test_core.py
danielhuppmann: Sorry, I meant - is it necessary to specify the columns explicitly?

nomenclature/cli.py
danielhuppmann: No, I mean validating the variables in the IamDataFrame (or maybe validate as part of RegionProcessor.apply`?). We don't want to perform region-processing without having explicit rules for aggregation (where having a code without region-aggregation-attributes is explicitly using a summation).

nomenclature/processor/region.py
danielhuppmann: Ignore my comment, I misunderstood the code.

nomenclature/cli.py
phackstock: Ok sounds good, I'll add that

tests/test_core.py
danielhuppmann: Ah, got it, thanks!

phackstock:
> One more comment: please make sure to reset_index() before writing compare to xlsx, otherwise it is very cumbersome to do some analysis later in Excel (e.g., filtering hides the model/etc index values).

Not sure I understand the point of why do to this, the output that I currently get of the difference file looks as attached. That should be fine, no?
[difference.xlsx](https://github.com/IAMconsortium/nomenclature/files/11231636/difference.xlsx)

danielhuppmann:
Thanks @phackstock for showing the file, I guess that pandas changed their output structure recently... 

But I still think that `df.reset_index().to_excel(index=False)` will avoid the bold formatting, so that looks cleaner.

nomenclature/processor/region.py
phackstock: Added as a keyword argument

nomenclature/cli.py
phackstock: Done

nomenclature/cli.py
phackstock: Done

nomenclature/cli.py
phackstock: Done

nomenclature/cli.py
phackstock: Switched to using `core.process` so that variable validation is included.

phackstock:
Not sure why test tests are failing, they're running fine on my local machine ... 
I'll check.

phackstock:
Tests are passing now. There are however, some API changes that I've had to make that I'm not super happy with:
* In order to get the differences all the way "out" of the region processing I've had to adjust the interface of `RegionProcessor.apply()` from returning and IamDataFrame to returning a tuple of an IamDataFrame and a pandas DataFrame. 
* As having variable validation for the cli run processing command is convenient I also adjust the return value of the `core.process` function to return the same tuple. 
* This is not idea as we're planning on adding different processors (e.g. for required variables) soon so this messes up the unified `Processor.apply() -> IamDataFrame`.
* I could of course create something like `RegionProcessor.apply_and_capture_differences() -> tuple[IamDataFrame, pd.DataFrame]` and change the apply function to:
```python
def apply(self, ...):
   return self.apply_and_capture_differences()[0]
```
this would preserve the interface.
* The issue with that is for running the processing locally with the variable validation and potentially a number of different validators using the `process` function is quite convenient and avoids code duplication. However, then piping out the differences is difficult as there could be a situation where we don't have any region processing, i.e. no differences data frame.

I can also keep it as is for now and cross that bridge when we get there, which will be very soon.

phackstock:
@danielhuppmann should be good for review now. 
I have settled on the following implementation:
* the `core.process` function will only ever return a single data frame with the results of the processing
* for obtaining the differences between aggregated and original data `RegionProcessor.apply()` with `RegionProcessor.return_aggregation_difference = True` must be used. In this case `RegionProcessor.apply()` returns **both** the aggregation result as well as the differences.
* This is wrapped up in a new cli endpoint called `run-region-processing`. When run as `nomenclature run-region-processing --export-differences differences.xlsx`, the differences will be exported to an Excel file called `differences.xlsx`.
* When running as part of the Scenario Explorer processing and changes in the aggregated and model native data are found, the following log message is written:
```console
If you want to get the differences, run the following command on your local machine:
`nomenclature run-region-processing your_data.xlsx --export-differences difference.xlsx`
 from the processing workflow directory. 
The differences will be exported to `difference.xlsx`."
```

danielhuppmann:
Thanks @phackstock, but I'm confused... I thought that our understanding was that every Processor class should have an `apply()` method that returns only an IamDataFrame, so that they can easily be chained in the `process()` function. This PR seems to violate that principle?

phackstock:
@danielhuppmann, fair point this PR does violate the `apply()` principle but I thought we'd agreed on that implementation.
Initially I had it where the differences would just be saved to a file to not returned by the `apply()` function. But I think we agreed that that's not the best either.
I could create a second function something along the lines of `apply_and_return_differences()` to keep it clean.

phackstock:
> Have a kwarg to apply() so that it by default it returns an IamDataFrame, and it returns a tuple if return_differences=True.

That's how I've currently implemented it https://github.com/phackstock/nomenclature/blob/8614f34cc5dca753280a3834b433445218018ab0/nomenclature/processor/region.py#L434-L436

nomenclature/processor/region.py
```diff
@@ -390,18 +392,17 @@ def validate_with_definition(cls, v, values):
...
         Parameters
         ----------
         df : IamDataFrame
             Input data that the region processing is applied to
-
         Returns
         -------
-        IamDataFrame
-            Processed data
+        Tuple[IamDataFrame, pd.DataFrame]:
```
danielhuppmann, line 404: If we stick with this approach, please extend the docstring to show that the return-type depends on the configuration.

danielhuppmann (commented):
Ok, that's why I missed this approach. Is there a reason that speaks in favor of this explicit setting rather than as a keyword argument? Not sure whether any user would ever realize that this option exists?

Also, I guess that two or three lines in a Jupyter notebook is actually the more convenient solution for our users than a CLI...

See also some inline comments.

phackstock:
>  Is there a reason that speaks in favor of this explicit setting rather than as a keyword argument?
I initially did it this way to keep the inputs of the apply function the same `Processor.apply(df: IamDataFrame)`. 

The main advantage though that I see of using an attribute is that since `process` takes a list of processors we can set all the attributes ahead of time and then just let `process` apply them all in a loop by only calling `Processor.apply(df)` (without any extra arguments needed):
https://github.com/IAMconsortium/nomenclature/blob/b83101fd4f101d45aa568d8aedfaaf338a5672f1/nomenclature/core.py#L68-L69

If we introduced a key-word argument to `apply` we would break this pattern. 
In this specific case it doesn't matter since the key word argument would change the return type of `apply` so the above loop would break anyway. 
But in the future we might want to have more options for the `apply` function of a processor that alter the behavior without changing the return type. In this case using object attributes we could still use the `process` function.

> Not sure whether any user would ever realize that this option exists?
If the user goes through the (at least for now) recommended way of using the cli to obtain a differences file, the user does not need to know of this option. It's all handled internally anyway. 
If you think that recommending a jupyter notebook is the better way to go that's a different story then...

Would you then change the log message to:
```console
Past the following into a jupyter notebook to get the differences:
    
    data_structure_definition = DataStructureDefinition(path / definitions)
    region_processor = RegionProcessor.from_directory(
        path / mappings, data_structure_definition
    )
    region_processor.return_aggregation_difference = True
    result, difference = region_processor.apply(IamDataFrame(input_data_file))
```
In which case the user also doesn't really need to know about the option if it's copy and paste.

If you think I should remove the cli option in favor of a jupyter notebook I can also do that. Would that be a jupyter notebook that's part of the GitHub workflow repo for every project or something that the user needs to create ad-hoc?

nomenclature/processor/region.py
```diff
@@ -565,14 +584,27 @@ def _compare_and_merge(original: pd.Series, aggregated: pd.Series) -> IamDataFra
...
+        (difference["original"] - difference["aggregated"]) / difference["original"]
+    )
+    difference = difference.sort_values("difference (%)", ascending=False)
+    if difference is not None and len(difference):
+        logging.warning(
+            f"Difference between original and aggregated data:\n{difference}"
+        )
+    if not return_aggregation_difference:
+        logging.info(
+            "If you want to get the differences, run the following command on your "
```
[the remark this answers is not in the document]
phackstock: It's true, I'd say though if it contains all the information necessary and saves the user a trip to the documentation, that's not a bad thing.
I can change it though and create an entry in the docs.

danielhuppmann:
I'm not convinced by your argument that we should do an overly complicated set-attribute-for-the-class implementation that changes the return-type because there might be other processors where we need such a complicated approach.

The more we go back and forth, the more I think that a separate function is the cleaner and more user-friendly approach - I'd go for `check_region_aggregation()` or similar.

About the log message - no, this is again extremely verbose, and still does not have enough information for any non-expert user to know what to do - how to get the definitions folder, how to do the imports, etc.

Instead, I would extend the User Guide in the documentation where the example code is shown, plus explanations where to get the definitions folder, etc.

nomenclature/processor/region.py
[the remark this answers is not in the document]
phackstock: Updated

phackstock:
I have implemented the following updates:
* added a function called `check_region_aggregation`, it takes a pyam dataframe and returns the differences between aggregated and original data.
* `RegionProcessor.apply()` now always returns only one pyam data frame which is the processing results.
* Attributes for altering `RegionProcessor.apply` behavior are gone
* Added a user guide to the documentation on how to locally get the differences
* Updated the log message that now links to the documentation

phackstock:
@danielhuppmann since you think the cli is too much for the non-expert user should I just remove the option?

danielhuppmann:
> @danielhuppmann since you think the cli is too much for the non-expert user should I just remove the option?

No, keep the CLI - I think having both the CLI implementation and the Python-code-example is the most user-friendly approach!

nomenclature/cli.py
```diff
@@ -93,3 +96,32 @@ def cli_valid_project(
...
     assert_valid_structure(path, definitions, mappings, required_data, dimensions)
+
+
+@cli.command("run-region-processing")
+@click.argument("input_data_file", type=click.Path(exists=True, path_type=Path))
+@click.option("-p", "path", type=click.Path(exists=True, path_type=Path), default=".")
+@click.option("-d", "--definitions", type=str, default="definitions")
+@click.option("-m", "--mappings", type=str, default="mappings")
+@click.option("--output", type=click.Path(path_type=Path), default="results.xlsx")
+@click.option("--export-differences", type=click.Path(path_type=Path), default=None)
```
danielhuppmann, line 107: For conciseness…
```suggestion
@click.option("--differences", type=click.Path(path_type=Path), default=None)
```

nomenclature/cli.py
```diff
@@ -93,3 +96,32 @@ def cli_valid_project(
...
+        path / mappings, data_structure_definition
+    )
+    region_processor.return_aggregation_difference = True
+
+    result, difference = region_processor.apply(IamDataFrame(input_data_file))
+
+    if output:
+        result.to_excel(output)
+    if export_differences:
+        difference.reset_index().to_excel(export_differences, index=False)
```
danielhuppmann, line 127: You can use the pyam-write-sheet utility to get nicer output formatting.

nomenclature/cli.py
```diff
@@ -93,3 +96,32 @@ def cli_valid_project(
...
     assert_valid_yaml(path)
     assert_valid_structure(path, definitions, mappings, required_data, dimensions)
+
+
+@cli.command("run-region-processing")
+@click.argument("input_data_file", type=click.Path(exists=True, path_type=Path))
+@click.option("-p", "path", type=click.Path(exists=True, path_type=Path), default=".")
+@click.option("-d", "--definitions", type=str, default="definitions")
+@click.option("-m", "--mappings", type=str, default="mappings")
+@click.option("--output", type=click.Path(path_type=Path), default="results.xlsx")
```
danielhuppmann, line 106:
```suggestion
@click.option("--processed-data", type=click.Path(path_type=Path), default="results.xlsx")
```

doc/source/user_guide/model-mapping.rst
```diff
@@ -120,9 +120,39 @@ are compared and combined according to the following logic:
...
+  from nomenclature import DataStructureDefinition, RegionProcessor
+
+  data = IamDataFrame("/path/to/your/input/data.xlsx")
+
+  dsd = DataStructureDefinition("definitions")
+  processor = RegionProcessor.from_directory("mappings", dsd)
+
+  # get the differences as a pandas dataframe
+  # the value for the relative tolerances can be adjusted, defaults to 0.01
+  differences = processor.check_region_aggregation(data, rtol_difference=0.01)
```
danielhuppmann, line 154: Shouldn’t this return a tuple?

doc/source/user_guide/model-mapping.rst
```diff
@@ -120,9 +120,39 @@ are compared and combined according to the following logic:
...
+  differences.to_excel("differences.xlsx")
+
+For details on the check_region_aggregation function please refer to
```
danielhuppmann, line 157:
```suggestion
For details on this feature please refer to
```

tests/test_core.py
```diff
@@ -363,7 +364,7 @@ def test_region_processing_skip_aggregation(model_name, region_names):
             [["m_a", "s_a", "World", "Primary Energy", "EJ/yr", 5, 6]],
             [
                 "Difference between original and aggregated data:",
-                "m_a   s_a      World  Primary Energy EJ/yr 2005         5           4",
+                "m_a   s_a      World  Primary Energy EJ/yr 2005         5",
```
danielhuppmann, line 367: Why does this not check two values?

tests/test_core.py
```diff
@@ -375,7 +376,7 @@ def test_region_processing_skip_aggregation(model_name, region_names):
             [["m_a", "s_a", "World", "Variable B", "EJ/yr", 4, 6]],
             [
                 "Difference between original and aggregated data:",
-                "m_a   s_a      World  Variable B EJ/yr 2005         4           3",
+                "m_a   s_a      World  Variable B EJ/yr 2010         6",
```
danielhuppmann, line 379: Why did the value change?

danielhuppmann (commented):
Thanks, a few minor suggestions and clarifications below…

danielhuppmann (commented):
Thanks, a few minor suggestions and clarifications below…

doc/source/user_guide/model-mapping.rst
```diff
@@ -120,9 +120,39 @@ are compared and combined according to the following logic:
...
       necessary to check the logs to find out if there were any differences. This is
       intentional since some differences might be expected.
+
+
+How to get aggregation differences locally
+------------------------------------------
+
+In order to get the full differences perform the following steps:
+
+1. Make sure you have `pyam-iamc >= 1.7.0`, `nomenclature-iamc>=0.9.2` and
```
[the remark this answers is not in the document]
phackstock: Sure let's make it 0.10 :smile:

nomenclature/cli.py
phackstock: Does that work for pandas dataframes as well?

doc/source/user_guide/model-mapping.rst
phackstock: The way that I've implemented it for now is that it only returns the differences. I could of course have it return both the result and the differences

tests/test_core.py
phackstock: It's because I added the relative difference column which makes the output too wide so now the warning message looks like this:
```console
2023-05-26 10:36:51 WARNING  Difference between original and aggregated data:
                                                 original  ...  difference (%)
model scenario region variable       unit  year            ...                
m_a   s_a      World  Primary Energy EJ/yr 2005         5  ...            20.0
```
I'm not super happy about it, I could change it to omit the _difference_ column and just put _original_ and _aggregated_.

tests/test_core.py
phackstock: It's because I now sort the differences by the relative difference and the most severe one is for 2010 and no longer for 2005. This is what the full warning message looks like:
```console
2023-05-26 10:39:50 WARNING  Difference between original and aggregated data:
                                             original  ...  difference (%)
model scenario region variable   unit  year            ...                
m_a   s_a      World  Variable B EJ/yr 2010         6  ...       33.333333
                                       2005         4  ...       25.000000
```
I just don't test for the entire message

nomenclature/cli.py
danielhuppmann: Yes, see https://github.com/IAMconsortium/pyam/blob/1f33e21468d67d365984d0dd7ec44de30c6c60be/pyam/utils.py#L102

But I just realize that it requires a `pd.ExcelWriter`, not just a string to the path, so maybe not worth the effort...

doc/source/user_guide/model-mapping.rst
danielhuppmann: I think returning both is more useful - otherwise, you have to run the computation twice if you want to have both the aggregated dataset *and* the differences (or you go to the internal method, which one should avoid).

tests/test_core.py
danielhuppmann: Indeed, but this depends anyway on the length of the variable name, right? Maybe `pd.set_option('display.max_columns', None)` is your friend?

phackstock: Thanks for the suggestion, it now wraps it around two lines:
```console
2023-05-30 14:02:56 WARNING  Difference between original and aggregated data:
                                                 original  aggregated  \
model scenario region variable       unit  year                         
m_a   s_a      World  Primary Energy EJ/yr 2005         5           4   

                                                 difference (%)  
model scenario region variable       unit  year                  
m_a   s_a      World  Primary Energy EJ/yr 2005            20.0  
```

doc/source/user_guide/model-mapping.rst
phackstock: Sounds good, updated

nomenclature/cli.py
phackstock: Tried getting it to work with `pyam.write_sheet` but for some reason the output file is empty. I'll leave it as is.

phackstock:
Updated the return value of `check_region_aggregation`, it now returns both the processing results and the differences.
Changed the width of the pandas console output so that it will write the full length of the differences data frame to the console.

doc/source/user_guide/model-mapping.rst
```diff
@@ -120,9 +120,39 @@ are compared and combined according to the following logic:
...
+  from nomenclature import DataStructureDefinition, RegionProcessor
+
+  data = IamDataFrame("/path/to/your/input/data.xlsx")
+
+  dsd = DataStructureDefinition("definitions")
+  processor = RegionProcessor.from_directory("mappings", dsd)
+
+  # get the differences as a pandas dataframe
+  # the value for the relative tolerances can be adjusted, defaults to 0.01
+  differences = processor.check_region_aggregation(data, rtol_difference=0.01)
```
danielhuppmann, line 154:
```suggestion
  processed_data, differences = processor.check_region_aggregation(data, rtol_difference=0.01)
```

nomenclature/cli.py
```diff
@@ -93,3 +96,68 @@ def cli_valid_project(
     """
     assert_valid_yaml(path)
     assert_valid_structure(path, definitions, mappings, required_data, dimensions)
+
+
+@cli.command("run-region-processing")
```
danielhuppmann, line 101: Maybe rename to `check-region-aggregation` for consistency with the function?

doc/source/user_guide/model-mapping.rst
```diff
@@ -120,9 +120,39 @@ are compared and combined according to the following logic:
...
 
       Please note that in case of differences no error is raised. Therefore it is
       necessary to check the logs to find out if there were any differences. This is
       intentional since some differences might be expected.
+
+
+How to get aggregation differences locally
+------------------------------------------
+
+In order to get the full differences perform the following steps:
```
danielhuppmann, line 134:
```suggestion
In order to get the differences between the original data (e.g., results reported by the model)
and the data aggregated according to the region mapping, perform the following steps:
```

nomenclature/cli.py
```diff
@@ -93,3 +96,68 @@ def cli_valid_project(
...
+def run_region_processing(
+    input_data_file: Path,
+    workflow_directory: Path,
+    definitions: str,
+    mappings: str,
+    processed_data: Optional[Path],
+    differences: Optional[Path],
+):
+    """Perform region processing an optionally return differences between aggregated
+    and model native data
```
danielhuppmann, line 124:
```suggestion
    """Perform region processing and compare aggregated and original data
```

doc/source/user_guide/model-mapping.rst
```diff
@@ -120,9 +120,39 @@ are compared and combined according to the following logic:
...
+
    .. note::
 
       Please note that in case of differences no error is raised. Therefore it is
       necessary to check the logs to find out if there were any differences. This is
       intentional since some differences might be expected.
+
+
+How to get aggregation differences locally
+------------------------------------------
```
danielhuppmann, line 132:
```suggestion
Computing differences between original and aggregated data
----------------------------------------------------------
```

danielhuppmann:
Almost there...

nomenclature/processor/region.py
```diff
@@ -565,14 +614,28 @@ def _compare_and_merge(original: pd.Series, aggregated: pd.Series) -> IamDataFra
...
+    difference = compare[
+        ~np.isclose(compare["original"], compare["aggregated"], rtol=rtol)
+    ]
+    difference["difference (%)"] = 100 * np.abs(
+        (difference["original"] - difference["aggregated"]) / difference["original"]
+    )
+    difference = difference.sort_values("difference (%)", ascending=False)
+    if difference is not None and len(difference):
+        with pd.option_context("display.max_columns", None):
+            logging.warning(
```
danielhuppmann, line 626: Why is this `logging` and not `logger`?

nomenclature/processor/region.py
```diff
@@ -565,14 +614,28 @@ def _compare_and_merge(original: pd.Series, aggregated: pd.Series) -> IamDataFra
...
+                f"Difference between original and aggregated data:\n{difference}"
+            )
+    if not return_aggregation_difference:
+        logging.info(
```
danielhuppmann, line 630: As above, maybe `logger`?

nomenclature/cli.py
phackstock: Good point

nomenclature/processor/region.py
phackstock: Ah good point, changed to logger.

danielhuppmann (approved):
Looks good, thanks!

PR #238, opened 2023-04-20
Updated nomenclature documentation by removing the methods from being listed in the sidebar

GretchenSchowalter:
By adding :noindex: to these files I was able to remove the methods from being listed in the sidebar of the nomenclature documentation. The one tradeoff is that they are also no longer listed on the api documentation page that has the table of contents for that section.


danielhuppmann (changes requested):
Sorry, but not showing the methods is not an option. This is how I implemented this for pyam: https://github.com/IAMconsortium/pyam/blob/main/docs/_templates/navigation.html

danielhuppmann (approved):
Looks good, thanks!

PR #239, opened 2023-04-20
RegionCode class added with one hierarchy attribute

### LINKED ISSUE #194: Add RegionCode class
For improved clarity it would be nice to add a `RegionCode` which has the additional attribute hierarchy.

GretchenSchowalter:
I just created a simple RegionCode class in code.py and updated the RegionCodeList class. Let me know if I missed something or did this incorrectly. Closes #194

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -208,4 +208,4 @@ class RegionCode(Code):
 
     """
 
-    hierarchy: Optional[str] = None
+    hierarchy: str = None
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -194,3 +194,18 @@ class VariableCode(Code):
             if self.region_aggregation is not None
             else self.pyam_agg_kwargs
         )
+
+
+class RegionCode(Code):
+    """A subclass of Code specified for regions
+
+    Attributes
+    ----------
+    name : str
+        Name of the RegionCode
+    hierarchy : str
+        Hierarchy of the RegionCode
+
+    """
+
+    hierarchy: Optional[str] = None
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -589,3 +589,38 @@ class RegionCodeList(CodeList):
 
         """
         return sorted(list(set(v.hierarchy for v in self.mapping.values())))
+
+    def filter(self, hierarchy: str) -> "RegionCodeList":
+        """Return a filtered RegionCodeList object
+
+        Parameters
+        ----------
+        hierarchy : str
+            String with filter parameter.
+
+        Raises
+        ------
+        ValueError
+            Provided hierarchy is not compatible with the given model.
+
+        Returns
+        -------
+        :class:'RegionCodeList'
+
+        Notes
+        -----
+        The following arguments are available for filtering:
+
+        - 'hierarchy': filter by the hierarchy of each region
+
+        """
+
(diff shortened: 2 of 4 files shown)


nomenclature/code.py
```diff
@@ -194,3 +194,18 @@ def agg_kwargs(self) -> Dict[str, Any]:
...
+    Attributes
+    ----------
+    name : str
+        Name of the RegionCode
+    hierarchy : str
+        Hierarchy of the RegionCode
+
+    """
+
+    hierarchy: Optional[str] = None
```
danielhuppmann, line 211: This should be a required attribute.
```suggestion
    hierarchy: str
```

nomenclature/codelist.py
```diff
@@ -548,7 +548,7 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
         RegionCodeList
 
         """
-        code_list: List[Code] = []
+        code_list: List[RegionCode] = []
```
phackstock, line 551: Just for completeness I would add this line above:
```python
mapping: Dict[str, RegionCode] = {}
```
and this one below:
```python
code_basis: ClassVar = RegionCode
```

nomenclature/codelist.py
```diff
@@ -562,14 +562,14 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
             for top_level_cat in _code_list:
                 for top_key, _codes in top_level_cat.items():
                     for item in _codes:
-                        code = Code.from_dict(item)
+                        code = RegionCode.from_dict(item)
```
phackstock, line 565: You could streamline this slightly: 
```suggestion
                        code_list.append(RegionCode.from_dict({
                        **item,
                         "hierarchy": top_key,
                          "file":yaml_file.relative_to(path.parent).as_posix()})
                          )
```
although I'm not sure it's much more readable.

phackstock (commented):
Two small suggestions and a comment other than that it looks good to me.

nomenclature/codelist.py
phackstock: As discussed bilaterally, please disregard.

danielhuppmann (commented):
I'm wondering why there is still a dedicated method `filter()` in the **RegionCodeList** class - wouldn't this now be covered by the inherited method of the **CodeList** class?

GretchenSchowalter:
Yep, that makes sense. I will take the method and the corresponding tests out.

danielhuppmann:
No, leave the tests! They should continue to work by automatically inheriting the `filter()` method from the **CodeList** class.

phackstock:
Also taking it out without changing anything about `CodeList.filter` would not be good because it return a generic `CodeList`.
Then we would have the strange situation in `RegionCodeList.filter` would return a `CodeList` and not a `RegionCodeList`.
You'd have to adapt `CodeList.filter` to return whatever CodeList it was called from, i.e. `CodeList` for a generic one and `RegionCodeList` for a RegionCodeList.

danielhuppmann:
Sorry, I think that my comment sent you down a rabbit hole that we don't need to open right now. Can you simply re-insert the `RegionCodeList.filter()` and the corresponding tests? Then the PR is good to be merged - and we can worry about proper sub-classing later...

phackstock (approved):
Looks good to me as well.

Issue #224, opened 2023-02-09
Validate "countries" attribute against ISO3 codes


danielhuppmann:
In https://github.com/iiasa/eu-climate-advisory-board-workflow/pull/32 (and similar pull requests), we added an attribute "countries" to the region definitions. It would be useful to add a validation that these are valid ISO3 codes.

Also, this might require some fidgeting translating from a human-readable list to a proper Python list.

danielhuppmann:
Per suggestion by @byersiiasa, fyi @phackstock

PR #240, opened 2023-05-02
Added the countries attribute and check_iso3_codes() method to RegionCode object

GretchenSchowalter:
Hello! This was my initial idea in tackling issue #224. Currently, the check_iso3_codes() method returns a Boolean, but we can change that if there is a better idea in how to do this!

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -1,7 +1,6 @@
 import json
 import re
 import pycountry
-import logging
 from keyword import iskeyword
 from typing import Any, Dict, List, Optional, Set, Union
 
@@ -213,9 +212,15 @@ class RegionCode(Code):
     hierarchy: str = None
     countries: List[str] = None
 
-    def check_iso3_codes(self) -> bool:
-        for country in self.countries:
+    @validator("countries")
+    def check_iso3_codes(cls, v, values) -> bool:
+        invalid_iso3_codes: List[str] = []
+        for country in v:
             if pycountry.countries.get(alpha_3=country) is None:
-                logging.warning(f"{country} is not a valid iso3 country code!")
-                return False
-        return True
+                invalid_iso3_codes.append(country)
+        if invalid_iso3_codes:
+            raise ValueError(
+                f"Region {values['name']} has invalid\n"
+                " ISO3 country codes: {invalid_iso3_codes}"
+            )
+        return v
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -4,7 +4,7 @@ import pycountry
 from keyword import iskeyword
 from typing import Any, Dict, List, Optional, Set, Union
 
-from pydantic import BaseModel, Field, validator, ValidationError
+from pydantic import BaseModel, Field, validator
 
 
 class Code(BaseModel):
@@ -219,8 +219,8 @@ class RegionCode(Code):
             if pycountry.countries.get(alpha_3=country) is None:
                 invalid_iso3_codes.append(country)
         if invalid_iso3_codes:
-            raise ValidationError(
+            raise ValueError(
                 f"Region {values['name']} has invalid\n"
-                " ISO3 country codes: {invalid_iso3_codes}"
+                f" ISO3 country codes: {invalid_iso3_codes}"
             )
         return v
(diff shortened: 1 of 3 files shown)


tests/test_code.py
```diff
@@ -55,3 +56,27 @@ def test_RegionCode_hierarchy_attribute():
     )
 
     assert reg.hierarchy == "R5"
+
+
+def test_RegionCode_iso3_code():
+    reg = RegionCode(
+        name="OECD",
```
danielhuppmann, line 63: I suggest to use a more "correct" test, e.g. Western Europe that has France, Germany, Spain, Portugal...

nomenclature/code.py
```diff
@@ -209,3 +211,11 @@ class RegionCode(Code):
     """
 
     hierarchy: str = None
+    countries: List[str] = None
+
+    def check_iso3_codes(self) -> bool:
+        for country in self.countries:
+            if pycountry.countries.get(alpha_3=country) is None:
+                logging.warning(f"{country} is not a valid iso3 country code!")
```
danielhuppmann, line 219: Better to keep a list all invalid countries in this check, and return a warning like:
```python
f"Region {name} has invalid ISO3 country codes: {invalid_iso3_codes}"
```

danielhuppmann (commented):
Looks good, two minor comments.

Good question whether to simply write a warning or do a "hard" ValueError - any preference, @phackstock?

danielhuppmann:
And you'll have to add the pycountry package to the dependencies...

phackstock:
> Good question whether to simply write a warning or do a "hard" ValueError - any preference, @phackstock?

@danielhuppmann hm, I could see arguments for both but my preference is always to fail as soon and as loudly as possible. So I'd vote error.

nomenclature/code.py
phackstock: That would be good to turn into a pydantic validator. Reason being is that it is then automatically called whenever you create an instance of `RegionCode`. For now it has to be called explicitly.

tests/test_code.py
```diff
@@ -55,3 +56,27 @@ def test_RegionCode_hierarchy_attribute():
...
+    assert reg.check_iso3_codes() is True
+
+
+def test_RegionCode_iso3_code_fail(caplog):
+    reg = RegionCode(
+        name="OECD",
+        hierarchy="R5",
+        countries=["ALB", "BGR", "DEX", "AUS"],
+    )
+    caplog.set_level(logging.WARNING)
```
phackstock, line 77: Do you need this and the following line?

phackstock (changes requested):
Looks almost ready to me. 
One suggestion below regarding the iso3 validator.

GretchenSchowalter:
Hello! I've been having a little trouble figuring out how to implement the @validator to the method. Currently the test_RegionCode_iso3_codes_fail() is still failing, so if you have any suggestions that would be great!

nomenclature/code.py
```diff
@@ -209,3 +210,17 @@ class RegionCode(Code):
...
+    countries: List[str] = None
+
+    @validator("countries")
+    def check_iso3_codes(cls, v, values) -> bool:
+        invalid_iso3_codes: List[str] = []
+        for country in v:
+            if pycountry.countries.get(alpha_3=country) is None:
+                invalid_iso3_codes.append(country)
+        if invalid_iso3_codes:
+            raise ValidationError(
```
phackstock, line 222: This needs to be a `ValueError` (or any kind of other error except `pydantic.ValidationError`:
```suggestion
            raise ValueError(
```
Reason being, pydantic collects all the errors found during validation and raises them as one common error which is a `pydantic.ValidationError`. The individual errors, however, cannot be `pydantic.ValidationError`.

nomenclature/code.py
```diff
@@ -209,3 +210,17 @@ class RegionCode(Code):
...
+                f"Region {values['name']} has invalid\n"
+                " ISO3 country codes: {invalid_iso3_codes}"
```
phackstock, line 224: You'll need another f here to make string interpolation work
```suggestion
                f" ISO3 country codes: {invalid_iso3_codes}"
```

phackstock (approved):
Looks good to me thanks @GretchenSchowalter.
Good to be merged form my side.

danielhuppmann (approved):
Nice work, thank you!

PR #244, opened 2023-05-09
General cleanup

phackstock:
As I ran into an issue with a mutable default argument in #243, this PR eliminates the two remaining ones in the package.

Since this is, on purpose, branched off of #243, the idea would be to first merge #243 and then this one.

### Update

I used pylint to find the mutable defaults. While I was there I also fixed some other issues that it found.

### DIFF
--- a/nomenclature/__init__.py
+++ b/nomenclature/__init__.py
@@ -30,7 +30,7 @@ except LookupError:
     __version__ = version("nomenclature-iamc")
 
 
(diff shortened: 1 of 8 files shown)


danielhuppmann:
@phackstock, can you please do a rebase before I review?

phackstock:
@danielhuppmann everything rebased, good to review now.

danielhuppmann (approved):
Ok, these are a lot of changes, but it all looks pretty straightforward to me...

PR #245, opened 2023-05-10
Streamline code list filter

phackstock:
As discussed at the end of #236 this PR removes `RegionCodeList.filter()` in favor of using the more generic `CodeList.filter()`.
The one slight difference, and reason why there's currently one test failing is that the logic of what happens if the filter function returns nothing is/was different between `RegionCodeList.filter()` and `CodeList.filter()`:

* If `CodeList.filter()` returns nothing, a warning is issued.
* If `RegionCodeList.filter()` returns nothing, an error is raised.

With removing the latter, there's now only ever a warning, which to me is fine. 

Looking forward to your input @GretchenSchowalter and @danielhuppmann.

-----

EDIT: I removed the the unit test for `RegionCodeList.filter()` that tests for the empty codelist. Reason being is that there's no longer any difference between `RegionCodeList.filter()` and `CodeList.filter()` and we already have a unit test for an empty return.

### DIFF
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -406,11 +406,11 @@ class CodeList(BaseModel):
         # Returns True if code satisfies all filter parameters
         def _match_attribute(code, kwargs):
             return all(
-                hasattr(code, attribute) and getattr(code, attribute, None) == value
+                hasattr(code, attribute) and getattr(code, attribute) == value
                 for attribute, value in kwargs.items()
             )
 
-        filtered_codelist = CodeList(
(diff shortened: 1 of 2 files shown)


GretchenSchowalter:
It all looks excellent to me! Thank you for looking into this.

phackstock:
> Oh, also I thought in pyam when returning an empty object it also gives a warning, so I would think just the warning is fine?

Good point, keeping consistency is usually a good idea.

GretchenSchowalter (approved):
This looks great to me! Thanks, @phackstock !

Issue #241, opened 2023-05-02
MetaCode brainstorming


phackstock:
Following our discussion earlier, I'm sketching out some ideas for a `MetaCodeList`.
I think it's easiest to start at what the yaml file is supposed to look like:

```yaml
- Meta category with boolean values:
  allowed_values: [true, false]
  default: false
- Meta category for tracking versions:
  data_type: int
  required: true
- Meta category only required for Downscaling:
  apply_to: "Downscaling*"
  required: True
```

the three above examples illustrate some of the use cases that I can see, listing them in order:
1. _Meta category with boolean values_ has allowed values `true` and `false`, defaulting to `false`
2. _Meta category for tracking versions_ is required and requires an integer. Allowed values not really make sense here since we expect increasing integer numbers.
3. _Meta category only required for Downscaling_ should only be applied to models matching the pattern `Downscaling*`, in which case, however, it is required.

Looking at this again @danielhuppmann I'll retract my previous suggestion of making the entries in `allowed_values` `MetaCodes` as that seems too complicated and not every entry in the sample `MetaCodeList` even features `allowed_values`.

It's probably better to translate the above structure into something like:

```python
class MetaCode(Code):
    allowed_values: Optional(list[any])
    default: Optional(any)
    data_type: Optional(type)
    required: Optional(bool)
```

then we'd have a `MetaCodeList` to hold a list of `MetaCodes`.

danielhuppmann:
I think that this is a good model to follow, but I suggest to implement it step by step.

The first PR should be the `MetaCode` and `MetaCodeList` classes read from all yaml files in a `meta` folder, with only the attribute `allowed_values`.

danielhuppmann:
The second PR should be a `MetaValidator` (with an `apply` function), which can be passed to the the processor-list in the `process`-function...

phackstock:
@danielhuppmann sounds good to me

danielhuppmann:
@GretchenSchowalter, can you take a first stab at this? Please look at this tutorial to better understand the "meta" indicators: https://pyam-iamc.readthedocs.io/en/stable/tutorials/pyam_first_steps.html

phackstock:
@danielhuppmann, I'd say we can close this, right?

danielhuppmann:
Closing, with a follow-up at #270

PR #246, opened 2023-05-10
MetaCode and MetaCodeList classes with allowed_values attribute

GretchenSchowalter:
Hello! This is my first attempt at addressing issue #241. I have added MetaCode and MetaCodeList classes, where MetaCode has the attribute allowed_values. Within the MetaCodeList class, there is a method from_yaml_files() (which is very similar to the from_directory() method in RegionCodeList) which reads the .yaml files and creates a MetaCodeList. I've also added two very simple tests. I wasn't really sure how to guarantee that the from_yaml_files() method read from a folder called meta, so I would be grateful for a suggestion!


nomenclature/codelist.py
```diff
@@ -624,3 +624,62 @@ def filter(self, hierarchy: str) -> "RegionCodeList":
...
+        Dictionary of `MetaCode` objects
+
+    """
+
+    # class variable
+    code_basis: ClassVar = MetaCode
+    validation_schema: ClassVar[str] = "meta indicators"
+
+    @classmethod
+    def from_yaml_files(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
```
danielhuppmann, line 646: Why not keep the `from_directory()` convention here? I think we should keep the meta-indicators in a "meta" folder for consistency.

tests/test_code.py
```diff
@@ -55,3 +55,12 @@ def test_RegionCode_hierarchy_attribute():
     )
 
     assert reg.hierarchy == "R5"
+
+
+def test_MetaCode_allowed_values_attribute():
+    reg = MetaCode(
```
danielhuppmann, line 61:
```suggestion
    meta = MetaCode(
```

nomenclature/definition.py
```diff
@@ -1,139 +1,143 @@
...
-        sheet_name : str, optional
-            Name of sheet that will have the codelist. If *None*, use the codelist name.
-        sort_by_code : bool, optional
-            Sort the codelist before exporting to file.
-        **kwargs
-            Passed to :class:`pandas.ExcelWriter` (if *excel_writer* is path-like).
-        """
-        # TODO write all dimensions to the file
-        self.variable.to_excel(excel_writer, sheet_name, sort_by_code, **kwargs)
+import logging
```
danielhuppmann, line 1: Not sure why GitHub thinks this entire file was changed?

nomenclature/definition.py
```diff
@@ -1,139 +1,143 @@
...
+
+from nomenclature.codelist import (
+    CodeList,
+    RegionCodeList,
+    VariableCodeList,
+)
+from nomenclature.validation import validate
+
+logger = logging.getLogger(__name__)
+SPECIAL_CODELIST = {"variable": VariableCodeList, "region": RegionCodeList}
```
danielhuppmann, line 17: Please add the MetaCodeList here.

danielhuppmann (commented):
Thanks @GretchenSchowalter, a few suggestions inline...

nomenclature/code.py
```diff
@@ -209,3 +209,16 @@ class RegionCode(Code):
     """
 
     hierarchy: str = None
+
+
+class MetaCode(Code):
+    """A subclass of Code
```
phackstock, line 215: Maybe writing on the purpose would be better. The fact that this is a subclass of Code is already given by the line above.

nomenclature/code.py
```diff
@@ -209,3 +209,16 @@ class RegionCode(Code):
...
+
+    Attributes
+    ----------
+    allowed_values : Optional(list[any])
+        An optional list of allowed values
+
+    """
+
+    allowed_values: Optional[List[Any]]
```
phackstock, line 224: Should this be optional?

nomenclature/codelist.py
```diff
@@ -624,3 +624,62 @@ def filter(self, hierarchy: str) -> "RegionCodeList":
...
+    name : str
+        Name of the MetaCodeList
+    mapping : dict
+        Dictionary of `MetaCode` objects
+
+    """
+
+    # class variable
+    code_basis: ClassVar = MetaCode
+    validation_schema: ClassVar[str] = "meta indicators"
```
phackstock, line 643: I couldn't find a usage of `validation_schema`.
```suggestion
```

nomenclature/codelist.py
```diff
@@ -624,3 +624,62 @@ def filter(self, hierarchy: str) -> "RegionCodeList":
...
+    ----------
+    name : str
+        Name of the MetaCodeList
+    mapping : dict
+        Dictionary of `MetaCode` objects
+
+    """
+
+    # class variable
+    code_basis: ClassVar = MetaCode
```
phackstock, line 642: I couldn't find a usage of `code_basis`.
```suggestion
```

tests/test_code.py
```diff
@@ -55,3 +55,12 @@ def test_RegionCode_hierarchy_attribute():
     )
 
     assert reg.hierarchy == "R5"
+
+
+def test_MetaCode_allowed_values_attribute():
```
phackstock, line 60: Not sure I see the point of this test. Creating an object directly and checking its value is something that should be covered by the python standard library. 
Reading a `MetaCode` from a file would perhaps be more useful

tests/data/meta/allowed_values_2.yaml
```diff
@@ -0,0 +1,4 @@
+- Meta cat with boolean values:
+  allowed_values: [True, False]
```
phackstock, line 2: In keeping consistency with the other CodeLists `allowed_values` should be indented like so:

```suggestion
    allowed_values: [True, False]
```
you'd need to slightly change the way you read the yaml file but I'd say it's worth keeping consistency.

nomenclature/codelist.py
```diff
@@ -624,3 +624,62 @@ def filter(self, hierarchy: str) -> "RegionCodeList":
...
+            with open(yaml_file, "r", encoding="utf-8") as stream:
+                _code_list = yaml.safe_load(stream)
+
+            for top_level_cat in _code_list:
+                code = MetaCode(
+                    name=list(top_level_cat.keys())[0],
+                    mapping=top_level_cat,
+                )
+                code.file = yaml_file.relative_to(path.parent).as_posix()
+                code.allowed_values = list(top_level_cat.values())[1]
```
phackstock, line 679: You'll need to update this line if you address my comment about the indentation of the `allowed_values` attribute in the yaml file.

phackstock (commented):
Thanks @GretchenSchowalter for starting this work on the MetaCodes.
Also a few suggestions from my side.

nomenclature/codelist.py
phackstock: I second that one it should definitely be called `from_directory()` otherwise `DataStructureDefinition` cannot use it.
Probably should implement a `MetaCodeList` class that enforces that ...

phackstock: Looking at this again, I don't think the method is actually needed at all. Since `MetaCodeList` inherits from `CodeList` and `CodeList` implements `from_directory` we should be good.

GretchenSchowalter:
Hello! Thanks for all your feedback. I decided to try getting rid of the method I created in the MetaCodeList class and just use the from_directory() from the CodeList class, but to do so I had to make a few adjustments. For example, making sure that from_directory() creates a list of MetaCodes, rather than a list of Codes. I also had to create a meta_schema.yaml file for the validate() to work within the from_directory() method.

nomenclature/validation_schemas/meta_schema.yaml
```diff
@@ -0,0 +1,36 @@
+$schema: 'https://json-schema.org/draft/2020-12/schema'
```
phackstock, line 1: I would leave this whole thing out. 
It was our initial attempt at making the yaml files more robust. In my opinion though I'd drop all the schema files (in a separate PR) as it usually doesn't add that much information but a lot of complexity in our machinery.

nomenclature/codelist.py
```diff
@@ -193,7 +193,7 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
         instance of cls (CodeList if not inherited)
 
         """
-        code_list: List[Code] = []
+        code_list: List[cls.code_basis] = []
```
phackstock, line 196: What you've written is not wrong but since all the special codes `RegionCode`, `VariableCode` inherit from `Code` anyway I would keep it like this, for readability reasons. 

```suggestion
        code_list: List[Code] = []
```

type hints in this case are purely for the developer anyway and I think it's more readable with `Code`.

nomenclature/codelist.py
```diff
@@ -212,7 +212,7 @@ def from_directory(cls, name: str, path: Path, file_glob_pattern: str = "**/*"):
                 code.file = yaml_file.relative_to(path.parent).as_posix()
                 code_list.append(code)
         code_list = cls._parse_and_replace_tags(code_list, path, file_glob_pattern)
-        mapping: Dict[str, Code] = {}
+        mapping: Dict[str, cls.code_basis] = {}
```
phackstock, line 215: Same argument as above
```suggestion
        mapping: Dict[str, Code] = {}
```

nomenclature/codelist.py
```diff
@@ -27,7 +27,7 @@ def read_validation_schema(i):
     return schema
 
 
-SCHEMA_TYPES = ("variable", "tag", "region", "generic")
+SCHEMA_TYPES = ("variable", "tag", "region", "generic", "meta")
```
phackstock, line 30: In conjunction with my comment about the meta schema I would leave it out here and just go with generic for now.
```suggestion
SCHEMA_TYPES = ("variable", "tag", "region", "generic")
```
I'll most likely remove all the schemas anyway in an upcoming PR.

nomenclature/codelist.py
```diff
@@ -624,3 +624,19 @@ def filter(self, hierarchy: str) -> "RegionCodeList":
...
+    ----------
+    name : str
+        Name of the MetaCodeList
+    mapping : dict
+        Dictionary of `MetaCode` objects
+
+    """
+
+    code_basis: ClassVar = MetaCode
+    validation_schema: ClassVar[str] = "meta"
```
phackstock, line 642: In conjunction with the push to remove all the schemas I'd leave this out and just use have `DataStructureDefinition` use the generic schema.
```suggestion
```

phackstock (changes requested):
Sorry for the schema confusion.
 @danielhuppmann please correct me if you're of a different opinion but I'd vote to remove all the schemas anyway soon. @GretchenSchowalter for some context, this was an early attempt at defining the file structures for codelists. The idea is that the yaml file is validated against the schema before it is even parsed into the codes. 
In reality this never added that much value since a faulty code list would just fail at a different point and the information given by the validation schema was not any better than the one from pydantic. 
I'd vote to remote the schema from this PR, after that it's good to be merged.

danielhuppmann (approved):
Nice work, thanks!

PR #247, opened 2023-05-10
Exclude defaults for yaml export

### LINKED ISSUE #231: Default-attributes are exported to yaml
When initializing a `VariableCode`, the attributes `region-aggregation` and `skip-region-aggregation` are automatically added and set to "False". When then exporting the CodeList to yaml, these attributes are written explicitly to the file, leading to very cumbersome, unwieldy files, see for example https://github.com/iiasa/ariadne-intern-workflow/pull/42

This is implemented in this line (and following)

https://github.com/IAMconsortium/nomenclature/blob/0b3dec191879faead7c0071d67a6d2bc5cb3035b/nomenclature/code.py#L142

I see two options:
1. don't automatically add these fields to the `VariableCode` if they are not defined in the source yaml/xlsx file
2. remove these attributes (if `False`) when exporting to yaml

phackstock:
closes #231.

Fixed the issue with a couple of additional changes

### Additional changes implemented

* `Code` and `CodeList` now feature `__eq__` methods which makes for easier comparisons for tests. For the equality operator of `Code` I explicitly exclude the `file` property as it does not influence functionality.
* `Code.file` is now a named attribute.
* Streamlined a few unit test accordingly.

### Open question

There is one case which is currently not covered very nicely which concerns the `create_yaml_from_xlsx` function.
Say we have an excel like this:

| Variable              | Description                                   | Unit  | Region-aggregation                              |
|-----------------------|-----------------------------------------------|-------|-------------------------------------------------|
| Primary Energy        | Total primary energy consumption              | EJ/yr | [{"Primary Energy (mean)": {"method": "mean"}}] |
| Primary Energy (mean) | Mean primary energy consumption               | EJ/yr |                                                 |
| Share\|Coal           | Share of coal in the total primary energy mix |       |                                                 |

this will be translated into the following yaml:

```yaml
- Primary Energy:
    description: Total primary energy consumption
    unit: EJ/yr
    region-aggregation:
      - Primary Energy (mean):
          method: mean
- Primary Energy (mean):
    description: Mean primary energy consumption
    unit: EJ/yr
    region-aggregation:
- Share|Coal:
    description: Share of coal in the total primary energy mix
    unit:
    region-aggregation:
```

The issue is that `Primary Energy (mean)` and `Share|Coal` both feature an empty _region-aggregation_ attribute since the column exists in the excel and the values for those two variables are `None`.
I don't think that there is a way from distinguishing this _unintentional_  `None` from the _intentional_ one for `Share|Coal` `unit`.
Functionality wise this does not change anything though.

I don't think that there's much we can do here, just wanted to point it out.


danielhuppmann (approved):
Nice, thank you!

PR #249, opened 2023-05-22
Meta validator

GretchenSchowalter:
Hi, Philip and Daniel! Apologies for the delay in getting this to you. This is my start for the MetaValidator class. My main question now is about the apply function. I know I need it in the MetaValidator class since it is inheriting from Processor, but would you like me to rename my validate_meta_indicators() function to apply()? Or should apply() be reserved for some other functionality? Also, the first step of the validate_meta_indicators() is to check if the meta_indicator is in the provided definitions/meta/<file_name>.yaml. Would you prefer that this actually check that the value of the meta indicator is not None? Looking forward to your feedback!


phackstock (commented):
A good first attempt @GretchenSchowalter :+1: 
To answer your question regarding `MetaValidator.apply()`, the apply function in this case should take a dataframe, validate all meta indicators and return the dataframe. 
The idea is that we have a number of `Processor` objects that we can chain since they all follow the `Processor.apply(df: IamDataFrame) -> IamDataFrame:` pattern.

Some more detailed comments below.

On a related note, probably way overkill for now, but there is a pandas validation library called pandera: https://towardsdatascience.com/validate-your-pandas-dataframe-with-pandera-2995910e564. 
Might be worth taking a look at.

GretchenSchowalter:
Hello! This is my next step with the `MetaValidator` class, but I am getting an error message for 2/3 of my tests and would appreciate any suggestions! 

For the first two tests the error message indicates "[nan] meta indicator value(s) in the string column are not allowed. Allowed values are ['foo', 'nan']". I am not sure if the issue is with how to represent nan in a .yaml file? The same error is happening in the second test as well because the code iterates through all the values in the columns before returning if any of the _meta_validators_ themselves are not allowed. I am doing it this way because I was thinking then the error message in the `apply()`  can be more comprehensive to list all the _meta_validator_ column names if there are more than one that don't match the .yaml file definitions (but I can also change this too).

Also, right now the `_if_allowed()` function returns a (True) boolean if all values are allowed and raises an error if not. Is this alright? Or should it return something else? I also am definitely open to renaming that method if there are any preferences.

tests/data/definitions3/meta/meta_indicators_more_data.yaml
```diff
@@ -0,0 +1,6 @@
+- exclude:
+    allowed_values: ['A', 'B']
+- number:
+    allowed_values: [1.0, 2.0, 3.0, 4.0]
+- string:
+    allowed_values: ['foo', nan]
```
danielhuppmann, line 6: "nan" and the Python object `None` often get confused by the yaml-parsing - I suggest to only work with actual strings for now.
```suggestion
    allowed_values: ['foo', 'bar']

```

nomenclature/processor/meta.py
```diff
@@ -0,0 +1,89 @@
...
+        for meta_indicator in df.meta.columns:
+            if meta_indicator not in self.meta_code_list.mapping:
+                unrecognized_meta_indicators.append(meta_indicator)
+            else:
+                values = df.meta[meta_indicator].values
+                allowed_values = self.meta_code_list.mapping[
+                    meta_indicator
+                ].allowed_values
+                self._values_allowed(values, allowed_values, meta_indicator)
+        if unrecognized_meta_indicators:
```
phackstock, line 83: I would swap the order of things, because right now you:
1. Check if there's any unrecognized meta indicators
2. Check if there's any values that are not allowed
  2.1. If there are values that are not allowed, you raise an error
3. If there are unrecognized meta indicators you raise an error

My suggestion would be to split the two steps more cleanly like so:

```python
# check for illegal columns
if unrecognized_meta_indicators := [
    meta_indicator
    for meta_indicator in df.meta.colums
    if meta_indicator not in self.meta_code_list.mapping
]:
    raise ValueError(
        f"{unrecognized_meta_indicators} is/are not recognized in the "
        f"meta definitions file. Allowed meta indicators are: "
        f"{list(self.meta_code_list.mapping)}"
    )
# check for illegal values
for meta_indicator in df.meta.columns:
    self._values_allowed(
        df.meta[meta_indicator].unique,
        self.meta_code_list.mapping[meta_indicator].allowed_values,
        meta_indicator,
    )
```

phackstock (changes requested):
A few minor things, should be good to go soon.

danielhuppmann (approved):
Nice work, thank you!

PR #250, opened 2023-05-23
Remove jsonschema

### LINKED ISSUE #192: Remove no longer needed validation schema for VariableCode
Now that the attributes of `VariableCode` are fixed in the class the validation schema is no longer needed.

phackstock:
closes #192 . 

This PR removes the previously used json/yaml schema files. 
This way we no longer have a two step validation process but rather a single one using pydantic.


nomenclature/processor/region.py
```diff
@@ -112,6 +112,19 @@ class RegionAggregationMapping(BaseModel):
...
 
+    @root_validator(pre=True)
+    def check_no_additional_attributes(cls, v):
+        if illegal_additional_attributes := [
+            input_attribute
+            for input_attribute in v.keys()
+            if input_attribute not in cls.__dict__["__fields__"]
+        ]:
+            raise ValueError(
+                "No additional attributes allowed, found: "
```
danielhuppmann, line 123:
```suggestion
                "Illegal attributes in 'RegionAggregationMapping': "
```

nomenclature/processor/region.py
```diff
@@ -112,6 +112,19 @@ class RegionAggregationMapping(BaseModel):
...
+                f"{illegal_additional_attributes} file {v['file']}"
```
danielhuppmann, line 124:
```suggestion
                f"{illegal_additional_attributes} (file {v['file']})"
```

nomenclature/processor/region.py
```diff
@@ -165,7 +178,7 @@ def check_native_or_common_regions(cls, values):
         ):
             raise ValueError(
                 "At least one of the two: 'native_regions', 'common_regions' must be"
-                f"given in {values['file']}"
+                f" given in {values['file']}"
```
danielhuppmann, line 181: Suggest to change message to "At least one of 'native_regions' and 'common_regions' must be provided in {value['file']}"

danielhuppmann (approved):
Thanks @phackstock, hope that this will be much easier to fix errors in the yaml files in the future. A few suggestions inline, then good to be merged!

danielhuppmann (approved):
Thanks @phackstock, hope that this will be much easier to fix errors in the yaml files in the future. A few suggestions inline, then good to be merged!

PR #252, opened 2023-05-24
Renamed 'countries' attribute to 'iso3_codes'in RegionCode class

### LINKED ISSUE #251: Use different name for ISO3-code attribute
Per a discussion at the ECEMF Consortium Meeting: the reserved word "countries" for the list of ISO3-codes is not ideal, because it is not obvious that only ISO3-codes are allowed.

Please change the package such that the reserved-word (including the validation) hinges on the term "iso3" instead.

GretchenSchowalter:
Hey, Daniel! This is a quick PR to address issue #251. I changed the attribute name from _countries_ to _iso3_codes_. Let me know if anything else needs to be changed!

closes #251

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -220,15 +220,15 @@ class RegionCode(Code):
     """
 
     hierarchy: str = None
-    countries: List[str] = None
+    iso3_codes: List[str] = None
 
-    @validator("countries")
+    @validator("iso3_codes")
     def check_iso3_codes(cls, v, values) -> List[str]:
         """Verifies that each ISO3 code is valid according to pycountry library."""
         invalid_iso3_codes: List[str] = []
-        for country in v:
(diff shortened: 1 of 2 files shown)


danielhuppmann (approved):
Looks good to me, thanks!

danielhuppmann:
@phackstock, are you ok with this change?

phackstock (approved):
@GretchenSchowalter, thanks for the quick work :+1: 
@danielhuppmann no major objections from my side. 
The only thing that I would change is the name of the attribute. `iso3_codes` sounds a little ambiguous to me. 
Something like `constituent_regions_in_iso3` would be longer but less ambiguous. 
No strong preference though, I'd totally be fine with `iso3_codes` as well.

danielhuppmann:
Thanks @phackstock - I think that your suggestion is unreadably long given that most values will be just 3 digits... But we can add some documentation later.

PR #256, opened 2023-06-05
Simplify docs folder structure, move to Black GitHub Actions

danielhuppmann:
Two small clean-ups because Stickler terminated its service and to keep consistency with the pyam package.


.github/workflows/build-docs.yml
```diff
@@ -10,20 +10,33 @@ on:
     branches: [ '**' ]
 
 jobs:
-  build:
+  pytest:
+    strategy:
+      matrix:
```
phackstock, line 15: Do we need a matrix for a single OS version with a single python version?

docs/make.bat
```diff
@@ -1,35 +1,37 @@
...
-
-%SPHINXBUILD% -M %1 %SOURCEDIR% %BUILDDIR% %SPHINXOPTS% %O%
-goto end
-
-:help
-%SPHINXBUILD% -M help %SOURCEDIR% %BUILDDIR% %SPHINXOPTS% %O%
-
-:end
-popd
+REM This file was generated with sphinx-quickstart version 3.5.0
```
phackstock, line 1: Do we actually need this file? From what I understand `.bat` files are for use on Windows and I don't think anyone is building the docs on Windows.

phackstock (commented):
Looks good, just two minor comments

.github/workflows/build-docs.yml
danielhuppmann: Simplified.

docs/make.bat
danielhuppmann: I'd keep it as long as there is no effort required to maintain it...

Issue #261, opened 2023-07-03
General options for a DataStructureDefinition


danielhuppmann:
While thinking about how to implement a list of countries in a project repository (ideally drawn directly from the **pycountry** package), I had the following idea:

In the definitions folder, there can be a config.yaml file with "general" specifications, e.g.

```yaml
- region:
    country: true
- time:
    datetime: true
...
```

Alternatively (or in addition), each `definition/<dimension>` folder could have such a config file.

The behavior would be as follows:

1. If region -> country == True, add all country names and ISO3-codes from the **pycountry** package to the **RegionCodeList**.
2. If time -> datetime == True, allow timeseries data with a `datetime` time domain. If false (default), fail the `validate()` method (see #54)

Any thoughts @phackstock?

phackstock:
I like the idea, smells like a pydantic specification to me ;)
Jokes aside though, I think having one file in `definitions/` called `general.yaml` is a good idea. As long as it does not get exceedingly big, I'd stay clear of having individual files for each dimension. It should be easier to have all the DataStructureDefinition configuration in a single spot.
One question regarding:

> If region -> country == True, add all country names and ISO3-codes from the pycountry package to the RegionCodeList.

Would that be an at-runtime thing or written into the yaml files?

danielhuppmann:
I would fill that at run-time - for the list of countries, this might still be possible (but duplicating) to create a yaml from the pycountry package. But for datetime-validation, writing a yaml would be pretty much pointless...

phackstock:
Right, makes sense to me.

dc-almeida:
Probably good to close, as both attributes have been implemented in `NomenclatureConfig`

phackstock:
Thanks for the housekeeping @dc-almeida, good to close from my side.

PR #262, opened 2023-07-10
Add a general-config feature to add all countries to RegionCodeList

danielhuppmann:
This PR adds a simple way to add all countries to a RegionCodeList directly from the **pycountry** package, using a "general-config" yaml file.

This implements the first idea of #261.


danielhuppmann:
When looking at the pycountry package in more detail for the SSP review process, I realized that it uses some country-names that are not consistent with the use in our community. So I guess that we will have to hard-code the country-names and ISO3-codes after all...

danielhuppmann:
After looking at the pycountry package some more, I suggest (and implemented) a mapping of "country-name-overrides" where we explicitly depart from pycountry standards. Seems like a preferable option than to duplicate an entire list of country names and ISO2/ISO3 codes...

The model-registration xlsx template has been updated accordingly.

nomenclature/config.py
```diff
@@ -0,0 +1,43 @@
...
+    """A class for configuration of a DataStructureDefinition
+
+    Attributes
+    ----------
+    config : dict
+        Attributes for configuring the DataStructureDefinition
+
+    """
+
+    config: Dict[str, Dict] = {}
```
phackstock, line 18: I would make the dimensions explicit here: 
```suggestion
    region: Dict[str, Any] = {}
```
and add more as we go along. 
Or to completely satisfy my love for pydantic:
```python
    region: Optional[RegionCodeListConfig]
...
class RegionCodeListConfig(BaseModel):
    country: Optional[bool]
```
This is all the `RegionCodeListConfig` class would need to be for now. Again, can be extended for new features. 
The advantage is that you get type checking for the `config.yaml` file for free and don't have to do it "manually" like in `codelist.py` line 572, 581 and 582.

nomenclature/config.py
```diff
@@ -0,0 +1,43 @@
...
+        file : str
+            File name
+        dimensions : list
+            List of valid dimensions
+
+        """
+        with open(path / file, "r", encoding="utf-8") as stream:
+            config = yaml.safe_load(stream)
+
+            if invalid_config := [i for i in config if i not in dimensions]:
```
phackstock, line 40: If you explicitly list all the dimensions in `DataStructureConfig` and set the config to not allow additional attributes you probably don't need the `dimensions` input parameter and this check.
I would also say that we can remove the `dimensions` input parameter from the `DataStructureDefinition` constructor. Reason being that I haven't come across a use case where we have a code list for a particular dimension but then don't want to validate that dimension. 
Maybe there is a use case though is which case this check is necessary. In any case this would be a new issue anyway.

phackstock (commented):
Looks good to me. 
The only comment I'd have is on the use of the `DataStructureConfig`. 
Right now its just a wrapper around a dictionary. I think you could save a lot of manual checking, for now and for the future, if you completely defined the structure of the `DataStructureConfig` and its dimension specific settings. 
My inline comments are regarding that.

nomenclature/config.py
danielhuppmann: We have such a use-case in openENTRANCE and ARIADNE, where we use the same workflow repo for different database instances - in each case, one instance checks the "scenario" dimension and the other(s) ignore it...

phackstock: Ok valid. I'd still using an iterator other than `i` so that it's easier to comprehend what you're iterating over. Maybe `dimension` would be better.

danielhuppmann:
Implemented the suggested changes, @phackstock, towards nested pydantic classes. And I also added some basic documentation - I'm planning to add more documentation in a follow-up PR, then we can do a release 0.11

phackstock (approved):
Looks good to me. 
The pydantic config classes make me very happy, thank you :)
One question that came to my mind was if it would be easier to just work with ISO codes directly as the region names. Might be easier for modeling teams to align their outputs.

danielhuppmann:
> One question that came to my mind was if it would be easier to just work with ISO codes directly as the region names. Might be easier for modeling teams to align their outputs.

Right, @willu47 made a similar point here: https://github.com/OSeMOSYS/osemosys2iamc/issues/33#issuecomment-1632033971

But I don't really like the idea for several reasons:
- ISO3 codes are not really user-friendly - it's difficult to get a quick intuition in a tool like Artelys' FutureSight or the Scenario Explorer, so we would probably invent some "conversion tool" again
- ISO3 codes are not fully universal either, e.g., Kosovo doesn't have one
- Using ISO3 codes in native-country naming also has issues, for example REMIND 3.2 uses "UKI" internally for "United Kingom and Ireland" - this can be very confusing when users try an ISO3-mapping on that... (see [here](https://github.com/iiasa/eu-climate-advisory-board-workflow/blob/bb2e63a145ce9e29edcbc3072799dd63cbfde9d2/mappings/REMIND_3.2.yml#L20)).

So I think it's better to nudge modelers to use intelligible names instead of abbreviations.

Nonetheless, we should add a simple ISO2/ISO3 mapping tool to the shortened country names added in this PR (and write documentation for it).

phackstock:
> But I don't really like the idea for several reasons:

Ah right, good points, it's never as simple as you think ...

> Nonetheless, we should add a simple ISO2/ISO3 mapping tool to the shortened country names added in this PR (and write documentation for it).

Good point, that would be surely appreciated. In the context of NGFS, the users were specifically asking for the country level data to be reported in ISO3 so that would come in handy there.

PR #265, opened 2023-07-20
Include definitions from GitHub repo

phackstock:
First version of the external repo fetching feature.

### How to use
In the `definitions` folder of a project place a file called definitions with the following content:

```yaml
region:
  repository: https://github.com/IAMconsortium/common-definitions.git/
  # hash: optionally specify a commit hash of the repository
  # release: optionally specify a release of the repository
```
the `hash` and `release` attributes are both optional. 
If neither is given, the last version of the main branch is used. 
They are also mutually exclusive. If both are given an error is raised.

### Implementation details

The repo fetching logic is implemented as follows:
1. If the repo is not present, clone it
2. Perform a hard reset to get rid of any changes
3. Check out the desired revision

### DIFF
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Dict, Optional
+from typing import Optional
 from pydantic import BaseModel, root_validator
 
 import yaml
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -1,12 +1,44 @@
 from pathlib import Path
 from typing import Dict, Optional
-from pydantic import BaseModel
+from pydantic import BaseModel, root_validator
 
 import yaml
+from git import Repo
 
 
 class CodeListConfig(BaseModel):
-    repository: Optional[Path]
+    repository: Optional[str]
+    hash: Optional[str]
+    release: Optional[str]
+
+    @root_validator()
+    def check_hash_and_release(cls, v):
+        if v.get("hash") and v.get("release"):
+            raise ValueError("Either 'hash' or 'release' can be provided, not both.")
+        return v
+
+    @property
+    def revision(self):
+        return self.hash or self.release or "main"
+
+    @property
+    def repository_name(self):
+        return self.repository.split("/")[-1].split(".")[0]
+
+    def fetch_repo(self, to_path="./common-definitions"):
+        to_path = to_path if isinstance(to_path, Path) else Path(to_path)
+
+        if not to_path.is_dir():
+            repo = Repo.clone_from(self.repository, to_path)
+        else:
+            repo = Repo(to_path)
+            repo.remotes.origin.fetch()
+        repo.git.reset("--hard")
+        repo.git.checkout(self.revision)
+        repo.git.reset("--hard")
+        repo.git.clean("-xdf")
+        if self.revision == "main":
+            repo.remotes.origin.pull()
 
 
 class RegionCodeListConfig(CodeListConfig):
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -9,7 +9,7 @@ from git import Repo
 class CodeListConfig(BaseModel):
     dimension: str
     repository: Optional[str]
-    repository_definition_path: Optional[str]
+    repository_dimension_path: Optional[Path]
 
     @root_validator()
     def set_repository_definition_path(cls, v):
@@ -29,7 +29,7 @@ class Repository(BaseModel):
     url: str
     hash: Optional[str]
     release: Optional[str]
-    path: Optional[Path]
+    local_path: Optional[Path]  # defined via the `repository` name in the configuration
 
     @root_validator()
     def check_hash_and_release(cls, v):
@@ -40,7 +40,7 @@ class Repository(BaseModel):
     @validator("path")
     def check_path_empty(cls, v):
         if v is not None:
-            raise ValueError("path must not be set as part of the config")
+            raise ValueError("The `path` must not be set as part of the config.")
         return v
 
     @property
@@ -95,8 +95,8 @@ class DataStructureConfig(BaseModel):
             ):
                 raise ValueError(
                     (
-                        f"Unknown repository {values.get(dimension).repository} in"
-                        " region.repository."
+                        f"Unknown repository '{values.get(dimension).repository}' in"
+                        f" {dimension}.repository."
                     )
                 )
         return values
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -23,7 +23,7 @@ class CodeListConfig(BaseModel):
 
     @property
     def repository_name(self):
-        return self.repository.split("/")[-1].split(".")[0]
+        return self.repository.split("/")[-2].split(".")[0]
 
     def fetch_repo(self, to_path="./common-definitions"):
         to_path = to_path if isinstance(to_path, Path) else Path(to_path)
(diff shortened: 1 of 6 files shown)


tests/data/general-config-definitions/config.yaml
```diff
@@ -1,3 +1,3 @@
 region:
```
danielhuppmann, line 1: We don't want to clone the same repo for variables, regions, etc.
```suggestion
repository:
  common-defintions:
    git: git@github.com:IAMconsortium/common-definitions.git
    hash:
    release:
region:
    repository:
```

phackstock: > We don't want to clone the same repo for variables, regions, etc.

Do you mean cloning the same repo multiple times? 
If so, this wouldn't happen anyway since the code checks if the repo is present. If it is it is just updated if need be. 

Although I do have to say it might be nice to not have to define the repo url multiple times.

> Plus the "dictionary" of a repo-name to the git-attributes can help to support multiple external repos in the future.

This should already work.

```yaml
region:
  repository: https://github.com/IAMconsortium/common-definitions.git/
  # hash: optionally specify a commit hash of the repository
  # release: optionally specify a release of the repository
variable:
  # use a different repo for the variables
  repository: https://github.com/IAMconsortium/common-definitions.git/
  # hash: optionally specify a commit hash of the repository
  # release: optionally specify a release of the repository
```
unless you mean multiple external repos for a single dimension. In this case the code would need to be modified but I'd say we cross that bridge when we get there.

nomenclature/config.py
```diff
@@ -1,12 +1,44 @@
...
+
+    @property
+    def revision(self):
+        return self.hash or self.release or "main"
+
+    @property
+    def repository_name(self):
+        return self.repository.split("/")[-2].split(".")[0]
+
+    def fetch_repo(self, to_path="./common-definitions"):
```
danielhuppmann, line 28: Why is the *to_path* hard-coded?

danielhuppmann:
The implementation looks good, thanks! But I still think that the configuration would be more elegant and future-proof if we do:

```yaml
repository:
  common-definitions:
    git: git@github.com:IAMconsortium/common-definitions.git
    hash:  # optional
    release:  # optional
region:
    path:  # optional, defaults to `definitions / regions`
variable:
    path:  # optional, defaults to `definitions / variable`
```

This would make the *repository_name* explicit and customizable.

And the `fetch_repo()` should happen when initializing the **DataStructureConfig** (if a *repository* is given), so that the **CodeListConfig** only has to know which path (`"repository" / <repository_name> / <path>`) to parse in addition to its "own" (local) folder.

phackstock:
> The implementation looks good, thanks! But I still think that the configuration would be more elegant and future-proof if we do:

Agreed, I'll change it that way.

nomenclature/config.py
phackstock: That was a left over from initial development, it's not removed.

phackstock:
# Update

@danielhuppmann I have updated `DataStructureConfig` and all associated classes according to your suggestions. 
The configuration file looks now like this:
```yaml
repository:
  common-definitions:
    url: https://github.com/IAMconsortium/common-definitions
region:
  country: true
variable:
  repository_definition_path: definitions / variable
```

I changed your `git` attribute to `url` but happy to change it back. The repository url needs to needs to end with _.git/_ otherwise the _gitpython_ library fails to fetch the repository. 
I've also changed the `path` attribute in the dimensions, _region_ and _variable_, in this example, to `repository_definition_path`. I understand that it's a bit clunky and will probably be changed but I was confused by just the name `path`. Reason being is that it could also refer to the path of the local repositories codelist folder. Happy to suggestions to rename. Or even leave out for now as I think we can enforce the structure of an external repository to be "repo/definitions/_dimension_".

I've also, according to your suggestion moved the repository fetching into the constructor of `DataStructureConfiguration`, much cleaner this way. Now all repos are automatically fetched upon initialization.

In addition I've extended the use of external repos from just region to any dimension. 
The test that I've modified to cover this also tests for the case that there is no local codelist folder. In case of the test, there's no variable folder. Works perfectly.

tests/test_definition.py
```diff
@@ -43,19 +44,26 @@ def test_empty_codelist_raises():
...
+        # imported from https://github.com/IAMconsortium/common-definitions repo
+        assert "World" in obs.region
+        # added via general-config definitions
+        assert "Austria" in obs.region
+        # added via general-config definitions renamed from pycountry name
+        assert "Bolivia" in obs.region
+        # added via general-config definitions in addition to pycountry.countries
+        assert "Kosovo" in obs.region
+
+        assert "Primary Energy" in obs.variable
```
danielhuppmann, line 61:
```suggestion
        # imported from https://github.com/IAMconsortium/common-definitions repo
        assert "Primary Energy" in obs.variable
```

nomenclature/config.py
```diff
@@ -1,18 +1,69 @@
...
 
 import yaml
+from git import Repo
 
 
 class CodeListConfig(BaseModel):
-    repository: Optional[Path]
+    dimension: str
+    repository: Optional[str]
+    repository_definition_path: Optional[str]
```
danielhuppmann, line 12: This does not have the "definitions" path, but the dimension within a definitions folder...

Also, should this be a `Path` object?
```suggestion
    repository_dimension_path: Optional[Path]
```

nomenclature/config.py
```diff
@@ -1,18 +1,69 @@
...
 
 class RegionCodeListConfig(CodeListConfig):
     country: Optional[bool]
 
 
+class Repository(BaseModel):
+    url: str
+    hash: Optional[str]
+    release: Optional[str]
+    path: Optional[Path]
```
danielhuppmann, line 32: Suggest to rename to clarify that this is the local (cloned) path...
```suggestion
    local_path: Optional[Path]  # defined via the `repository` name in the configuration
```

nomenclature/config.py
```diff
@@ -1,18 +1,69 @@
...
+    @root_validator()
+    def check_hash_and_release(cls, v):
+        if v.get("hash") and v.get("release"):
+            raise ValueError("Either 'hash' or 'release' can be provided, not both.")
+        return v
+
+    @validator("path")
+    def check_path_empty(cls, v):
+        if v is not None:
+            raise ValueError("path must not be set as part of the config")
```
danielhuppmann, line 43:
```suggestion
            raise ValueError("The `path` must not be set as part of the config.")
```

nomenclature/config.py
```diff
@@ -23,7 +74,32 @@ class DataStructureConfig(BaseModel):
...
+            if (
+                values.get("repository")
+                and values.get(dimension)
+                and values.get(dimension).repository
+                and values.get(dimension).repository not in values.get("repository")
+            ):
+                raise ValueError(
+                    (
+                        f"Unknown repository {values.get(dimension).repository} in"
+                        " region.repository."
```
danielhuppmann, line 99:
```suggestion
                        f" {dimension}.repository."
```

nomenclature/config.py
```diff
@@ -23,7 +74,32 @@ class DataStructureConfig(BaseModel):
...
+        for dimension in ("region", "variable"):
+            if (
+                values.get("repository")
+                and values.get(dimension)
+                and values.get(dimension).repository
+                and values.get(dimension).repository not in values.get("repository")
+            ):
+                raise ValueError(
+                    (
+                        f"Unknown repository {values.get(dimension).repository} in"
```
danielhuppmann, line 98:
```suggestion
                        f"Unknown repository '{values.get(dimension).repository}' in"
```

danielhuppmann (approved):
Very nice! A few suggestions for alternative keyword names, but good to be merged either way...

PR #266, opened 2023-07-24
Improve the docs of the countries feature

danielhuppmann:
This PR adds a link to the model-registration Excel to the countries-module docs for users who want an easy (non-Python) way to get the ist of countries.


danielhuppmann:
> Just checked quickly and Kosovo is not in the Excel file.

It was at the end of the list, not in the alphabetical order - changed it to avoid confusion

PR #268, opened 2023-07-25
Import mappings from external repo

### LINKED ISSUE #267: Enable using model mappings from external repository
After #265, the next step is to enable loading model mappings from an external repository.

phackstock:
closes #267.

### Changes 

* There is now a new configuration class called `NomenclatureConfig`. Example:
```yaml
repositories:
  common-definitions:
definitions:
  region:
  variable:
mappings:
```
* In analogy to the readthedocs configuration this configuration **must be** in a file called **.nomenclature.yaml** located in the **root** of the repository.

### Open questions

* I'm open for suggestions on the name of the configuration class and the configuration file. I just had to put something there for now.
* With the option now to instantiate `CodeList` and `RegionProcessor` objects entirely from and external repository, the `from_directory` methods of both classes no longer accurately reflect their functions. 
* Additionally, the folders for definitions is currently required to exist, even if it's empty.

Maybe the last two points are better addressed in a follow-up PR though.


nomenclature/codelist.py
```diff
@@ -191,22 +191,26 @@ def from_directory(
         instance of cls (:class:`CodeList` if not inherited)
 
         """
-        code_list = cls._parse_codelist_dir(path, file_glob_pattern)
 
-        if config is not None:
+        if (
```
danielhuppmann, line 195: Maybe a try-except is the more pythonic approach here?

nomenclature/codelist.py
```diff
@@ -191,22 +191,26 @@ def from_directory(
...
-                )
+            codelistconfig = getattr(config.definitions, dimension)
+            repo_path = (
+                config.repositories[codelistconfig.repository].local_path
+                / codelistconfig.repository_dimension_path
+            )
+            code_list = cls._parse_codelist_dir(
+                repo_path,
+                file_glob_pattern,
+            ) + cls._parse_codelist_dir(path, file_glob_pattern)
```
danielhuppmann, line 211: Even if you keep the if-statement, I'd pull out that line because it's duplicated in both *if* and *else*...

nomenclature/definition.py
```diff
@@ -44,23 +44,18 @@ def __init__(self, path, dimensions=None):
         if not path.is_dir():
             raise NotADirectoryError(f"Definitions directory not found: {path}")
 
-        if (path / "config.yaml").exists():
-            self.config = DataStructureConfig.from_file(
-                path=path,
-                file="config.yaml",
-            )
+        if (file := path.parent / ".nomenclature.yaml").exists():
```
danielhuppmann, line 47: I like the idea of the file name, but I worry that Windows & Mac will hide the file it it starts with `.`... And in contrast to RTD-config, this is quite important for our users.
```suggestion
        if (file := path.parent / "nomenclature.yaml").exists():

```

nomenclature/processor/region.py
```diff
@@ -349,7 +350,22 @@ def from_directory(cls, path: DirectoryPath, dsd: DataStructureDefinition):
         """
         mapping_dict: Dict[str, RegionAggregationMapping] = {}
         errors: List[ErrorWrapper] = []
-        for file in (f for f in path.glob("**/*") if f.suffix in {".yaml", ".yml"}):
+
+        mapping_files = [f for f in path.glob("**/*") if f.suffix in {".yaml", ".yml"}]
+
+        if (file := path.parent / ".nomenclature.yaml").exists():
```
danielhuppmann, line 356: Not ideal if the nomenclature-config file is parsed several times - but does it mean that the cloned repo is also pulled several times?

nomenclature/config.py
```diff
@@ -101,24 +98,60 @@ def check_repository_consistency(cls, values):
...
+    mappings: Optional[RegionMappingConfig]
+
+    @root_validator
+    def check_definitions_repository(cls, v):
+        definitions_repos = v.get("definitions").repos if v.get("definitions") else {}
+        mapping_repos = (
+            {"mappings": v.get("mappings").repository} if v.get("mappings") else {}
+        )
+        repos = {**definitions_repos, **mapping_repos}
+        if repos and not v.get("repositories"):
```
danielhuppmann, line 126: I guess this error is overly pydantic - the next check will catch it if the repository is not defined, right?

danielhuppmann (commented):
Looks good overall, a few minor comments and suggestions inline.

nomenclature/codelist.py
phackstock: Ah yes of course, thanks. I couldn't think of that one...

nomenclature/definition.py
phackstock: Fair point, I'll remove the leading dot.

nomenclature/config.py
phackstock: It would produce a different kind of error. 
If `repositories` is not defined the check below would resolve as:
```python
if repository not in None
```
which would result in this error:
```console
Traceback (most recent call last):
TypeError: argument of type 'NoneType' is not iterable
```

nomenclature/processor/region.py
phackstock: It's not ideal that's true. One way around it would be to separately create a `NomenclatureConfig` object once and pass it to `DataStructureDefinition` and `RegionProcessor`. 
I could implement that. It's just something that would need to be added to every workflow though ...

nomenclature/config.py
danielhuppmann: Ok, got it!

nomenclature/processor/region.py
phackstock: True, good idea, that should work.

nomenclature/codelist.py
phackstock: Good point, done

danielhuppmann (approved):

PR #269, opened 2023-07-25
Rename US/British Virgin Islands in `countries` module

danielhuppmann:
Working with the SSP-review and common-definitions, I noticed that the ISO 3166-1 names for the US and British Virgin Islands are not elegant to work with in a yaml format due to comma or brackets. For simplicity, this PR adds these two countries to the replace-mapping.


phackstock (approved):
Good to merge

PR #272, opened 2023-08-07
Refactor MetaValidator tests to not use `exclude` indicator 

danielhuppmann:
This PR implements the change that the pyam-meta-attribute does not (any more) have an indicator `exclude`.

This PR is not compatible with the latest release-version of pyam, but I think that this is not a concern because:
- the pypi-release-versions are compatible
- the latest versions on GitHub are compatible
- we add a dependency on pyam>2.0 in nomenclature after the next release

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -6,8 +6,9 @@ name: nightly
 
 on:
   schedule:
-    # runs once per week on Sunday at 3am UCT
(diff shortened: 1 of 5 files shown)


danielhuppmann:
Also, the nightly tests will be run on Sunday so as to not spoil my weekend...

phackstock (approved):
Thanks for the quick fix @danielhuppmann. 
Good to merge from my side.

PR #273, opened 2023-08-08
Extend the `Countries` docstring

danielhuppmann:
Starting from a simple typo-correction, this PR adds a few lines of explanation to the **Countries** docstring.

Note that a more detailed explanation is already given via the module-page on readthedocs, see https://nomenclature-iamc.readthedocs.io/en/stable/api/countries.html

### DIFF
--- a/nomenclature/countries.py
+++ b/nomenclature/countries.py
@@ -42,7 +42,6 @@ ALTERNATIVE_ALPHA2_CODES = {
 
 
 class Countries(pycountry.ExistingCountries):
-
     """List of countries based on simplified ISO 3166 database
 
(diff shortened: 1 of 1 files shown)


nomenclature/countries.py
```diff
@@ -42,7 +42,13 @@
 
 
 class Countries(pycountry.ExistingCountries):
-    """List of countries based on ISO 3166 database with simplications"""
+
```
phackstock, line 45: I think you can remove that line
```suggestion
```

phackstock (commented):
One tiny comment other than that good to merge.

PR #276, opened 2023-08-08
Check if path is contained in cwd

### LINKED ISSUE #274: Get relative path fails if file is not part of current working directory
This occurred while using the climate-processor's `RequiredDataValidator`. Here the required data file was located outside of the current working directory.

phackstock:
Closes #274


danielhuppmann (approved):
Thanks, looks good!

PR #279, opened 2023-09-04
Add RegionProcessor.revert()

### LINKED ISSUE #111: Implement processor to "undo" region processing
It can be useful/required to "undo" or reverse the renaming done as part of the region-processing, so taking the model-native-in-ensemble-region-names and rename to model-native-as-used-originally-region-names.

My hunch is that this would be better implemented as a new type of processor (instead of an argument of `RegionProcessor.apply()` or a new method there), so that we can easily call a list of processors using

```python
process(IamDataFrame, DataStructureDefinition, processor=[ReverseRegionProcessor, SomeOtherProcessor])
```

# Question to be discussed

Whether to drop or keep common regions as part of the operation, and (if that would be optional) how to configure/specify this...

phackstock:
Closes #111.
I needed this for some climate processing for the ENGAGE project so I figured I might as well add it.
I made the choice to drop common regions, not sure if it's the best one but we can always change it later if need be.

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -603,6 +603,18 @@ class RegionProcessor(Processor):
         # cast processed timeseries data and meta indicators to IamDataFrame
         return IamDataFrame(_data, meta=model_df.meta), difference
 
+    def revert(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
+        model_dfs = []
+        for model in df.model:
+            model_df = df.filter(model=model)
+            if mapping := self.mappings.get(model):
+                # remove common regions
+                model_df = model_df.filter(
+                    region=mapping.common_region_names, keep=False
+                ).rename(region=mapping.reverse_rename_mapping)
+            model_dfs.append(model_df)
+        return pyam.concat(model_dfs)
+
 
 def _aggregate_region(df, var, *regions, **kwargs):
     """Perform region aggregation with kwargs catching inconsistent-index errors"""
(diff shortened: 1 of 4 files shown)


nomenclature/processor/region.py
```diff
@@ -667,3 +671,17 @@ def _check_exclude_region_overlap(values: Dict, region_type: str) -> Dict:
...
+class ReverseRegionProcessor(RegionProcessor):
+    def apply(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
+        model_dfs = []
+        for model in df.model:
+            model_df = df.filter(model=model)
+            if mapping := self.mappings.get(model):
+                # remove common regions
+                model_df = model_df.filter(
+                    region=mapping.common_region_names, keep=False
+                ).rename(region=mapping.reverse_rename_mapping)
```
danielhuppmann, line 685: Maybe better readability?
```suggestion
                model_df = (
                    model_df
                    .filter(region=mapping.common_region_names, keep=False)
                    .rename(region=mapping.reverse_rename_mapping)
                )
```

danielhuppmann (commented):
Looks good in principle, but is there a particular reason for using a subclass instead of a method like `RegionProcessor.revert()`?

phackstock:
Updated to `RegionProcessor.revert()`

nomenclature/processor/region.py
```diff
@@ -599,6 +603,18 @@ def _apply_region_processing(
...
 
+    def revert(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
+        model_dfs = []
+        for model in df.model:
+            model_df = df.filter(model=model)
+            if mapping := self.mappings.get(model):
+                # remove common regions
+                model_df = model_df.filter(
+                    region=mapping.common_region_names, keep=False
+                ).rename(region=mapping.reverse_rename_mapping)
```
danielhuppmann, line 614: Deja-vu
```suggestion
                model_df = (
                    model_df
                    .filter(region=mapping.common_region_names, keep=False)
                    .rename(region=mapping.reverse_rename_mapping)
                )
```

phackstock: Ah yes just double checked black does not like that one and if I change it, it reverts it back.

danielhuppmann: 🤯

PR #280, opened 2023-09-04
Update supported python versions

phackstock:
As I created #279 I noticed that a number of tests were failing. The ones for python versions 3.8 and 3.9 are failing because the latest version of pyam does not longer support them.
So I updated nomenclature accordingly. 

Note: The tests are currently failing because of the pyam update for the RequiredDataValidator. I'll update that next.


danielhuppmann (approved):
Looks good, thanks!

PR #284, opened 2023-09-25
Raise error for unknown region

### LINKED ISSUE #283: Invalid regions do not raise an error as part of RegionProcessor
The method `RegionProcessor.apply()` does not raise an error when the IamDataFrame includes (or aggregates to) a region that is not listed in the region definitions.

This is because the `validate_items()` method only returns the invalid items but does not itself raise an error - so the following line does not have any effect.
https://github.com/IAMconsortium/nomenclature/blob/c2769559263617c6b0a066e958ca1f1627080b55/nomenclature/processor/region.py#L453

phackstock:
Closes #283.

This PR adds an error that is raised is case `RegionProcessor` encounters an unknown region.
The reporting is done in the same style as `DataStructureDefinition.validate` does it.

As a side note, I think this was a bit for a corner case, unknown regions would only slip by if the following conditions are met:

* A `RegionProcessor` is used (otherwise `process` validates the regions)
* The unknown regions are attached to a model for which no model mapping is defined. This is because as part of the region processing, the mapping is checked for consistency _and_ and error is raised if input data contain regions that are not mentioned in `native_regions`, `common_regions` or `exclude_regions`

In any case, the bug is fixed now.

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -26,6 +26,7 @@ from nomenclature.error.region import (
 )
 from nomenclature.processor import Processor
 from nomenclature.processor.utils import get_relative_path
+from nomenclature.validation import log_error
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


danielhuppmann (approved):
Nice, thank you! Anfror what it’s worth, I did run into exactlythis corner case… 😜

phackstock:
> Nice, thank you! And for what it’s worth, I did run into exactlythis corner case… 😜

Ah ok, I see, I'll see to set it live soon.

PR #285, opened 2023-10-09
Prototype for a model-registration parser

danielhuppmann:
A discussion for a model-registration reminded me of this prototype for a model-registration parser utility, so this PR adds the initial implementation for future use and improvement.

My suggestion is to to a quick review, then merge, and extend features, etc. as the need arises.


nomenclature/__init__.py
```diff
@@ -52,3 +54,44 @@ def create_yaml_from_xlsx(source, target, sheet_name, col, attrs=None):
...
+
+    common_region_groups = [r for r in regions.columns if r not in [native, rename]]
+
+    # write region definitions file
+    with open(f"definitions/region/model_native_regions/{file_name}.yaml", "w") as file:
+        file.write(f"- {model}:{sep}" + sep.join(regions[rename].values))
+
+    # write mappings file
+    with open(f"mappings/{file_name}.yaml", "w") as file:
+        file.write(f"model:{sep}{model}\n")
```
phackstock, line 84: Similar idea as above, I would create a dictionary that represents the model mapping and then dump directly to yaml.
Alternatively, we could also use the `RegionAggregationMapping` class directly, although that might be a bit overkill.
```python
model_mapping = {"model": model}
model_mapping["native_regions"]=[{} for _, row in regions.iterrows()]
model_mapping["common_regions"]=...
```

phackstock (commented):
I think there's a few updates necessary to make this functional and nice. 
If you want you can go ahead with the merge as is but I'd suggest that I'll update your branch for clean up.

phackstock:
Finally got around to finishing this one.

phackstock:
I modified `RegionAggregationMapping.from_file` so that it chooses which parser to use depending on whether we read yaml or excel. In addition I added a to_yaml method so that you can write the class to a file.
With those two components `parse_model_registration` is nice and compact :)
@danielhuppmann, since this is your PR I think I can't chose you as a reviewer, is this ok or should I open it again under my name so that you can review it properly?

danielhuppmann:
Thanks for continuing to work on my initial prototype, looks very nice!

PR #288, opened 2023-10-11
Fix for general config failing without definitions

### LINKED ISSUE #286: General-config fails if folder is not present
Adding a dimension via the "nomenclature.yaml" general-config file fails if no (possibly empty) folder with that dimension name exists in the definitions directory.

phackstock:
Closes #286. 

I slightly modified the check if the definitions folder is present. 
Now it is called after reading the general config and only fails if there is no definitions folder **and** no external repos.

### DIFF
--- a/tests/test_definition.py
+++ b/tests/test_definition.py
@@ -41,16 +41,15 @@ def test_empty_codelist_raises():
         DataStructureDefinition(TEST_DATA_DIR / "simple_codelist")
 
 
-@pytest.mark.parametrize("external_only", [True, False])
-def test_definition_from_general_config(external_only):
-    workflow_folder = "general-config-only" if external_only else "general-config"
+@pytest.mark.parametrize("workflow_folder", ["general-config-only", "general-config"])
+def test_definition_from_general_config(workflow_folder):
     obs = DataStructureDefinition(
         TEST_DATA_DIR / workflow_folder / "definitions",
         dimensions=["region", "variable"],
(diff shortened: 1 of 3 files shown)


tests/test_definition.py
```diff
@@ -64,6 +64,27 @@ def test_definition_from_general_config():
         clean_up_external_repos(obs.config.repositories)
 
 
+def test_only_definition_from_general_config():
```
danielhuppmann, line 67: Maybe merge with the test above (as two alternative cases) to reduce duplication?

phackstock: Good point, done.

tests/test_definition.py
```diff
@@ -41,14 +41,17 @@ def test_empty_codelist_raises():
...
+    workflow_folder = "general-config-only" if external_only else "general-config"
     obs = DataStructureDefinition(
-        TEST_DATA_DIR / "general-config" / "definitions",
+        TEST_DATA_DIR / workflow_folder / "definitions",
         dimensions=["region", "variable"],
     )
     try:
         # explicitly defined in `general-config-definitions/region/regions.yaml`
-        assert "Region A" in obs.region
+        if not external_only:
```
danielhuppmann, line 53:
```suggestion
        if workflow == "general-config":
```

danielhuppmann (approved):
Looks good, just one minor clean-up suggestion...

PR #292, opened 2023-11-27
Add link from model-mapping page to region-aggregation-attributes

danielhuppmann:
Per a comment by @lvreedenburgh, this PR adds a link from the model-mapping user guide page to the section where there region-aggregation-attributes are explained.


phackstock (approved):
Thanks for the useful addition.

PR #293, opened 2023-12-12
Use poetry

phackstock:
Switching to poetry for development of nomenclature.


DEVELOPING.rst
```diff
@@ -1,3 +1,43 @@
...
+# (using "--with dev,docs,server" if dev and docs dependencies should be installed as well)
+poetry install --with dev,docs
+
+# Activate virtual environment
+poetry shell
+
+# Copy the template environment configuration
+cp template.env .env
+
+# Add a test platform
```
danielhuppmann, line 34: I guess that this is not relevant for nomenclature

danielhuppmann (approved):
One minor case of copying more than you should, otherwise good to go!

DEVELOPING.rst
phackstock: Ah ups, good catch

PR #297, opened 2023-12-19
Update to pydantic v2

### LINKED ISSUE #214: Replace usage of pydantic ValidationErrors outside of actual validators
When trying to emulate the very convenient "catch all errors and raise them afterwards" behavior of pydantic I (ab)used the `pydantic.ValidationError` class. It is better to replace this with the following pattern: 

```python
    errors: List[str] = []
    for file in (requiredDatadir).iterdir():
        try:
            RequiredDataValidator.from_file(file).validate_with_definition(dsd)
        except SomeError as e:
            errors.append(str(e))
    if errors:
        all_errors = "\n".join(errors)
        raise ValueError(f"Found error(s) in required data files: {all_errors}")
```

and leave the pydantic.ValidationErrors in the actual validators.

### LINKED ISSUE #291: Upgrade to use latest pydantic
**ixmp4** now uses pydantic > 2.0, which is incompatible with the requirements of this package. This should be synced so that the next **pyam** release does not create problems for the dependency chain nomenclature -> pyam -> ixmp4.

phackstock:
Closes #291, closes #214.

# Changes

Pydantic v2 introduced a lot of API changes and deprecated a lot of classes and functions so quite a few changes were necessary.

## CustomErrors

Previously we made use of a lot of custom errors. Pydantic v2 no longer allows this. There is a class `PydanticCustomError` but it cannot be subclassed. 
After consideration I came to the conclusion that it's still useful to make use of these custom errors since this gives us finer control over the formatting of the error messages. 
As a fix I ripped out the error folder and replaced it with a single `error.py` module. In this module there's now a dictionary called `pydantic_custom_error_config` that defines the values for our various custom errors. For each error this is a short name and the error message. 
Custom errors can now be raised like so:
```python
from pydantic_core import PydanticCustomError
from nomenclature.error import custom_pydantic_errors
...
raise PydanticCustomError(*custom_pydantic_errors.MyCustomError, {"value1": value1, "value2": value2})
```
where `value1` and `value2` are to be interpolated as part of the error message.

## Error collection

Previously I had (ab)-used pydantic achieve the behavior of collecting a number of errors and then raising them all at once. This is no longer possible since pydantic 2 removed the class I used to do that. It was also never really the cleanest way to implement this mechanic since pydantic errors should be reserved for validation at object creation.
I implemented a class called `ErrorCollector` which collects errors in a list to be raised at a later point in time.
This is how you use it:
```python
    errors = ErrorCollector()
    for file in required_data_dir.iterdir():
        try:
            RequiredDataValidator.from_file(file).validate_with_definition(dsd)
        except ValueError as error:
            errors.append(error)
    if errors:
        raise ValueError(f"Found error(s) in required data files:\n{errors}")
```
Output wise it is currently not ideal but I'd say this PR is big enough and I'll address this in a follow up PR. 

## Each item

Pydantic v2 changed the rather straightforward `each_item` option for validators. Previously this option would tell pydantic to apply a validator to each item of a list of dictionary. 
Now this has to be done via the `Annotated` keyword in the variable type definition. 
I don't really like it in terms of readability but there seems to be no other option. 

### Optional to |

In a number of type hints, updated `Optional` to the recommended `| None`.

### Used the new model_dump

In a couple of places the new `Basemodel.model_dump()` came in handy for shortening code.

### Removed unused code

* `DataStructureConfig.check_repository_consistency` was always covered by another validator
* In `RegionProcessor._apply_region_processing` it is not necessary to check that all the regions are valid. This is already done at initialization of the processor.


danielhuppmann (approved):
Thank you for the thorough refactoring, looks all good to me!

PR #300, opened 2024-01-05
Bring cli back

### LINKED ISSUE #299: CLI is gone in v0.13.0 release
As pointed out in iiasa/geo7-workflow#22 the v0.13.0 unintentionally dropped the CLI.
This also illustrated that we currently do not test whether the CLI is actually installed.

phackstock:
Closes #299.

This PR brings back the cli which was accidentally dropped in v0.13.0. 
It also adds an explicit test to see if the command `nomenclature` is recognized by the system after install. 
Previously we'd only been using the test runner from click which seemingly also works if the cli is not actually installed. This is why this error was not caught.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,8 @@ sphinx = "^7.2.6"
 sphinx-click = "^5.1.0"
 numpydoc = "^1.6.0"
 
+[tool.poetry.scripts]
+nomenclature = 'nomenclature.cli:cli'
 
 [tool.poetry-dynamic-versioning]
 bump = true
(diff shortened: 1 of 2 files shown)


danielhuppmann (approved):
Thanks, looks good to me (I assume that the failing tests on Windows are just a windows-issue with the test, not the cli itself).

phackstock:
> I assume that the failing tests on Windows are just a windows-issue with the test, not the cli itself

I would assume so too, seeing as there were already some special settings required in the test GitHub action to get it running on Windows I would say it's OS related.

I tried a couple of things but couldn't get it to work easily so I (for now) marked it as expected to fail.

Issue #104, opened 2022-02-24
Raise all duplicate-errors during initialization


danielhuppmann:
Currently, only a unique duplicate-code error is raised during initialization (e.g., from `validate-project`). It would be great to collect all duplicates and raise one pydantic-error-collection.

Issue #309, opened 2024-01-16
Improve test coverage


phackstock:
Although we have overall very good test coverage (97%), there are a few modules which are lacking. Especially the config module has a few error cases that are currently not tested.
| **Module**                              | **statements** | **missing** | **excluded** | **coverage** |
|-----------------------------------------|----------------|-------------|--------------|--------------|
| nomenclature/__init__.py                | 25             | 7           | 0            | 72%          |
| nomenclature/cli.py                     | 43             | 7           | 0            | 84%          |
| nomenclature/processor/processor.py     | 7              | 1           | 0            | 86%          |
| nomenclature/config.py                  | 89             | 9           | 0            | 90%          |
| nomenclature/countries.py               | 35             | 2           | 0            | 94%          |
| nomenclature/codelist.py                | 252            | 9           | 0            | 96%          |
| nomenclature/processor/region.py        | 308            | 11          | 0            | 96%          |
| nomenclature/code.py                    | 109            | 3           | 0            | 97%          |
| nomenclature/processor/required_data.py | 115            | 3           | 0            | 97%          |
| nomenclature/testing.py                 | 65             | 2           | 0            | 97%          |
| nomenclature/validation.py              | 35             | 1           | 0            | 97%          |
| nomenclature/definition.py              | 52             | 1           | 0            | 98%          |
| nomenclature/core.py                    | 22             | 0           | 0            | 100%         |
| nomenclature/error/__init__.py          | 0              | 0           | 0            | 100%         |
| nomenclature/error/codelist.py          | 4              | 0           | 0            | 100%         |
| nomenclature/error/region.py            | 16             | 0           | 0            | 100%         |
| nomenclature/error/required_data.py     | 2              | 0           | 0            | 100%         |
| nomenclature/error/variable.py          | 10             | 0           | 0            | 100%         |
| nomenclature/processor/__init__.py      | 3              | 0           | 0            | 100%         |
| nomenclature/processor/meta.py          | 21             | 0           | 0            | 100%         |
| nomenclature/processor/utils.py         | 3              | 0           | 0            | 100%         |
| tests/conftest.py                       | 32             | 0           | 0            | 100%         |
| tests/test_check_aggregate.py           | 26             | 0           | 0            | 100%         |
| tests/test_cli.py                       | 81             | 0           | 0            | 100%         |
| tests/test_code.py                      | 42             | 0           | 0            | 100%         |
[...]

PR #311, opened 2024-01-22
Use multiple external repos for definitions and mappings

### LINKED ISSUE #295: Mix multiple external repos in a single dimension
The following use case is currently not supported. 
Import **multiple** external repositories into the **same** definition dimension:

```yaml
repositories:
  common-definitions:
    url: https://github.com/iamconsortium/common-definitions
  legacy-definitions:
    url: https://github.com/iamconsortium/common-definitions
definitions:
  variable:
    repository: [common-definitions, legacy-definitions]
```

### LINKED ISSUE #313: Removing repositories after tests fails on Windows
As discovered in #311, the function `clean_up_external_repos` fails on Windows due to a permission error.
This was previously undiscovered since `ignore_errors=True` was used.

phackstock:
Closes #295, closes #313.

This PR allows for using multiple external repos combined in a single dimension.
The complete test I implemented for that pulls from [common-definitions](https://github.com/IAMconsortium/common-definitions/) and [legacy-definitions](https://github.com/IAMconsortium/legacy-definitions). 
Since legacy-definitions currently contains duplicates from common-definitions initiating a CodeList will result in an error. The test tests for this error. 
Once this PR is merged, I'll do the following:
1. Address #104 so that all the duplicate codes are raised in a single error
2. Fix legacy-definitions so that they are compatible with common-definitions
3. At this point the multiple repo test here will fail since it will no longer produce an error so I'll fix the test.

All of that should be possible today.

### DIFF
--- a/nomenclature/__init__.py
+++ b/nomenclature/__init__.py
@@ -57,8 +57,7 @@ def parse_model_registration(
     model_registration_file : str, path, file-like object
         Path to xlsx model registration file.
     output_directory : str, path, file-like object
-        Directory where the model mapping and region file will be saved;
-        defaults to current working directory
+        Directory, where the model mapping and region file will be saved; default: "."
     """
     if not isinstance(output_directory, Path):
         output_directory = Path(output_directory)
--- a/nomenclature/__init__.py
+++ b/nomenclature/__init__.py
@@ -57,7 +57,8 @@ def parse_model_registration(
     model_registration_file : str, path, file-like object
         Path to xlsx model registration file.
     output_directory : str, path, file-like object
-        Directory, where the model mapping and region file will be saved; default: "."
+        Directory where the model mapping and region file will be saved;
+        defaults to current working directory
     """
     if not isinstance(output_directory, Path):
         output_directory = Path(output_directory)
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -202,17 +202,17 @@ class CodeList(BaseModel):
         with suppress(AttributeError):
             dimension = path.name
             codelistconfig = getattr(config.definitions, dimension)
-            repo_path = (
-                config.repositories[codelistconfig.repository].local_path
-                / codelistconfig.repository_dimension_path
-            )
-            code_list = (
-                cls._parse_codelist_dir(
-                    repo_path,
-                    file_glob_pattern,
+            for repo in codelistconfig.repositories:
+                repo_path = (
+                    config.repositories[repo].local_path / "definitions" / dimension
                 )
-                + code_list
-            )
+                code_list = (
+                    cls._parse_codelist_dir(
+                        repo_path,
+                        file_glob_pattern,
+                    )
+                    + code_list
+                )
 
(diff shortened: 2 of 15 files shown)


nomenclature/__init__.py
```diff
@@ -57,8 +57,7 @@ def parse_model_registration(
     model_registration_file : str, path, file-like object
         Path to xlsx model registration file.
     output_directory : str, path, file-like object
-        Directory where the model mapping and region file will be saved;
-        defaults to current working directory
+        Directory, where the model mapping and region file will be saved; default: "."
```
danielhuppmann, line 60: This reverts my previous suggestions

nomenclature/codelist.py
```diff
@@ -615,17 +615,18 @@ def from_directory(
                         code_list.append(RegionCode(name=c.name, hierarchy="Country"))
 
             # importing from an external repository
-            if config.definitions.region.repository:
+            for repository in config.definitions.region.repositories:
```
danielhuppmann, line 618: For consistency with the loop above...
```suggestion
            for repo in config.definitions.region.repositories:
```

nomenclature/codelist.py
```diff
@@ -639,7 +640,7 @@ def from_directory(
         mapping: Dict[str, RegionCode] = {}
         for code in code_list:
             if code.name in mapping:
-                raise DuplicateCodeError(name=name, code=code.name)
+                raise ValueError(f"Trying to set a duplicate code {code.name}")
```
danielhuppmann, line 643: Is this reversion to a ValueError done on purpose here?

nomenclature/config.py
```diff
@@ -1,22 +1,41 @@
...
+)
+
+
+def convert_to_set(v: str | list[str] | set[str]) -> set[str]:
+    match v:
+        case str(v):
+            return {v}
+        case list(v):
+            return set(v)
+        case set(v):
```
danielhuppmann, line 23: Do we want to allow a set type?

phackstock:
Looks like we're having some Windows issues... I hope I can resolve them quickly.

danielhuppmann (commented):
Looks good in principle, a few suggestions inline

nomenclature/codelist.py
phackstock: Yes it is, this was actually a left over from the move to pydantic v2. In v2 pydantic does not really allow you to create custom errors anymore so I removed them all. This one slipped through the cracks also because there was no test for it. When working on #309 I noticed that but forgot to address it there.
If you want I can bring back the `DuplicateCodeError` as a custom error. This time though it would be derived from a `ValueError` and not from a pydantic error.

danielhuppmann: Ok, thanks!

nomenclature/__init__.py
phackstock: Sorry for that, got lost in the rebasing process. Fixed now.

danielhuppmann (approved):
Looks good, thanks!

PR #314, opened 2024-01-22
Collect duplicate code errors

### LINKED ISSUE #104: Raise all duplicate-errors during initialization

### LINKED ISSUE #312: Fix multiple repo test
Once IAMconsortium/legacy-definitions#1 is merged the test `test_multiple_external_repos` will fail as it will no longer raise an error.
The `VariableCodeList` will be created successfully so the test will then need to test for items out of common-definitions and legacy-definitions.

phackstock:
Closes #104, closes #312.

Instead of reporting duplicate code errors one by one we now collect them and raise them as one single error. For better readability the individual errors are indented and numbered.
From previously reporting:

```console 
ValueError: Duplicate item in variable codelist: Secondary Energy|Liquids
```
we now have:

```console
ValueError: Collected 529 errors:
  1. Duplicate item in variable codelist: Secondary Energy|Liquids
  2. Duplicate item in variable codelist: Secondary Energy|Liquids|Biomass
  3. Duplicate item in variable codelist: Secondary Energy|Liquids|Electricity
  4. Duplicate item in variable codelist: Secondary Energy|Liquids|Other
  5. Duplicate item in variable codelist: Secondary Energy|Solids
  6. Duplicate item in variable codelist: Secondary Energy|Solids|Biomass
  7. Duplicate item in variable codelist: Secondary Energy|Solids|Fossil
  8. ...
```

### Update 24.01.2024 10:09

Since iamconsortium/legacy-definitions#1 has already been resolved, I had to also address #311.
The current way the test `test_multiple_external_repos` is configured is not the most elegant way but it should be ok.
It tests for the following:

1. Are the two external repos correctly parsed from the yaml file
2. Are the two external repos correctly fetched from GitHub
3. Is the overall size of the resulting variable list (common + legacy definitions) bigger than 2000 entries. 

The reason I opted for 3. instead of testing for the presence of specific variables is that we expect to migrate variables from legacy to common definitions. This way it could happen that at some point all the variables that I would test for to prove that both legacy and common definitions are imported are now wholly in common-definitions. 
That's why I opted to check for the overall size of the code list. Also not 100% future proof but good enough for now I'd say.

### DIFF
--- a/nomenclature/__init__.py
+++ b/nomenclature/__init__.py
@@ -57,8 +57,7 @@ def parse_model_registration(
     model_registration_file : str, path, file-like object
         Path to xlsx model registration file.
     output_directory : str, path, file-like object
-        Directory where the model mapping and region file will be saved;
-        defaults to current working directory
+        Directory, where the model mapping and region file will be saved; default: "."
     """
     if not isinstance(output_directory, Path):
         output_directory = Path(output_directory)
--- a/nomenclature/__init__.py
+++ b/nomenclature/__init__.py
@@ -57,7 +57,8 @@ def parse_model_registration(
     model_registration_file : str, path, file-like object
         Path to xlsx model registration file.
     output_directory : str, path, file-like object
-        Directory, where the model mapping and region file will be saved; default: "."
+        Directory where the model mapping and region file will be saved;
+        defaults to current working directory
     """
     if not isinstance(output_directory, Path):
         output_directory = Path(output_directory)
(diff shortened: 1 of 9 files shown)


danielhuppmann (approved):
Looks good, thanks!

PR #320, opened 2024-02-08
Allow nomenclatureconfig to have region country only

### LINKED ISSUE #319: Using only country in nomenclature config breaks CodeList
I saw this in https://github.com/iiasa/openmod4africa-workflow, where the latest nomenclature update in combination with a nomenclature config where in the region dimension there's only the country attribute:

```yaml
repositories:
  openENTRANCE:
    url: https://github.com/openENTRANCE/openentrance
definitions:
  variable:
    repository: openENTRANCE
  region:
    country: true
```

It broke in the test here: https://github.com/iiasa/openmod4africa-workflow/actions/runs/7829213013/job/21360733654. 

Issue is that https://github.com/IAMconsortium/nomenclature/blob/86d38f9b4457f897445bee387554d35e45ae873e/nomenclature/codelist.py#L622 
attempts to iterate over the region.repositories attribute which is empty.

phackstock:
Closes #319.

I took this issues as an opportunity to substitute all the `None` values in `NomenclatureConfig` for empty instances of e.g. `CodeListConfig`.
The advantage of the is that you no longer need to check if a attribute exists in `NomenClatureConfig` but instead you can just iterate over it.

### DIFF
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -196,21 +196,14 @@ class CodeList(BaseModel):
 
         """
         code_list = cls._parse_codelist_dir(path, file_glob_pattern)
-
-        with suppress(AttributeError):
-            dimension = path.name
-            codelistconfig = getattr(config.definitions, dimension)
-            for repo in codelistconfig.repositories:
-                repo_path = (
-                    config.repositories[repo].local_path / "definitions" / dimension
-                )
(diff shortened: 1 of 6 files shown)


danielhuppmann (approved):
I don't want to pretend that I understand everything, but it looks nice...

PR #322, opened 2024-02-09
Making aggregation compatible with pyam 2.1

### LINKED ISSUE #321: Update to pyam 2.1 breaks test_aggregation_differences_export
When working on #320, I discovered that upgrading from pyam 2.0 to 2.1 breaks the third configuration of the test `test_aggregation_differences_export`:

```python
# check that region-aggregation with missing weights passes (inconsistent index)
        # TODO check the log output
        (
            "weighted_aggregation",
            [
                ["model_a", "scen_a", "World", "Primary Energy", "EJ/yr", 4, 6],
                ["model_a", "scen_a", "World", "Emissions|CO2", "Mt CO2", 5, np.nan],
            ],
            dict(variable="Emissions|CO2", year=2010, keep=False),
        ),
```

with the following error:

```console
E           ValueError: Missing weights for the following data rows:
E                model scenario    region  year
E           0  model_a   scen_a  region_A  2010
E           1  model_a   scen_a  region_B  2010
```

@danielhuppmann is this error expected and we should adjust the test or is there something wrong in nomenclature?
From what I saw looking at it quickly the way the test is built I'd expect it to fail since you're filtering out data from a required weight variable:

https://github.com/IAMconsortium/nomenclature/blob/86d38f9b4457f897445bee387554d35e45ae873e/tests/test_core.py#L203-L253

Not sure though why it's only failing now after updating from pyam 2.0 -> 2.1.

Looking at the diff of pyam 2.0 vs 2.1 it looks indeed like this is expected since the error of the missing weights is new (https://github.com/IAMconsortium/pyam/compare/v2.0.0...v2.1.0#diff-f71eedde8b40510caa9782ec1b2adb3eb3c8a5ed0d3c248b1ecde9637445c35eR225-R229).

phackstock:
Closes #321.
Thanks a lot @danielhuppmann for finding the issue.


nomenclature/processor/region.py
```diff
@@ -733,9 +733,12 @@ def _aggregate_region(df, var, *regions, **kwargs):
     try:
         return df.aggregate_region(var, *regions, **kwargs)
     except ValueError as error:
-        if str(error) == "Inconsistent index between variable and weight!":
+        if str(error).startswith("Missing weights for the following data"):
             logger.info(
-                f"Could not aggregate '{var}' for region '{regions[0]}' ({kwargs})"
+                (
+                    f"Could not aggregate '{var}' for region '{regions[0]}' "
+                    f"({kwargs})\nerror: {error}"
```
danielhuppmann, line 740: No need to write "error" in the log message, right?
```suggestion
                    f"({kwargs})\n{error}"
```

danielhuppmann (approved):
Looks good, just one suggestion for more readable log messages. Thanks for diving into this issue!

nomenclature/processor/region.py
phackstock: Fair point

PR #324, opened 2024-02-09
Disallow external repo double stacking

### LINKED ISSUE #316: Double stacking external repos does not work as expected
I just ran into this potential issue when trying to create a DataStructure definition based of off legacy definitions.
What I did is:
1. Create a new (local) repository with an empty definitions folder and a `nomenclature.yaml` file with the following entry:
```yaml
repositories:
  legacy-definitions:
    url: https://github.com/IAMconsortium/legacy-definitions.git/
definitions:
  variable:
    repository: legacy-definitions
```
2. Since common definitions, as of right now (https://github.com/IAMconsortium/legacy-definitions/blob/main/nomenclature.yaml) contains a `nomenclature.yaml` file itself I thought that the above configuration should import common-definitions as well. However common-definitions was not fetched and instead just the yaml files from legacy-definitions were read.

I would see two ways to fix this, one easy one slightly more involved (because recursion).

### Option 1 (the quick fix)

Disallow any external repo to have a `nomenclature.yaml` on its own. This is easy enough to check and we would simply throw an error if this occurs.

### Option 2 (recursion)

We allow, potentially arbitrary many times, stacking of external repos. 

My preferred option would be 1. No only because it is faster to implement but also because the use case 2 currently does not exist and I'm not sure we even want to allow it for the sake of keeping things flat and easy ( "*Flat is better than nested.*" from https://peps.python.org/pep-0020/). 

What are your thoughts on the matter @danielhuppmann?

phackstock:
Closes #316.
 
This PR implements a check that an external repo does not reference another external repo, thus safeguarding against external repo double stacking.

### DIFF
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -92,8 +92,8 @@ class Repository(BaseModel):
             if config.get("repositories"):
                 raise ValueError(
                     (
-                        "No external repos allowed in external repo, found in "
-                        f"nomenclature.yaml in '{self.url}'"
+                        "External repos cannot again refer to external repos, "
+                        f"found in nomenclature.yaml in '{self.url}'"
                     )
                 )
 
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -93,7 +93,7 @@ class Repository(BaseModel):
                 raise ValueError(
                     (
                         "No external repos allowed in external repo, found in "
-                        f"nomenclature.yaml in {self.url}"
+                        f"nomenclature.yaml in '{self.url}'"
                     )
                 )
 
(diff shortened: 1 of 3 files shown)


nomenclature/config.py
```diff
@@ -82,6 +82,20 @@ def fetch_repo(self, to_path):
...
+    def check_external_repo_double_stacking(self):
+        nomenclature_config = self.local_path / "nomenclature.yaml"
+        if nomenclature_config.is_file():
+            with open(nomenclature_config, "r") as f:
+                config = yaml.safe_load(f)
+            if config.get("repositories"):
+                raise ValueError(
+                    (
+                        "No external repos allowed in external repo, found in "
+                        f"nomenclature.yaml in '{self.url}'"
```
danielhuppmann, line 96:
```suggestion
                        "External repos cannot again refer to external repos, "
                        f"found in nomenclature.yaml in '{self.url}'"
```

danielhuppmann (approved):
Very nice, just one minor suggestion on the error message

PR #330, opened 2024-02-19
Do not show full error message for missing weights

danielhuppmann:
While working on the ECEMF-workflow migration, I realized that showing the full log message from the "missing-weights" error from pyam was really too much in the nomenclature output...

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -735,9 +735,7 @@ def _aggregate_region(df, var, *regions, **kwargs):
     except ValueError as error:
         if str(error).startswith("Missing weights for the following data"):
             logger.info(
-                (
-                    f"Could not aggregate '{var}' for region '{regions[0]}' ({kwargs})"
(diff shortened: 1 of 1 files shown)


phackstock (approved):
Looks good to me in principle but if I remember correctly I put the full error there so I would get the complete output of all the missing values making it easier for debug.
I can of course also see the need for just getting a compact list.

danielhuppmann:
But the error message only shows the first few lines anyway, and in this way, it's not really helpful...
image

phackstock:
Ah I see, not super useful indeed if there's lots of things missing. I'd still want a way to know exactly what's missing though. Maybe something like this:

```console
missing for
WITCH 5.4, DIAG-BASE, CANADA, [2005, 2010, 2015, 2020]
WITCH 5.4, DIAG-BASE, BRAZIL, [2005, 2010, 2015, 2020]
...
```

where the years are compacted

PR #331, opened 2024-03-11
Extend the `to_excel()` method and add to CLI

danielhuppmann:
This PR adds extends the method `DataStructureDefinition.to_excel()` to include all dimensions and add an overview-sheet with project name, a timestamp and urls/commits/timestamps of external repositories.


nomenclature/definition.py
```diff
@@ -136,22 +146,67 @@ def check_aggregate(self, df: IamDataFrame, **kwargs) -> None:
             error = pd.concat(lst)
             return error if not error.empty else None
 
-    def to_excel(
-        self, excel_writer, sheet_name=None, sort_by_code: bool = False, **kwargs
-    ):
-        """Write the *variable* codelist to an Excel sheet
+    def to_excel(self, excel_writer, sort_by_code: bool = False, **kwargs):
```
phackstock, line 149: I would simplify the `excel_writer` input to just be a str or path object. I don't think the use case where we would want to feed in an ExcelWriter object really exists.
To further simplify I would also take out the `sort_by_code` input and manually set it to true for all code lists. 
Also if there's no concrete use case for the kwargs, I would take them out as well.

tests/test_cli.py
```diff
@@ -312,3 +313,23 @@ def test_check_region_aggregation(tmp_path):
         )
     )
     assert_iamframe_equal(IamDataFrame(tmp_path / "results.xlsx"), exp_result)
+
+
+def test_cli_export_to_excel(simple_definition, tmpdir):
```
phackstock, line 318: Looks like this variable is not needed
```suggestion
def test_cli_export_to_excel(tmpdir):
```

nomenclature/definition.py
```diff
@@ -136,22 +146,67 @@ def check_aggregate(self, df: IamDataFrame, **kwargs) -> None:
...
+def time_format(x):
+    return x.strftime("%Y-%m-%d %H:%M:%S")
+
+
+def git_attributes(name, repo):
+    return {
+        "repository": name,
+        "url": repo.remote().url,
+        "commit": repo.commit(),
+        "timestamp": time_format(repo.commit().committed_datetime),
```
phackstock, line 199: Maybe change it to "last commit at" to make it more obvious what it means:
```suggestion
        "last commit at": time_format(repo.commit().committed_datetime),
```

nomenclature/definition.py
```diff
@@ -136,22 +146,67 @@ def check_aggregate(self, df: IamDataFrame, **kwargs) -> None:
...
-        self.variable.to_excel(excel_writer, sheet_name, sort_by_code, **kwargs)
+        with pd.ExcelWriter(excel_writer, engine="xlsxwriter", **kwargs) as writer:
+
+            # create dataframe with attributes of the DataStructureDefinition
+            arg_dict = {
+                "project": self.working_dir.absolute().parts[-1],
+                "file created": time_format(datetime.now()),
+                "": "",
+            }
+            if self.repo is not None:
```
phackstock, line 170: Should we allow creating an Excel list from a non-git repository or should we require it?

nomenclature/definition.py
```diff
@@ -41,11 +44,18 @@ def __init__(self, path, dimensions=None):
         if not isinstance(path, Path):
             path = Path(path)
 
-        if (file := path.parent / "nomenclature.yaml").exists():
+        self.working_dir = path.parent
```
phackstock, line 47: I think the name of this variable is confusing. It's supposed to be the parent folder where the definition files are located, right? Normally, working directory is the directory where you call a script from.
Maybe project folder is a better name:
 ```suggestion
        self.project_folder = path.parent

```

File: nomenclature/definition.py [hunk 6]
```diff
@@ -136,22 +146,67 @@ def check_aggregate(self, df: IamDataFrame, **kwargs) -> None:
...
+        with pd.ExcelWriter(excel_writer, engine="xlsxwriter", **kwargs) as writer:
+
+            # create dataframe with attributes of the DataStructureDefinition
+            arg_dict = {
+                "project": self.working_dir.absolute().parts[-1],
+                "file created": time_format(datetime.now()),
+                "": "",
+            }
+            if self.repo is not None:
+                arg_dict.update(git_attributes("", self.repo))
```
phackstock, line 171: You could also include a check here that there are no unstaged chages in the git repository and only allow exporting of the Excel file if there's none.

nomenclature/cli.py
```diff
@@ -154,3 +154,42 @@ def check_region_aggregation(
         results_df.to_excel(processed_data)
     if differences:
         differences_df.reset_index().to_excel(differences, index=False)
+
+
+@cli.command("export-project")
```
phackstock, line 159: I would call it `export-definition`, we've used `project` in the past to mean an entire workflow directory, including model mappings, required data, etc...

nomenclature/cli.py
```diff
@@ -154,3 +154,42 @@ def check_region_aggregation(
...
+@click.argument("path", type=click.Path(exists=True, path_type=Path))
+@click.argument("target", type=click.Path(path_type=Path))
+@click.option(
```
phackstock, line 162: I would simplify and remove this option. If the definitions folder is not called definitions, I'm pretty sure other parts of the program would break. Unless there's a concrete use case for having a definitions folder not called _definitions_, I'd remove it.

nomenclature/cli.py
```diff
@@ -154,3 +154,42 @@ def check_region_aggregation(
...
+    "--definitions",
+    help="Optional name for definitions folder",
+    type=str,
+    default="definitions",
+)
+@click.option(
```
phackstock, line 168: Same argument as above, I'd simplify the command by just fixing the exporting for all dimensions. Unless of course there's a good reason for this option.

phackstock (commented):
Looks very nice, thanks a lot for this. 
A few comments and questions in line.

nomenclature/definition.py
```diff
@@ -136,22 +146,67 @@ def check_aggregate(self, df: IamDataFrame, **kwargs) -> None:
...
+
+
+def time_format(x):
+    return x.strftime("%Y-%m-%d %H:%M:%S")
+
+
+def git_attributes(name, repo):
+    return {
+        "repository": name,
+        "url": repo.remote().url,
```
phackstock, line 197: You could prepend the _url_, _commit_ and _timestamp_ attributes with the name of the repository to get e.g. `common-definitions.url` and `common-definitions.commit` just to make it easier to read when there's multiple external repos.

nomenclature/definition.py
danielhuppmann: Requiring a git repository would (at least) make the tests more complex, and I don't see why it should be a hard requirement.

nomenclature/definition.py
danielhuppmann: Agree on the simplificatons, but let's leave in the kwargs - they are also less ambiguous now thanks to the simplification.

tests/test_cli.py
danielhuppmann: Thanks for catching that.

nomenclature/definition.py
danielhuppmann: This is basically superseded by your later suggestion, I guess...

nomenclature/definition.py
danielhuppmann: Nice...

nomenclature/cli.py
danielhuppmann: Ok.

nomenclature/definition.py
phackstock: True, good point.

nomenclature/definition.py
phackstock: Fair point.

tests/test_definition.py
```diff
@@ -78,18 +78,33 @@ def test_to_excel(simple_definition, tmpdir):
...
 
 
+def test_to_excel_with_external_repo(tmpdir):
+    """Check writing a DataStructureDefinition with an external repo to file"""
+    file = tmpdir / "testing_export.xlsx"
+
+    dsd = DataStructureDefinition(TEST_DATA_DIR / "general-config" / "definitions")
+    dsd.to_excel(file)
+
+    obs = pd.ExcelFile(file)
```
phackstock, line 93: Although not a critical as in the pyam issue iamconsortium/pyam#817, I'd also use a context manager here to make sure the file is properly closed at the end of the test:
```suggestion
    with pd.ExcelFile(file) as obs:
```

phackstock (changes requested):
One small remark, after that good to merge from my side.

tests/test_definition.py
danielhuppmann: Implemented, thanks for the suggestion

PR #334, opened 2024-03-11
Revert country name to "Turkey"

danielhuppmann:
In September 2023, pycountry implemented a. change in the name of Turkey to Türkiye, which now causes some issues in the SSP- and SSP-Extensions databases. This PR implements an override for this country to keep existing IIASA databases operational with the use in nomenclature.

This also indicates that we should have a strategy to avoid different behavior with different versions of pycountry.


phackstock (approved):
Thanks for the quick fix. I'd be happy to pin pycountry for the future.

danielhuppmann:
Added a pin for pycountry to this PR.

PR #335, opened 2024-03-12
Set repository attribute correctly

### LINKED ISSUE #332: Setting of repository attribute slightly broken
This is a leftover from when there was only one external repository per dimension.
We're setting the entire list of external repositories as this attribute instead of just the actual repo.

phackstock:
Closes #332.


danielhuppmann (approved):
Looks good!

PR #336, opened 2024-03-14
Use correct virtual environment for building the docs on ReadTheDocs

danielhuppmann:
Seems that since poetry 1.8, there is an issue that the new poetry installs to the wrong virtual environment, see https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-poetry.

This PR mirrors https://github.com/IAMconsortium/pyam/pull/836

FYI @phackstock

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -13,30 +13,35 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


danielhuppmann:
Correctly rendered docs at https://nomenclature-iamc.readthedocs.io/en/devops-docs-poetry/

PR #338, opened 2024-03-14
Improve duplicate code output

### LINKED ISSUE #337: Show source file and line for duplicate items

phackstock:
Closes #337.

This came up when trying to combine variable lists for the IAM and short term models for NGFS.
There were 77 duplicate entries and I had to figure out which ones to keep, remove, rename, etc...

To this end I've updated the output of the error in case we have duplicate CodeList entries.
From:
```console

```

we now have:
```console
  1. Identical duplicate items in 'variable' codelist: 'Some Variable'
  2. Conflicting duplicate items in 'variable' codelist: 'Some other Variable'
    {'file': 'variable/duplicate_entries.yaml', 'extra_attributes': {'unit': 'a'}}
    {'file': 'variable/duplicate_entries.yaml', 'extra_attributes': {'unit': 'b'}}
```

The two key changes are:
1. As the error is raise the two codes are compared. The result of the comparison is included in the error message for quicker debugging.
2. If the codes are not identical, their values are shown so that you can work through the differences faster.

### DIFF
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -1,5 +1,6 @@
 import logging
 from pathlib import Path
+from textwrap import indent
 from typing import ClassVar, Dict, List
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


danielhuppmann:
Thanks, really nice work!

One suggestion: would it be helpful to also include the location (file/repository) for identical code items, like

```python
  1. Identical duplicate items in 'variable' codelist: 'Some Variable'
    {'file': 'variable/file_1.yaml'}
    {'file': 'variable/file_2.yaml'}
  2. Conflicting duplicate items in 'variable' codelist: 'Some other Variable'
    {'file': 'variable/duplicate_entries.yaml', 'extra_attributes': {'unit': 'a'}}
    {'file': 'variable/duplicate_entries.yaml', 'extra_attributes': {'unit': 'b'}}
```

PR #349, opened 2024-05-28
Pin versions of common and legacy definitions for test

### LINKED ISSUE #348: Nightly tests failing because of legacy definitions update
Since the update https://github.com/IAMconsortium/legacy-definitions/commit/bad8c0cf66638e8cf1a00ae925c39fb9b99bae07, the test that pulls from common- and legacy-definitions has been failing. The reason for that is that it tested for the presence of the variable `Employment` in the legacy-definitions. With the above commit it was moved. 
To me the most obvious fix is to use the already existing pin feature for common- and legacy-definitions rather than adjust the variables we test for.

FYI @danielhuppmann

phackstock:
Closes #348.

### DIFF
--- a/tests/data/nomenclature_configs/multiple_repos_per_dimension.yaml
+++ b/tests/data/nomenclature_configs/multiple_repos_per_dimension.yaml
@@ -1,8 +1,10 @@
 repositories:
   common-definitions:
     url: https://github.com/IAMconsortium/common-definitions.git/
+    hash: 013295a
   legacy-definitions:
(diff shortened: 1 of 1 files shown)


danielhuppmann:
Is there already a test for the hash feature? If yes, I‘d prefer to change the variable that is being tested. Otherwise ok.

phackstock:
Just checked, there actually isn't a test for the hash feature.

danielhuppmann (approved):
Thanks!

Issue #342, opened 2024-04-15
Replace tags in attribute-lists


danielhuppmann:
Per https://github.com/IAMconsortium/common-definitions/pull/91, the package should support replacing tags if the attribute is a list of strings.

phackstock:
Closing, since addressed per #352.

PR #352, opened 2024-06-25
Replace tags in attribute lists and dicts

korsbakken:
This PR (hopefully) solves issue #342 . It modifies `Code.replace_tag` to descend into attribute lists and dicts, replaing each string in list elements of dict values that contain a tag. Adds a test and test data that contains a code with tags in the `components` attribute.


tests/data/tagged_codelist/foo_attr_list.yaml
```diff
@@ -0,0 +1,16 @@
+- Emissions|Species:
+    # the "Sector" guards against feature regression (conflict with "Sector" tag)
+    definition: Emissions of Species
+    unit: tSpecies/yr
+
```
danielhuppmann, line 5: This feature regression is already tested in `foo.yaml` so suggest to remove here for simplicity.
```suggestion
```

tests/data/tagged_codelist/tag_species.yaml
```diff
@@ -0,0 +1,7 @@
+- Species:
+  - CO2:
+     definition: CO2
+     unit: tCO2
```
danielhuppmann, line 4: For consistency with the species in [common-definitions](https://github.com/IAMconsortium/common-definitions/blob/e009dfe10b3eb2c8440b114e44ee3e929c0cabae/definitions/variable/emissions/tag_level-1-species.yaml#L4).

Also, if "definition" would not be specified here, the nomenclature package would replace the tag in "definition" in the code item by the name (i.e., also "CO2"), so replacing it with a longer string helps to make sure that it uses the right replacement (the replacement from the "definition", not the item name.)
```suggestion
     definition: carbon dioxide (CO2)
     unit: Mt CO2
```

tests/data/tagged_codelist/tag_species.yaml
```diff
@@ -0,0 +1,7 @@
...
+  - CH4:
+      definition: CH4
+      unit: tCH4
```
danielhuppmann, line 7:
```suggestion
        description: methane (CH4)
        unit: Mt CH4/yr

```

tests/test_codelist.py
```diff
@@ -110,6 +110,45 @@ def test_tagged_codelist():
...
+        "variable", TEST_DATA_DIR / "tagged_codelist"
+    )
+    # The test should test that the tags in the definitions in the
+    # tagged_codelist/foo_attr_list_dict.yaml file are expanded correctly.
+
+    exp = {
+        "Emissions|Species": {
+            "description": "Emissions of Species",
+            "unit": "tSpecies/yr",
+        },
```
danielhuppmann, line 125:
```suggestion
```

tests/test_codelist.py
```diff
@@ -110,6 +110,45 @@ def test_tagged_codelist():
...
+        "Emissions|CO2": {
+            "description": "Total emissions of CO2",
+            "unit": "tCO2/yr",
```
danielhuppmann, line 128:
```suggestion
            "description": "Total emissions of carbon dioxide (CO2)",
            "unit": "Mt CO2/yr",

```

tests/test_codelist.py
```diff
@@ -110,6 +110,45 @@ def test_tagged_codelist():
...
+            "check_aggregate": True,
+            "components": {
+                "By source": ["Emissions|CO2|Fossil", "Emissions|CO2|Renewables"],
+                "By sector": ["Emissions|CO2|Energy", "Emissions|CO2|Industry"],
+            },
+        },
+        "Emissions|CH4": {
+            "description": "Total emissions of CH4",
+            "unit": "tCH4/yr",
```
danielhuppmann, line 137:
```suggestion
            "description": "Total emissions of methane (CH4)",
            "unit": "Mt CH4/yr",
```

nomenclature/code.py
```diff
@@ -112,21 +112,28 @@ def replace_tag(self, tag: str, target: "Code") -> "Code":
...
-                    )
-                # otherwise insert the name
+                if _attr in target.flattened_dict:
+                    return _value.replace("{" + tag + "}", getattr(target, _attr))
+                # otherwise return the name
                 else:
-                    mapping[attr] = value.replace("{" + tag + "}", target.name)
-            # otherwise append as is
+                    return _value.replace("{" + tag + "}", getattr(target, "name"))
+            # If the attribute is a list, descend into it and replace
```
danielhuppmann, line 124:
```suggestion
            # if the attribute is a list, iterate over the items and replace tags
```

nomenclature/code.py
```diff
@@ -112,21 +112,28 @@ def replace_tag(self, tag: str, target: "Code") -> "Code":
...
+            elif isinstance(_value, list):
+                return [_replace_or_recurse(_attr, _v) for _v in _value]
+            # If the attribute is a dict, descend into it and replace the values
```
danielhuppmann, line 127:
```suggestion
            # if the attribute is a mapping, iterate over the items and replace tags
            # in the values (not the keys)
```

danielhuppmann (commented):
Thanks @korsbakken, very nice - just a few suggestions for consistency...

korsbakken:
> Thanks @korsbakken, very nice - just a few suggestions for consistency...

Thanks @danielhuppmann, those suggestions make good sense, I have added and committed them to the PR now.

korsbakken:
The suggested changes appear to cause tests to fail, not clear why. It introduces an unintentional blank line in the middle of a definition, and an inconsistency in the CO2 and CH4 unit definitions, but new less obvious failures appear after fixing that.

tests/data/tagged_codelist/foo_attr_list.yaml
```diff
@@ -0,0 +1,11 @@
+- Emissions|{Species}:
+    definition: Total emissions of {Species}
+    unit: "{Species}/yr"
```
danielhuppmann, line 3:
```suggestion
    unit: "{Species}"
```

tests/data/tagged_codelist/tag_species.yaml
```diff
@@ -0,0 +1,7 @@
+- Species:
+  - CO2:
+     definition: carbon dioxide (CO2)
+     unit: Mt CO2
```
danielhuppmann, line 4:
```suggestion
        definition: carbon dioxide (CO2)
        unit: Mt CO2/yr
```

danielhuppmann (commented):
Ha, the "/yr" was specified both in the code itself and the tag...

tests/data/tagged_codelist/tag_species.yaml
korsbakken: Thanks. There was also inconsistent indentation in tag_species.yaml that was causing errors. I had already made fixes locally. I have committed and pushed those now, so I haven't committed your suggestions.

My commit is slightly different from yours. I kept `unit: {Species}/yr` in the `Emissions|{Species}` tagged variable, and instead removed `/yr` from both unit attributes in tag_species.yaml. The species tag could in principle also be used for variables representing cumulative emissions, emission intensity, etc. Those could all have units with `Mt {Species}` in the denominator, but would need different or no units in the denominator.

It doesn't matter here of course, so I'm happy to move where `/yr` goes if you have a strong preference.

danielhuppmann (approved):
Looks good, thanks!

tests/data/tagged_codelist/tag_species.yaml
danielhuppmann: Good point, thanks.

PR #356, opened 2024-07-09
Fix enum use in config

### LINKED ISSUE #355: Dimensions attribute in config file not correctly converted into path
Currently for the dimension path, we get the following:

```console
str(path / dim)
'definitions/DimensionEnum.region'
```

when we should get:
```console
'definitions/region'
```

phackstock:
Closes #355 and adds an explicit test for getting `DataStructureDefinitions` from config file.


danielhuppmann (approved):
Looks good to me, thanks!

PR #360, opened 2024-07-26
Create new `IamcDataFilter` class

danielhuppmann:
This PR is a preparation step for a new **DataValidator** processor, see https://github.com/IAMconsortium/common-definitions/pull/104

Given the overlap with the already-existing **RequiredData** processor, the first step is to pull out a common underlying class to collect IAMC-format filters.

In the process, I've been debating with myself whether the **DataValidator** should support the *measurand* feature...

Any thoughts @phackstock?


phackstock:
> In the process, I've been debating with myself whether the DataValidator should support the measurand feature...

At some point it might be nice/necessary to do that. For a first implementation though I'd go for maximum simplicity. If the unit is not what we expect, the numbers will be wrong and it would be caught by the validator anyway.

danielhuppmann:
Seems like upgrading to matplotlib 3.9.1.post1 does the trick...

nomenclature/iamc/datapoint.py
```diff
@@ -0,0 +1,37 @@
...
+    variable: List[str] | None = None
+    unit: List[str] | None = None
+    year: List[int] | None = None
+
+    @field_validator(*IAMC_DIMENSIONS, mode="before")
+    @classmethod
+    def single_input_to_list(cls, v):
+        return v if isinstance(v, list) else [v]
+
+    def validate(self, dsd: DataStructureDefinition) -> None:
```
phackstock, line 23: The `validate` is already defined from `BaseModel` and overwriting it would break the pydantic validation. You'll have to chose a different name. On my local machine mypy warns about this.

nomenclature/processor/required_data.py
```diff
@@ -41,14 +42,11 @@ def cast_to_RequiredMeasurand(v) -> RequiredMeasurand:
     return RequiredMeasurand(variable=variable, **v[variable])
 
 
-class RequiredData(BaseModel):
+class RequiredData(IamcDataFilter):
     measurand: (
         List[Annotated[RequiredMeasurand, BeforeValidator(cast_to_RequiredMeasurand)]]
         | None
     ) = None
-    variable: List[str] | None = None
```
phackstock, line 49: By changing the base class of `RequiredData` you're introducing some unwanted changes. 
With your changes `RequiredData` now has the additional attributes `model` and `scenario` which it did not have before.
Now the `RequiredDataValidator` which uses `RequiredData` features the `model` attribute twice. 
Once as `RequiredDataValidator.model` and as `RequiredDataValidator.required_data[0].model`. This is an issue since the latter `model` attribute does not do anything. 
As it stands a required data yaml file like this:

```yaml
required_data:
  - model: model_a
  - scenario: scenario_a
  - measurand:
    - Primary Energy:
        unit: [GWh/yr, Mtoe]
    year: [2005, 2010, 2015]
```

would be parsed without any problems **but** the requirement would be applied to **all** models and scenarios rather than the ones written since the rest of `RequiredData` and `RequiredDataValidator` is not equipped to deal with `model` and `scenario`.
If we wanted to actually apply required data only to a specific model, we'd have to do it this way:

```yaml
model: model_a
required_data:
  - measurand:
    - Primary Energy:
        unit: [GWh/yr, Mtoe]
    year: [2005, 2010, 2015]
```

I'd see a number of different solutions to the issue:

1. Remove `model` and `scenario` from `IamcDataFilter` and add them at the `DataValidator(IamcDataFilter)` level
2. Change the way that `RequiredDataValidator` works so that the first yaml example above works. 

My preference would be for option 1

EDIT: I just realized, the same argument as above also applies for `IamcDataFilter.unit`

nomenclature/iamc/datapoint.py
phackstock, line 1: Do we expect more files in the iamc folder? If not I'd move this module into the processor folder since I think it makes the folder structure unnecessarily complicated.
I'd also call the file datapoint_filter or something along those lines since that's what's defined in here.

nomenclature/iamc/datapoint.py
```diff
@@ -0,0 +1,37 @@
...
+
+class IamcDataFilter(BaseModel):
+    model: List[str] | None = None
+    scenario: List[str] | None = None
+    region: List[str] | None = None
+    variable: List[str] | None = None
+    unit: List[str] | None = None
+    year: List[int] | None = None
+
+    @field_validator(*IAMC_DIMENSIONS, mode="before")
```
phackstock, line 18: I think you're not applying the validator to `year` since it's not in `IAMC_DIMENSIONS`.

phackstock (changes requested):
Thanks for the PR, couple of questions/comments below.

nomenclature/iamc/datapoint.py
```diff
@@ -0,0 +1,37 @@
+from typing import List
+from pydantic import BaseModel, field_validator
+
+
+from nomenclature.definition import DataStructureDefinition
+
+IAMC_DIMENSIONS = ["model", "scenario", "region", "variable", "unit"]
```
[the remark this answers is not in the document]
danielhuppmann: Done

nomenclature/iamc/datapoint.py
danielhuppmann: Rename to `validate_with_definition()` in parallel to the `RequiredData` class.

danielhuppmann:
Thanks @phackstock for the thorough review. I played around a bit, but it always ended up being way more complicated than it should be, so I concluded that a little bit of duplication is better than a very complex piece of code...

nomenclature/processor/required_data.py
```diff
@@ -16,6 +16,7 @@
 
 from nomenclature.definition import DataStructureDefinition
 from nomenclature.error import ErrorCollector
+from nomenclature.processor.iamc import IamcDataFilter
```
phackstock, line 19: I'd say for now at least you can take this one out since `IamcDataFilter` is not actually used yet:
```suggestion
```

phackstock (approved):
Thanks for the re-write @danielhuppmann, fully agree with "_a little bit of duplication is better than a very complex piece of code_" :+1: 
One small in line comment but good to merge from my side.

PR #361, opened 2024-07-30
Refactor Codelist.validate

phackstock:
As a preparatory step to #357, I cleaned up the way `DataStructureDefinition` works.
It now uses the newly introduced `Codelist.validate_dataframe()` function directly, eliminating the need for `validation.py`.
As a side effect I changed the data type of a dimensionless unit from None to an empty string. 
This made the unit check for the VariableCodeList a lot easier to implement. As a side effect we are now in line with how pyam represents a dimensionless unit.


danielhuppmann (approved):
Thanks, looks good! (if you want to merge iteratively)

phackstock:
Closing as #362 was just merged.

PR #362, opened 2024-07-31
Add function to add missing variables directly to yaml codelist

### LINKED ISSUE #357: Add units for unknown variables so they can be copied directly into the variable template

phackstock:
Closes #357.

This PR also adds the "list-missing-variables" functionality as a CLI command. The intended usage is something like this:

```console
nomenclature add-missing-variables path-to-data/data.xlsx
```

if you're in the workflow directory that you want to work on, this command will add all missing variables to `definitions/variable/variables.yaml`.
Alternatively you can specify which file the new variables should be added to:

```console
nomenclature list-missing-variables path-to-data/data.xlsx --target-file specific_variable_file.yaml
```

would write to `definitions/variable/specific_variable_file.yaml`

## TODO

- [x] Write full CLI test
- [x] Add documentation about intended usage


nomenclature/codelist.py
```diff
@@ -588,6 +600,48 @@ def vars_kwargs(self, variables: List[str]) -> List[VariableCode]:
...
+            return False
+        return True
+
+    def validate_dataframe(self, df: IamDataFrame, dimension: str) -> bool:
+        # validate variables
+        all_variables_valid = super().validate_dataframe(df, dimension)
+        all_units_valid = self.validate_units(df.unit_mapping)
+        return all_variables_valid and all_units_valid
+
+    def add_missing_variables(
```
danielhuppmann, line 626: This method doesn't actually add the variables, just prints them out in an easy-to-add format...
```suggestion
        def list_missing_variables(
```

nomenclature/codelist.py
```diff
@@ -588,6 +600,48 @@ def vars_kwargs(self, variables: List[str]) -> List[VariableCode]:
...
+                mapping={
+                    variable: VariableCode(
+                        name=variable,
+                        unit=df.unit_mapping[variable],
+                    )
+                    for variable in missing_variables
+                },
+            ).to_yaml()
+
+            with open(file, "a") as f:
```
danielhuppmann, line 642: If `file` is None, does this work?

nomenclature/codelist.py
```diff
@@ -588,6 +600,48 @@ def vars_kwargs(self, variables: List[str]) -> List[VariableCode]:
...
+                f"'{v}' - expected: {'one of ' if isinstance(e, list) else ''}"
+                f"'{e}', found: '{u}'"
+                for v, u, e in invalid_units
+            ]
+            msg = "The following variable(s) are reported with the wrong unit:"
+            logging.error("\n - ".join([msg] + lst))
+            return False
+        return True
+
+    def validate_dataframe(self, df: IamDataFrame, dimension: str) -> bool:
```
danielhuppmann, line 620: Suggest to rename because only the `data` attribute is validated, not `meta`.
```suggestion
    def validate_data(self, df: IamDataFrame, dimension: str) -> bool:
```

nomenclature/codelist.py
```diff
@@ -96,6 +98,12 @@ def keys(self):
     def values(self):
         return self.mapping.values()
 
+    def validate_dataframe(self, df: IamDataFrame, dimension: str) -> bool:
```
danielhuppmann, line 101: Suggest to rename because only the `data` attribute is validated, not `meta`.
```suggestion
    def validate_data(self, df: IamDataFrame, dimension: str) -> bool:
```

nomenclature/cli.py
```diff
@@ -173,3 +174,45 @@ def cli_export_definitions_to_excel(
...
+
+@cli.command("add-missing-variables")
+@click.argument("data", type=click.Path(exists=True, path_type=Path))
+@click.option(
+    "--workflow-directory",
+    default=lambda: Path.cwd(),
+    type=Path,
+)
+@click.option("--target-file", type=str)
+def cli_add_missing_variables(
```
danielhuppmann, line 187:
```suggestion
def cli_list_missing_variables(
```

nomenclature/cli.py
```diff
@@ -173,3 +174,45 @@ def cli_export_definitions_to_excel(
...
+        Name of the target variable definition file, optional, defaults to
+        'variables.yaml'
+
+    Example
+    -------
+
+    The following command will add all the missing variables to the file
+    new_variables.yaml located in my_workflow/definitions/variable:
+
+    $ nomenclature add-missing-variables input_data.xlsx --workflow-directory
```
danielhuppmann, line 209:
```suggestion
    $ nomenclature list-missing-variables input_data.xlsx --workflow-directory
```

nomenclature/cli.py
```diff
@@ -173,3 +174,45 @@ def cli_export_definitions_to_excel(
...
+                        my_workflow
+
+    """
+    codelist_path = workflow_directory / "definitions" / "variable"
+    target_file = target_file if target_file is None else codelist_path / target_file
+    VariableCodeList.from_directory(
+        "variable",
+        codelist_path,
+    ).add_missing_variables(IamDataFrame(data), target_file)
```
danielhuppmann, line 218:
```suggestion
    ).list_missing_variables(IamDataFrame(data), target_file)
```

danielhuppmann (commented):
Nice work, a few suggestions inline...

nomenclature/codelist.py
phackstock: If `file` is `None` line 629 is intended to change it to `Path.cwd() / "definitions" / "variable"/ ...`.

danielhuppmann: Ah, got it... Just wondering, though, if copy-pasting directly from the log message wouldn't be easier in some use cases, so `file=None` could do just that...

nomenclature/cli.py
```diff
@@ -173,3 +174,45 @@ def cli_export_definitions_to_excel(
         Path and file name for the exported file
     """
     DataStructureDefinition(path / "definitions").to_excel(target)
+
+
+@cli.command("add-missing-variables")
```
danielhuppmann, line 179:
```suggestion
@cli.command("list-missing-variables")
```

nomenclature/cli.py
```diff
@@ -173,3 +174,45 @@ def cli_export_definitions_to_excel(
...
+    "--workflow-directory",
+    default=lambda: Path.cwd(),
+    type=Path,
+)
+@click.option("--target-file", type=str)
+def cli_list_missing_variables(
+    data: Path, workflow_directory: Path, target_file: Path | None
+):
+    """Add variables from a IAMC data set that are currently not part of a
+    project variable codelist
```
danielhuppmann, line 191:
```suggestion
    """Create a list of variables that are not part of the variable codelist
```

tests/test_codelist.py
```diff
@@ -412,3 +412,20 @@ def test_variable_codelist_with_duplicates_raises(CodeList):
         CodeList.from_directory(
             "variable", TEST_DATA_DIR / "duplicate-code-list" / "variable"
         )
+
+
+def test_variablecodelist_add_missing_variables_to_new_file(simple_df, tmp_path):
```
danielhuppmann, line 417:
```suggestion
def test_variablecodelist_list_missing_variables_to_new_file(simple_df, tmp_path):

```

danielhuppmann (approved):
Two more updates to use `list...` instead of `add..`. Also, did you consider my suggestion to only print to log if `file=None`?

phackstock:
> Also, did you consider my suggestion to only print to log if file=None?

I like the idea, not 100% sure how to best implement it. 
The reason why I set it up in the current way is that I'd assume the typical use case is that you are in the workflow folder and would want to have the result saved into `variable.yaml`. I could put that directly into the variable default, allow `None` and print to console in this case. 
Not sure how I'd do it with the cli command then though, because I think the only way to get None is to not call the option. 
On a more general level, this brings the question of what the default should be:

* print to console _or_ 
* print to `current working directory / "definitions" / "variable" / "variables.yaml"`

I'd lean towards the latter but if you think the former is better I'll happily change it around.

danielhuppmann:
I guess you'll need that feature more often than me, so let's go with your preference - good to merge once tests pass

PR #363, opened 2024-08-01
Run workflow file directly from cli

### LINKED ISSUE #351: Add CLI option to run workflow file from command line
For a lot of local upload trouble shooting it would be nice to run something along the lines of:

```console
nomenclature run_workflow input_data.xlsx
```
 
`run_workflow` would default to importing `workflow.py` from the current working directory and run the function `main` from `workflow.py`

phackstock:
Closes #351.

This PR adds a CLI option to directly run a processing function (in most cases `main`) from the command line:

```console
nomenclature run-workflow input_data.xlsx --output-file output.xlsx
```
to reproduce the processing and find out what might be going on. 

Using `--output-file` is optional since we're not always interested in the output of the processing.
If the current working directory is not the workflow directory and/or the processing function is not `main`, this is also covered:

```console
nomenclature run-workflow input_data.xlsx --workflow-file path/to/workflow.py --workflow-function not_main --output-file output.xlsx
```

### DIFF
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -1,5 +1,7 @@
 from pathlib import Path
 from typing import List, Optional
+import importlib.util
+import sys
 
 import click
 
@@ -198,7 +200,6 @@ def cli_list_missing_variables(
     target_file : Path | None
         Name of the target variable definition file, optional, defaults to
         'variables.yaml'
-
     Example
     -------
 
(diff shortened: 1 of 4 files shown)


nomenclature/cli.py
```diff
@@ -173,3 +175,63 @@ def cli_export_definitions_to_excel(
...
+):
+    """Run a given input file through a workflow function defined in a workflow.py
+
+    Parameters
+    ----------
+    input_file : Path
+        Input data file, must be IAMC format, .xlsx or .csv
+    workflow_file : Path
+            Path to the workflow file,
+            default: current working directory / "workflow.py"
```
danielhuppmann, line 203: Is it on purpose that "optional" is not added to the type?
```suggestion
        Path to the workflow file,
        default: current working directory / "workflow.py"
```

nomenclature/cli.py
```diff
@@ -173,3 +175,63 @@ def cli_export_definitions_to_excel(
...
+
+    Example
+    -------
+
+    Say a user is reporting issues with a Scenario Explorer processing.
+    Download the file, navigate into the workflow folder and run:
+
+    $ nomenclature run-workflow input_data.xlsx --output-file output.xlsx
+    """
+
```
danielhuppmann, line 223: I'd prefer to avoid direct references to the "Scenario Explorer" in the nomenclature docstrings... Rather put that into the docs-user-guide, including instructions on how to install nomenclature and clone the workflow repo
```suggestion
```

nomenclature/cli.py
```diff
@@ -173,3 +175,63 @@ def cli_export_definitions_to_excel(
...
+    spec.loader.exec_module(workflow)
+
+    if not hasattr(workflow, workflow_function):
+        raise ValueError(f"{workflow} does not have a function `{workflow_function}`")
+
+    if output_file is not None:
+        getattr(workflow, workflow_function)(IamDataFrame(input_file)).to_excel(
+            output_file
+        )
+    getattr(workflow, workflow_function)(IamDataFrame(input_file))
```
danielhuppmann, line 237: Less repetition if you do it like this...
```suggestion
    df =  getattr(workflow, workflow_function)(IamDataFrame(input_file))
    if output_file is not None:
        df.to_excel(output_file)
```

danielhuppmann (approved):
A few minor suggestions inline, then good to be merged!

nomenclature/cli.py
phackstock: Ah no, good point, I"ll fix that.

PR #366, opened 2024-08-12
Add parsing of data-validation yaml files

danielhuppmann:
This PR adds a class `DataValidator`, enables parsing the validation criteria from a yaml file and makes sure that the filter criteria are defined in a DataStructureDefinition (only if that dimension is defined there).


tests/test_data_validation.py
```diff
@@ -0,0 +1,59 @@
...
+                }
+            ],
+            "file": DATA_VALIDATION_TEST_DIR / "simple_validation.yaml",
+        }
+    )
+    obs = DataValidator.from_file(DATA_VALIDATION_TEST_DIR / "simple_validation.yaml")
+    assert obs == exp
+
+    dsd = DataStructureDefinition(TEST_DATA_DIR / "validation" / "definition")
+    obs.validate_with_definition(dsd) is None
```
phackstock, line 29: I think you'd need another `assert` here: 
```suggestion
    assert obs.validate_with_definition(dsd) is None
```

tests/test_data_validation.py
```diff
@@ -0,0 +1,59 @@
...
+        DATA_VALIDATION_TEST_DIR / f"validation_unknown_{dimension}.yaml"
+    )
+
+    # validating against a DataStructure with all dimensions raises
+    dsd = DataStructureDefinition(TEST_DATA_DIR / "validation" / "definition")
+    with pytest.raises(ValueError, match=match):
+        data_validator.validate_with_definition(dsd)
+
+    # validating against a DataStructure without the offending dimension passes
+    dsd = DataStructureDefinition(
```
phackstock, line 55: You're already testing a compliant `DataValidator` above in `test_DataValidator_from_file` so I'd say you can remove those lines from the test and just test for the error.

nomenclature/processor/iamc.py
```diff
@@ -14,7 +14,7 @@ class IamcDataFilter(BaseModel):
     unit: List[str] | None = None
     year: List[int] | None = None
 
-    @field_validator("*", mode="before")
+    @field_validator(*IAMC_IDX + ["year"], mode="before")
```
phackstock, line 17: You could've still kept using `"*"`, no? 
Might be easier to understand that this validator is supposed to be applied to all fields for someone who's not familiar with `IAMC_IDX`

nomenclature/processor/iamc.py
```diff
@@ -24,12 +24,13 @@ def validate_with_definition(self, dsd: DataStructureDefinition) -> None:
 
         # check for filter-items that are not defined in the codelists
         for dimension in IAMC_IDX:
-            if codelist := getattr(dsd, dimension, None) is None:
+            codelist = getattr(dsd, dimension, None)
+            if codelist is None:
                 continue
             if invalid := codelist.validate_items(getattr(self, dimension) or []):
```
phackstock, line 30: Not from this PR but you can directly give `getattr` the default/fallback:
```suggestion
            if invalid := codelist.validate_items(getattr(self, dimension, [])):
```

phackstock (approved):
Two small comments in line but already good to merge.

tests/test_data_validation.py
danielhuppmann: Well, the idea here was to test here that really only the dimensions are tested where a codelist exists in the DataStructureDefinition. Tried to explain that in the docstring, but not clear enough, I guess...

nomenclature/processor/iamc.py
danielhuppmann: This doesn't work because it would also be applied to the `upper_bound` and 'lower_bound` attributes of the class that inherits from the IamcDataFilter...

phackstock: Ah, you're right, never mind.

PR #367, opened 2024-08-13
Add check for `validate_data` consistency to `validate-project` CLI

danielhuppmann:
This PR extends the `validate-project` CLI to check the consistency of all yaml files in the `validate_data` folder.

The error output is
```
Error(s) in files for DataValidator:
Collected 2 errors:
  1. In file data/validation/validate_data/validate_unknown_region.yaml:
  Collected 1 error:
    1. The following regions are not defined in the DataStructureDefinition:
       Asia

  2. In file data/validation/validate_data/validate_unknown_variable.yaml:
  Collected 1 error:
    1. The following variables are not defined in the DataStructureDefinition:
       Final Energy|Industry
```

Could be extended to clearly state the folder/type that is being checked in the first row...


danielhuppmann:
Reworked the error log message to be more concise:

```
Collected 2 errors when checking files for 'DataValidator' (validation/validate_data):
  1. Collected 1 error when checking in file 'validation/validate_data/validate_unknown_region.yaml':
    1. The following regions are not defined in the DataStructureDefinition:
       Asia

  2. Collected 1 error when checking in file 'validation/validate_data/validate_unknown_variable.yaml':
    1. The following variables are not defined in the DataStructureDefinition:
       Final Energy|Industry
```

nomenclature/testing.py
```diff
@@ -63,35 +68,36 @@ def _check_mappings(
         raise FileNotFoundError(f"Mappings directory not found: {path / mappings}")
 
 
-def _collect_RequiredData_errors(
-    required_data_dir: Path, dsd: DataStructureDefinition
+def _collect_processor_errors(
+    path: Path, processor: Processor, dsd: DataStructureDefinition
```
phackstock, line 72: I think the type hint for processor is not correct, the way it's written right now, it would expect an instance of `Processor` not the class itself.
Since `Processor` currently only implements `apply` and not `from_file` or `validate_with_definition` we'd have to use the following:

```suggestion
    path: Path, processor: type[RequiredDataValidator] | type[DataValidator], dsd: DataStructureDefinition
```

I thought about adding `from_file` and `validate_with_definition` to `Processor` as abstract methods but `RegionProcessor` also inherits from `Processor` and implements neither function. 
I think the above is the best fix.

phackstock (approved):
Looks good to me, one small comment in line, after that good to merge

nomenclature/testing.py
danielhuppmann: Thanks, done

PR #373, opened 2024-08-30
Add preliminary alpha3 code for Kosovo

danielhuppmann:
It's becoming a nuisance in automated processing that Kosovo does not have an official ISO3166 code, so this PR adds "KOS" following the IOC.


phackstock (approved):
Maybe you could add a quick addition to the docs to state that we now have an ISO code for Kosovo?

PR #374, opened 2024-09-02
Add a `countries` attribute to the region codelist

danielhuppmann:
While working on model-registrations for common-definitions, I noticed that we've not been consistent whether to add a list of countries or ISO3-codes to clearly define model-native regions.

In favor of readability, I'd prefer to use country-names, see for example https://github.com/IAMconsortium/common-definitions/pull/118.

This PR adds an optional `countries` attribute to the **RegionCode** class, and validates any names in this list against the **nomenclature.countries** list.

As a follow-up PR, I suggest that we deprecate the `iso3_codes` attribute and instead add an `alpha_3` property that returns the countries-elements translated to ISO3 codes. This way, we have good readability and consistency in the GitHub repos and anyone still prefering to work with ISO3 codes can still comfortably work with those.


phackstock:
> As a follow-up PR, I suggest that we deprecate the iso3_codes attribute and instead add an alpha_3 property that returns the countries-elements translated to ISO3 codes. This way, we have good readability and consistency in the GitHub repos and anyone still prefering to work with ISO3 codes can still comfortably work with those.

I am on board with that suggestion. We'll just might have to update some instance workflow repositories that are currently using the `iso3_codes` attribute. That shouldn't prevent us from improving the readability though which this PR does in my opinion.

nomenclature/code.py
```diff
@@ -233,13 +233,27 @@ class RegionCode(Code):
...
-    iso3_codes: Union[List[str], str] = None
+    countries: Optional[List[str]] = None
+    iso3_codes: Optional[Union[List[str], str]] = None
+
+    @field_validator("countries")
+    def check_countries(cls, v: List[str], info: ValidationInfo) -> List[str]:
+        """Verifies that each country name is defined in `nomenclature.countries`."""
+        if invalid_country_names := set(v) - set(countries.names):
+            raise ValueError(
+                f"Region '{info.data['name']}' uses non-standard country name(s): "
```
phackstock, line 252: I'd reference the exact ISO norm that we're using, just "standard" is a little ambiguous in my opinion:

```suggestion
                f"Region '{info.data['name']}' uses non-ISO3166-1 compliant country name(s): "
```

tests/data/general-config-only/nomenclature.yaml
```diff
@@ -4,6 +4,5 @@ repositories:
 definitions:
   region:
     repository: common-definitions
-    country: true
```
phackstock, line 7: As an alternative you could leave this in and use the pin feature to pin this config to the latest pre-G20 common-definitions revision:
```yaml
repositories:
  common-definitions:
    hash: 4f77be8
definitions:
  region:
    country: true
  variable:
```

tests/data/general-config/nomenclature.yaml
```diff
@@ -5,6 +5,5 @@ repositories:
 definitions:
   region:
     repository: common-definitions
-    country: true
```
phackstock, line 8: Same as above, to fix the tests you could pin common-definitions:

```yaml
repositories:
  common-definitions:
    hash: 4f77be8
definitions:
  region:
    country: true
  variable:
```

tests/test_definition.py
```diff
@@ -51,12 +51,6 @@ def test_definition_from_general_config(workflow_folder):
             assert "Region A" in obs.region
         # imported from https://github.com/IAMconsortium/common-definitions repo
         assert "World" in obs.region
-        # added via general-config definitions
-        assert "Austria" in obs.region
```
phackstock, line 55: If you introduce the pin for common-definitions, you can bring this one back (I think)

phackstock (approved):
Couple of suggestions in-line, good to merge otherwise.

nomenclature/code.py
danielhuppmann: We do not strictly follow ISO-3166-1, we use our own adaptation, see 
https://github.com/IAMconsortium/nomenclature/blob/main/nomenclature/countries.py.

I added a link to the documentation for clarity.

tests/data/general-config-only/nomenclature.yaml
danielhuppmann: We do have a separate test for `country: true`, so I'd rather keep the tests consistent with the latest version of common-definitions to be warned about future conflicts early. There isn't so much extra benefit to the joint test of common-definitions and `country: true`.

PR #375, opened 2024-09-03
Bring back Windows tests

### LINKED ISSUE #364: Windows tests failing due to matplotlib error
As found here https://github.com/IAMconsortium/nomenclature/actions/runs/10248178276/job/28348892224:

```console
Warning: The file chosen for install of matplotlib 3.9.1 (matplotlib-3.9.1.tar.gz) is yanked. Reason for being yanked: The Windows wheels, under some conditions, caused segfaults in unrelated user code.  Due to this we deleted the Windows wheels to prevent these segfaults, however this caused greater disruption as pip then began to try (and fail) to build 3.9.1 from the sdist on Windows which impacted far more users.  Yanking the whole release is the only tool available to eliminate these failures without changes to on the user side.  The sdist, OSX wheel, and manylinux wheels are all functional and there are no critical bugs in the release.   Downstream packagers should not yank their builds of Matplotlib 3.9.1.  See https://github.com/matplotlib/matplotlib/issues/28551 for details.
```

since nomenclature does not feature matplotlib as an explicit dependency, I believe this to be a pyam issue.

### LINKED ISSUE #372: Windows tests currently (Aug 29 2024) broken
Most likely a GH error:

```console
Command ['D:\\a\\nomenclature\\nomenclature\\.venv\\Scripts\\python.exe', '-I', '-W', 'ignore', '-c', 'import sys\n\nif hasattr(sys, "real_prefix"):\n    print(sys.real_prefix)\nelif hasattr(sys, "base_prefix"):\n    print(sys.base_prefix)\nelse:\n    print(sys.prefix)\n'] errored with the following return code 103

Error output:
No Python at '"C:\hostedtoolcache\windows\Python\3.12.4\x64\python.exe'
```

as seen here: https://github.com/IAMconsortium/nomenclature/actions/runs/10617131911/job/29429187755?pr=371.

We'll disable the Windows tests for now and come back when the issue is fixed.

phackstock:
Closes #372, closes #364.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -15,7 +15,7 @@ jobs:
         shell: bash
     strategy:
       matrix:
-        os: ["macos", "ubuntu", "windows"]
+        os: ["macos", "ubuntu", "windows-latest"]
         # keep consistent with py-version badge in README.md and docs/index.rst
         python-version: ["3.10", "3.11", "3.12"]
       fail-fast: false
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -15,7 +15,7 @@ jobs:
         shell: bash
     strategy:
       matrix:
-        os: ["macos", "ubuntu"]
+        os: ["macos", "ubuntu", "windows"]
         # keep consistent with py-version badge in README.md and docs/index.rst
         python-version: ["3.10", "3.11", "3.12"]
       fail-fast: false
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -15,7 +15,7 @@ jobs:
         shell: bash
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


phackstock:
Removing the cached environment fixes the tests since the failing step is skipped.

danielhuppmann (approved):
Thanks!

phackstock:
@danielhuppmann, sorry for the confusion, I'm not sure that this one is actually fixed.
I think the tests fail if the step "Install library" actually needs to be run because the loaded cached environment is not sufficient.
This is from the latest run that succeeded:
image
I think it succeeded because I manually deleted the environment which forced a cache refresh which means that "Install library" could be skipped.
I'm not fully sure what's going on so I'd keep this PR open for now. 
Seems that we need to change a few things for making running on Windows smooth https://github.com/snok/install-poetry#running-on-windows (this is the GH action we use to set up poetry).

PR #380, opened 2024-09-10
Add Ruff

### LINKED ISSUE #369: Add ruff to nomenclature
@dc-almeida since you just added ruff to pysquirrel this would be a good opportunity to add it here as well. 
This way all (I think) of our projects follow the same standards.

dc-almeida:
Closes #369
poetry.lock was set using `poetry lock --no-update`, thus keeping package versions, only updating to include Ruff (@phackstock there are some stray mentions of black in the lock file, since it is updated by poetry I didn't touch them, but not sure if they are safe to remove)


phackstock (approved):
Good to be merged, thanks @dc-almeida.

PR #390, opened 2024-09-17
Add ErrorCollector to iso3_codes validation

### LINKED ISSUE #379: ErrorCollection for countries-validation
The validation of countries-elements in the RegionCode class raises a ValueError directly. It would be better to use the error-collection utility (rather than only raising the first problem in a region-yaml-file.

To be specific: this error https://github.com/IAMconsortium/nomenclature/blob/65f005c40787548f59a4e0d068c38fda424cad0c/nomenclature/code.py#L267

should be raised as part of an error-collection like

https://github.com/IAMconsortium/nomenclature/blob/65f005c40787548f59a4e0d068c38fda424cad0c/nomenclature/codelist.py#L219

dc-almeida:
Closes #379

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -12,6 +12,8 @@ from pydantic import (
     ValidationInfo,
 )
 
+from nomenclature.error import ErrorCollector
+
 from pyam.utils import to_list
 
 from .countries import countries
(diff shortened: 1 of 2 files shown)


phackstock (approved):
Good to be merged thanks @dc-almeida :+1:

PR #392, opened 2024-09-17
Remove missing model mapping warning

### LINKED ISSUE #358: Remove the warning of missing model mapping
This causes more harm than good, should be removed.

dc-almeida:
Closes #358


phackstock (approved):
Thanks @dc-almeida, hopefully this will avoid some confusion in the future.

PR #396, opened 2024-09-19
Filter external repos

### LINKED ISSUE #326: Allow attribute filtering in nomenclature.yaml for importing definitions form external repo
When importing from an external repository we should be able to filter by attributes. This way we don't import the whole definition if it's not needed.
The first three use cases that come to mind are:

* filtering by hierarchy for regions
* filtering by `tier` for variables
* filtering any dimension by name 

The question would be how to integrate this into the existing `nomenclature.yaml` structure.
I've tried a few things now and this is my current favorite:

```yaml
repositories:
  common-definitions:
definitions:
  region:
    repository-filters:
      hierarchy: R5
  variable:
    repository-filters:
      name: Final Energy*
    country: true
```

This would import all R5 regions from common-definitions and all variables starting with `Final Energy*`.

The above format would also allow for more complex filtering such as:

```yaml
repositories:
  common-definitions:
  legacy-definitions:
    url: https://github.com/IAMconsortium/legacy-definitions.git/
definitions:
  variable:
    repository-filters:
      - repository: common-definitions
        tier: 1
      - name: Final Energy*
    country: true
```

here we have multiple filters for the variable dimension:
1. We take all variables from **common-definitions** that have the attribute **tier** with the value **1**. 
2. We take all variables from common-definitions and legacy-definitions (no _repository_ filter) that match the pattern `Final Energy*`

Would love to hear your thoughts @danielhuppmann, @dc-almeida.

phackstock:
Closes #326.

This PR adds the feature to filter external repositories using `include` and `exclude` filters.
Any number of filters combining any attributes can be defined, example:

```yaml
repositories:
  common-definitions:
definitions:
  variable:
    repository:
      name: common-definitions
      include:
        - name: [Primary Energy*, Final Energy*]
        - name: "Population*"
          tier: 1
      exclude:
        - name: "Final Energy|*|*"
  region:
    repository:
      name: common-definitions
      include:
        - hierarchy: [R5, R10]
```

### For the variable section
in the example above we are including:
1. All variables starting with *Primary Energy* or *Final Energy*
2. All variables starting with *Population* **and** with the tier attribute equal to 1

From this list we are then **excluding** all variables that match "Final Energy|*|*".
This means that the final resulting list will contain no Final Energy variables with
three or more levels.

### For the region section
we are taking only R5 and R10 regions.

## Changes

One of the changes that I have made is that all repositories in the definition section need to have the `name` key-word. 
This would be a breaking change so all workflow repos that use external repositories would need to be updated. 
I'd be happy to streamline it again so that just the name is allowed but for code simplicity I opted against that for now.


docs/user_guide/config.rst
```diff
@@ -49,19 +49,55 @@ multiple external repositories can be used as the example below illustrates for
...
+  definitions:
+    variable:
+      repository:
+        name: common-definitions
+        include:
+          - name: [Primary Energy*, Final Energy*]
+          - name: "Population*"
+            tier: 1
+        exclude:
+          - name: "Final Energy|*|*"
```
danielhuppmann, line 91: This is a misleading example - it seems to show only level-2 exclusion but it excludes all variables at level 2 or below. Also, pyam excludes the filter-String from the level-calculation. Better to use **pyam.utils.pattern_match** for consistency with pyam-filtering and support the level-argument explicitly.
```suggestion
          - name: "Final Energy|*"
             level: "1+"
```

nomenclature/config.py
```diff
@@ -126,12 +181,22 @@ def repos(self) -> dict[str, str]:
         }
 
 
+class MappingRepository(BaseModel):
+    name: str
```
danielhuppmann, line 185: Das The mapping will also have to inherit the region-filters, right? Otherwise, a model could map to a region that is not included in the **DataStructureDefinition**.

tests/data/nomenclature_configs/multiple_repos_for_mapping.yaml
```diff
@@ -5,5 +5,5 @@ repositories:
     url: https://github.com/IAMconsortium/legacy-definitions.git/
 mappings:
   repositories:
-    - common-definitions
-    - legacy-definitions
+    - name: common-definitions
```
danielhuppmann, line 8: Is the name now required? Could cause a lot of headache of that is now mandatory…

danielhuppmann (commented):
A few suggestions inline.

As with my other review, I’d strongly advise against using **fnmatch**.

docs/user_guide/config.rst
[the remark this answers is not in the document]
phackstock: Seems that this is subjective then. For me it was totally clear that this excludes anything level 2 and beyond.
I can see your point though about this being ambiguous

tests/data/config_filter/nomenclature.yaml
```diff
@@ -0,0 +1,17 @@
+repositories:
```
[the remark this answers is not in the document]
phackstock: Sure, good idea.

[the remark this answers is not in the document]
phackstock: #399 implements a cleanup of the test data folder, once that PR is merged, I'll rebase this one

tests/data/nomenclature_configs/multiple_repos_for_mapping.yaml
danielhuppmann: Yes, better to add a field-validator that translates a string to `name: {value}` to avoid breaking current workflows

tests/data/config_filter/nomenclature.yaml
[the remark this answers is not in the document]
phackstock: Closing since #399 is merged.

docs/user_guide/config.rst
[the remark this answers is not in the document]
phackstock: I'll change the filtering logic to mirror that of pyam. This way we'll get a consistent API across both packages.

nomenclature/config.py
phackstock: That's already taken care of here: https://github.com/phackstock/nomenclature/blob/27878178da182a48b85b524e857d2c80688b644a/nomenclature/processor/region.py#L485-L496. The external mapping repository contents are parsed like any other mapping file and are therefore checked for region consistency.

docs/user_guide/config.rst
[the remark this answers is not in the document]
danielhuppmann: FYI, I added a new filter-argument *depth* in pyam (see https://github.com/IAMconsortium/pyam/pull/886), because *level* changes behavior depending on the *variable*. string, so can lead to confusion.

I suggest to use *depth* in any examples.

danielhuppmann:
Quick clarifying question, one can only have one **include** and one **exclude** statement per repo-import, right?

phackstock:
> Quick clarifying question, one can only have one include and one exclude statement per repo-import, right?

Correct. You can put multiple filters in both the include and exclude though.

docs/user_guide/config.rst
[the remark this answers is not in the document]
phackstock: Thanks for the info. After looking at the PR for a couple of minutes and consulting the pyam documentation I think I understand the difference between the two. Happy to use depth for this then.

If I understand correctly, using _depth_ is the preferred option. If so I'd suggest to update the pyam tutorials (https://pyam-iamc.readthedocs.io/en/stable/tutorials/pyam_first_steps.html#Filtering-by-variables-and-levels) to use _depth_.
In addition, I'd also suggest removing level as an argument since it can lead to confusion to have two very similar but subtly different options. Especially without an example to illustrate the difference (at least I didn't find an example in the docs).

phackstock:
@danielhuppmann, I added a _depth_ option now so you can match for the given level of variable depth. 
I didn't implement the "1-" option, since it would make the code a lot less read- and maintainable and using fnmatch we already have that functionality. 
Say you want to select "Final Energy" with the depth "1-" option, in pyam logic. You'd use

```yaml
...
      include:
        - name: Final Energy*
          depth: [0, 1]
```

which I think is plenty readable. 
Since this option uses fnmatch, I chose to keep it.

docs/user_guide/config.rst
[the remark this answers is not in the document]
phackstock: Updated the example to use _depth_

phackstock:
@danielhuppmann, as discussed bilaterally, I changed the string pattern matching to use the pyam character escape and regex.
Should be good now.

docs/user_guide/config.rst
```diff
@@ -57,11 +57,46 @@ multiple external repositories can be used as the example below illustrates for
...
+        name: common-definitions
+        include:
+          - name: [Primary Energy*, Final Energy*]
+          - name: "Population*"
+            tier: 1
+        exclude:
+          - name: "Final Energy|Industry*"
+            depth: 2
+
+If a filter is being used for repositories, the *name* keyword **must be used**.
```
danielhuppmann, line 91:
```suggestion
If a filter is being used for repositories, the *name* attribute **must be used**
for the repository.
```

nomenclature/config.py
```diff
@@ -11,29 +12,94 @@
     field_validator,
     model_validator,
     ConfigDict,
-    BeforeValidator,
 )
+from nomenclature.code import Code
+from pyam.str import escape_regexp
+
+
+class RepositoryWithFilter(BaseModel):
```
danielhuppmann, line 20: The "with-filter" seems important now because of the changes, but it isn't really the defining characteristic of this class - and there is no **RepositoryWithoutFilter** class.
```suggestion
class CodeListFromRepository(BaseModel):
```

tests/data/region_processing/external_repo_test/nomenclature.yaml
```diff
@@ -7,8 +7,11 @@ repositories:
     hash: cb85704
 definitions:
   region:
-    repository: common-definitions
+    repository:
+      name: common-definitions
   variable:
-    repository: common-definitions
+    repository:
```
danielhuppmann, line 13: Would it make sense to keep the short-hand notation here (without the *name* attribute) to make sure that this implementation continues to work? (Please add a comment if you revert).

danielhuppmann (commented):
Thanks, looks good, a few minor suggestions inline.

One larger question: if an imported model-mapping includes a common region that is filtered out in the region-codelist, the initialization of the region-processor will fail, right? (That's ok, just make sure to create a follow-up issue so that this can be tackled later).

tests/data/region_processing/external_repo_test/nomenclature.yaml
phackstock: Good point, I restored the original notation on all other files but missed that one. It's fixed now.

phackstock:
> One larger question: if an imported model-mapping includes a common region that is filtered out in the region-codelist, the initialization of the region-processor will fail, right? (That's ok, just make sure to create a follow-up issue so that this can be tackled later).

Looking at the code it looked fine to me but just to be sure I added a test in 60d77ce. I import all the mappings from common-definitions and only load the world region. It fails as expected.

danielhuppmann (approved):
Nice, thank you!

PR #397, opened 2024-09-19
Wildcard matching

### LINKED ISSUE #294: Allow wildcard variables
Per a discussion with @volker-krey and @phackstock:

Previous projects added a "counter" to capacity- and capital-cost variables to allow reporting of different types of plants, e.g., "Capacity|Electricity|Biomass|1", with a mapping of the number to a particular type given in a comments sheet.

This is a not very elegant approach and not user-friendly when viewing the data via the Scenario Explorer.

It would be preferable to explicitly allow wildcards in a variable name, e.g., `Capacity|Electricity|Biomass|*` - and the nomenclature-validation would allow all variables starting with *Capacity|Electricity|Biomass*. This way, each modelling team could define the types individually, using either numbering or more useful description like "CCGT" directly in the variable name.

phackstock:
Closes #294.

While this PR is ready in principle (minus documentation), I'd see it as a basis for discussion if we actually want this feature.


dc-almeida (approved):
Noted. Looks good to me!

phackstock:
Following the discussion in ScenarioMIP, I think we might want to have full regex capabilities.
The requirements there should allow for submitting the standard scenarios with an **optional** `_[a-z]` suffix. As far as I can see this is not (easily) possible with the fnmatch that this PR uses.

tests/data/codelist_wildcard/scenario/scenario.yaml
```diff
@@ -0,0 +1 @@
+- scen_*
```
danielhuppmann, line 1: Could make sense to start putting codelist-validation test data into an own subfolder

nomenclature/codelist.py
```diff
@@ -116,7 +117,7 @@ def validate_items(self, items: List[str]) -> List[str]:
         list
             Returns the list of items that are **not** defined in the codelist
         """
-        return [c for c in items if c not in self.keys()]
+        return [item for item in items if not any(fnmatch(item, code) for code in self)]
```
danielhuppmann, line 120: **fnmatch** does not work work square brackets as discussed in https://github.com/IAMconsortium/pyam/issues/542, so I suggest to use **pyam.utils.pattern_match** for consistent behaviour with pyam-filtering.

tests/test_validation.py
```diff
@@ -92,3 +84,17 @@ def test_validation_with_custom_dimension(simple_df):
...
+    definition = DataStructureDefinition(
+        TEST_DATA_DIR / "codelist_wildcard",
+        dimensions=["scenario"],
+    )
+
+    assert definition.validate(simple_df) is None
+
+    with pytest.raises(ValueError, match=MATCH_FAIL_VALIDATION):
+        definition.validate(
+            simple_df.rename(scenario={"scen_a": "does not match scen_*"})
```
danielhuppmann, line 99: The new scenario name looks like the expected error message, I find this confusing.
```suggestion
            simple_df.rename(scenario={"scen_a": "foo"})
```

danielhuppmann (commented):
I think the feature is useful, but I would not use **fnmatch** based on earlier experience (see inline comment).

about regexp, maybe we can abuse regexp-notation (as a follow-up PR) like

```yaml
- scen_a  # direct match
- scen_*  # wildcard match
- r’scen_[.]’  # regexp match
```

tests/data/codelist_wildcard/scenario/scenario.yaml
phackstock: I cleaned up the complete test data folder in #399. Once that PR is merged, I'll rebase this one and move this file.

phackstock:
Updated to using `pyam.utils.pattern_match`.

danielhuppmann (approved):
Looks good, thanks!

Issue #406, opened 2024-10-03
Loading RegionProcessor fails on Windows for regions with non-ASCII UTF8 characters


korsbakken:
When instantiating a `RegionProcessor` object from `RegionProcessor.from_directory`, pydantic raises a ValidationError for regions with non-ASCII characters, stating they are not in the RegionCodeList even though they in fact are defined with the exact same characters in the DSD region codelist.

Screenshot of example error message attached (from another user, sorry, I can't test this on Windows myself).
![image](https://github.com/user-attachments/assets/2c87236b-b24a-470e-9a83-b8321eb9466b)

I think the problem is in the file open statement in `RegionAggregationMapping.from_yaml` here:
https://github.com/IAMconsortium/nomenclature/blob/30fb5b14d34da4c8905f54df4d16f2a38a24abc3/nomenclature/processor/region.py#L271

It does not specify the encoding. Contrast that with `CodeList.._parse_codelist_dir` here: https://github.com/IAMconsortium/nomenclature/blob/30fb5b14d34da4c8905f54df4d16f2a38a24abc3/nomenclature/codelist.py#L294

The latter explicitly specifies utf8 encoding, which creates an inconsistency if the system doesn't use utf8 as the default encoding.

I would suggest to explicitly set utf8 encoding in `RegionAggregationMapping.from_yaml` (and maybe we should check that it's crystal clear from the documentation that yaml files must be in utf8). Or, if there is some reason not to, instead remove it from elsewhere for consistency.

I can submit a PR, but would be great to get feedback on any constraints, and whether this might touch on broader issues that I'm not aware of.

korsbakken:
Just an update: There are numerous places in the code where text files are read both with and without explicit encoding. I think the bigger question here is: Do we want to 1) force utf-8 as the encoding for all yaml files, or should we 2) let the user (or codelist/mapping creators) decide freely and live with the consequences?

Option 1 would make the code more robust and behavior more consistent across platforms, but forces Windows users to check that their editor saves yaml files in UTF8 rather than latin1 or whatever their language settings prefer.

Option 2 makes it easier for Windows users to create and edit yaml files with their text editor of choice, but can lead to unpredictable behavior for non-ASCII region names, especially if run on other platforms.

I assume option 1 is the only one that really makes sense, especially considering the chaos that option 2 can cause when mixing yaml files from different sources. In which case we should specify `encoding="utf8"` in all `open` statements for text files and text streams.

[...]

PR #407, opened 2024-10-03
Added explicit encoding to all `open` statements.

korsbakken:
Added `encoding="utf-8"` to all calls to `open` that open a yaml file or stream, where encoding was not already specified. This is intended to deal with inconsistencies that arise from not specifying or inconsistently specifying encoding on platforms where utf-8 is not necessarily the default text file encoding.

I have only checked for calls to `open`. There might be other types of calls that could be affected by platform-dependent encoding defaults and that don't specify encoding, such as calls through `pyam` or `pandas` methods, code that deals with Excel files. But especially in the latter case, it's maybe less clear that we should enforce UTF-8.

This PR addresses #406 and potentially resolves it. Unfortunately I don't have a suitable Windows system to test on.


phackstock (commented):
Thanks a lot @korsbakken for catching and fixing this oversight.
With regards to Excel, we should be fine since the information about the encoding used should be part of the file itself.
Since we run all of our tests in GitHub on Windows, MacOS and Ubuntu you can just add a test for reading a character that would require utf-8. This way we can verify that it at least works on the GitHub actions version of Windows. It might also be interesting to see if we can reproduce the error in a GH action.

PR #410, opened 2024-10-08
Check for forbidden characters in attributes

### LINKED ISSUE #401: Check for forbidden characters in attributes
Issue https://github.com/IAMconsortium/common-definitions/pull/133 identified a problem where tags in code-attributes (description in that case) were not replaced because they were not misplaced.

We already have a validation of stray tags in code-names, see 
https://github.com/IAMconsortium/nomenclature/blob/9314fef41229a8d5cd855ea3e3b5ad77a9dd83fd/nomenclature/codelist.py#L48

It would be very helpful to extend that test to all attributes that are (lists of) strings and also guard against other problematic characters, like `"`, `:`, `;`.

FYI @phackstock @orichters

dc-almeida:
Closes #401
Expands on stray tag check in code names, adding checks for double-quotes, and for all code attributes.
Recursively checks list and dict attributes for strings.
Tests expanded.
Validation function name and test name changed.


dc-almeida:
After bilateral discussion, proposed fix was changed

PR #411, opened 2024-10-09
Extend stray tag check

dc-almeida:
Partial resolution of #401.
Expands on stray tag check in code names, adding checks for brackets for all code attributes.
Recursively checks list and dict attributes for strings.
Tests expanded.


tests/data/codelist/stray_tag/char_in_str/variables.yaml
```diff
@@ -6,4 +6,4 @@
     unit: EJ/yr
 - Share|{Fuel}:
     definition: Share of {Fuel} in the total primary energy mix
-    unit:
+    unit: EJ/yr
```
danielhuppmann, line 9: Share should not be given in EJ (possibly in %).

tests/test_codelist.py
```diff
@@ -213,12 +213,24 @@ def test_to_csv():
 
 
 def test_stray_tag_fails():
-    """Check that typos in a tag raises expected error"""
+    """Check that stray brackets from, e.g. typos in a tag, raises expected error"""
```
danielhuppmann, line 216: This could be nicely implemented as a pytest-mark to reduce repetition, see https://github.com/IAMconsortium/nomenclature/blob/f5289357e4d58503d3deac9137408940bd9c3e10/tests/test_core.py#L179

danielhuppmann (commented):
Nice, see a suggestion for cleaner test-code inline.

Also, I suggest to also add a test-case where a stray tag is in a attribute-string, like
```yaml
- Primary Energy:
     definition: Total primary energy { consumption
     unit: EJ/yr
```

dc-almeida:
> Nice, see a suggestion for cleaner test-code inline.
> 
> Also, I suggest to also add a test-case where a stray tag is in a attribute-string, like
> 
> ```yaml
> - Primary Energy:
>      definition: Total primary energy { consumption
>      unit: EJ/yr
> ```

Initially I had added one, but removed it since the original test does so by checking for the stray tag in the code name. Can easliy add another more explicit one.

nomenclature/codelist.py
```diff
@@ -46,13 +46,28 @@ def __eq__(self, other):
...
-                    " Check if the tag was spelled correctly."
-                )
+        """Check that no stray tags are left in codes after tag replacement"""
+        forbidden = ["{", "}"]
+
+        def _check_string(value):
+            if isinstance(value, str):
+                if any(char in value for char in forbidden):
+                    raise ValueError(
+                        f"Unexpected bracket in codelist: {code.name}."
```
danielhuppmann, line 56:
```suggestion
                        f"Unexpected bracket in {self.name} '{code.name}'."
```

tests/test_codelist.py
```diff
@@ -212,13 +212,19 @@ def test_to_csv():
...
 
 
-def test_stray_tag_fails():
-    """Check that typos in a tag raises expected error"""
-
-    match = r"Unexpected {} in codelist: Primary Energy\|{Feul}"
+@pytest.mark.parametrize(
+    "subfolder, match",
+    [
+        ("char_in_str", r"Unexpected bracket in codelist: Primary Energy\|{Feul}"),
```
danielhuppmann, line 218:
```suggestion
        ("char_in_str", r"Unexpected bracket in variable 'Primary Energy\|{Feul}'"),

```

tests/test_codelist.py
```diff
@@ -212,13 +212,19 @@ def test_to_csv():
...
+        ("char_in_list", r"Unexpected bracket in codelist: Share\|Coal"),
```
danielhuppmann, line 219:
```suggestion
        ("char_in_list", r"Unexpected bracket in variable 'Share\|Coal'"),
```

tests/test_codelist.py
```diff
@@ -212,13 +212,19 @@ def test_to_csv():
...
+        ("char_in_dict", r"Unexpected bracket in codelist: Primary Energy"),
```
danielhuppmann, line 220:
```suggestion
        ("char_in_dict", r"Unexpected bracket in variable 'Primary Energy'"),
```

danielhuppmann (approved):
One more little suggestionfor better error messages, then good to be merged.

PR #412, opened 2024-10-10
Allow countries attribute as string

### LINKED ISSUE #398: `countries` attribute doesn't allow for single country string input
https://github.com/IAMconsortium/nomenclature/blob/73715e4da38f109d07c17cc1d3c5cb198f88969e/nomenclature/code.py#L246-L250

A YAML file with `countries: <country_name>` raises a ValidationError, unlike the `iso3_codes` attribute (`iso3_codes: <iso3_code>` is valid)

dc-almeida:
Closes #398


danielhuppmann (commented):
I think it's more user-friendly to always have the `countries` as a list internally - otherwise, users wanting to work with that attribute have to check whether it's a list every time in their scripts...

See the implementation as suggestions inline.

danielhuppmann (approved):
Sweet!

PR #413, opened 2024-10-14
Update pysquirrel dependency

dc-almeida:
For a minor version bump to resolve pysquirrel errors due to NUTS data file hash update


danielhuppmann (approved):
Thanks!

PR #415, opened 2024-10-15
Read dimensions other than `variable` and `region` from external repo

### LINKED ISSUE #414: Dimensions other than `region` and `variable` are not read from repositories listed in nomenclature.yaml
When creating a new `DataStructureDefinition` object from a path that includes external repositories in a `nomenclature.yaml` file, only the `region` and `variable` dimensions get read from the repository. If I list, e.g., `model` and `scenario` in the `dimensions` list in nomenclature.yaml, I get a ValueError with message  `"Empty codelist: model, scenario"`. Those dimensions *are* present with valid yaml files in the repository, and are read without any problems if I clone the repository to my computer and read its `definitions` folder locally.

I have pasted the nomenclature.yaml file at the bottom. It will reproduce the problem if put in a folder with empty `definitions` and `mappings` subdirectories, and read with `dsd = nomenclature.DataStructureDefinition('./definitions')`.

I'm not fluent enough in pydantic to diagnose the problem 100% certainly or know for sure how to fix it, but I think the problem lies in the definition of `config.DataStructureConfig`, which defines the `definitions` field in [`config.NomenclatureConfig`](https://github.com/IAMconsortium/nomenclature/blob/829844a776d06bb618bdd469a8dbc3219c80c5c2/nomenclature/config.py#L157-L161). The definition looks suspiciously like it only defines `region` and `variable` as valid fields, which I suspect means that the other dimensions under `definitions` in `nomenclature.yaml` get discarded. Relevant part of the code here:

https://github.com/IAMconsortium/nomenclature/blob/829844a776d06bb618bdd469a8dbc3219c80c5c2/nomenclature/config.py#L115-L139

I'd greatly appreciate if someone who is familiar with the pydantic definitions and how they are processed can take a look at this relatively soon. At the moment it's a roadblock for deploying a validation tool in an ongoing project.

Here is the content of the nomenclature.yaml file I used:
```
repositories:
  iamcompact-nomenclature-definitions:
    url: https://github.com/ciceroOslo/iamcompact-nomenclature-definitions.git
    release: validation-ui
dimensions:
  - variable
  - region
  - model
  - scenario
definitions:
  variable:
    repository: iamcompact-nomenclature-definitions
  region:
    repository: iamcompact-nomenclature-definitions
  model:
    repository: iamcompact-nomenclature-definitions
  scenario:
    repository: iamcompact-nomenclature-definitions
mappings:
  repository: iamcompact-nomenclature-definitions
```

The error traceI get is the following:
```
File ~/src/repos/iamcompact/iamcompact-nomenclature/.venv/lib/python3.12/site-packages/nomenclature/definition.py:81, in DataStructureDefinition.__init__(self, path, dimensions)
     76     self.__setattr__(
     77         dim, codelist_cls.from_directory(dim, path / dim, self.config)
     78     )
     80 if empty := [d for d in self.dimensions if not getattr(self, d)]:
---> 81     raise ValueError(f"Empty codelist: {', '.join(empty)}")

ValueError: Empty codelist: model, scenario
```

If I enter debug mode, I get the following content in `self.config.dict()`:
```
{'dimensions': ['variable', 'region', 'model', 'scenario'],
 'repositories': {'iamcompact-nomenclature-definitions': {'url': 'https://github.com/ciceroOslo/iamcompact-nomenclature-definitions.git',
   'hash': None,
   'release': 'validation-ui',
   'local_path': PosixPath('iamcompact-nomenclature-definitions')}},
 'definitions': {'region': {'dimension': 'region',
   'repositories': {'iamcompact-nomenclature-definitions'},
   'country': False,
   'nuts': None},
  'variable': {'dimension': 'variable',
   'repositories': {'iamcompact-nomenclature-definitions'}}},
 'mappings': {'repositories': {'iamcompact-nomenclature-definitions'}}}
```

In other words, only `variable:` and `region:` have been read from the `definitions` section of the config file.

korsbakken:
This PR fixes #414, as far as I can see. All tests pass, the error cited in #414 goes away, and the resulting `DataStructureDefinition` object has the expected dimensions.

If possible, it would be great if this PR could be merged as soon as possible, since it's a bug that can cause significant problems. Some possible items that might be considered in the future (none of which I have the capacity to work on in the next few months):

* Add a test. A proper test for this issue would probably be an integration test, which tries to read from a test repo that has all the valid dimensions and a non-empty codelist for each them, and checks that the resulting `DataStructureDefinition` object has all the expected dimensions and the expected codes in each codelist. As far as I can see, there is no such test currently?
* This PR still hardcodes the dimensions to be in `{'region', 'variable', 'model', 'scenario'}`. The question is whether other arbitrary dimensions should be allowed. But at the moment, I think that would cause issues multiple other places in the code, so probably goes far beyond the scope of this PR.

### DIFF
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -122,12 +122,12 @@ class DataStructureConfig(BaseModel):
 
     """
 
-    region: Optional[RegionCodeListConfig] = Field(default_factory=RegionCodeListConfig)
-    variable: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
     model: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
     scenario: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
+    region: Optional[RegionCodeListConfig] = Field(default_factory=RegionCodeListConfig)
+    variable: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
 
-    @field_validator("region", "variable", "model", "scenario", mode="before")
+    @field_validator("model", "scenario", "region", "variable", mode="before")
     @classmethod
     def add_dimension(cls, v, info: ValidationInfo):
         return {"dimension": info.field_name, **v}
@@ -136,7 +136,7 @@ class DataStructureConfig(BaseModel):
     def repos(self) -> dict[str, str]:
         return {
             dimension: getattr(self, dimension).repositories
-            for dimension in ("region", "variable", "model", "scenario")
+            for dimension in ("model", "scenario", "region", "variable")
             if getattr(self, dimension).repositories
         }
 
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -127,7 +127,7 @@ class DataStructureConfig(BaseModel):
     model: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
     scenario: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
 
-    @field_validator("region", "variable", mode="before")
+    @field_validator("region", "variable", "model", "scenario", mode="before")
     @classmethod
     def add_dimension(cls, v, info: ValidationInfo):
         return {"dimension": info.field_name, **v}
@@ -136,7 +136,7 @@ class DataStructureConfig(BaseModel):
     def repos(self) -> dict[str, str]:
         return {
             dimension: getattr(self, dimension).repositories
-            for dimension in ("region", "variable")
+            for dimension in ("region", "variable", "model", "scenario")
             if getattr(self, dimension).repositories
         }
 
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -124,6 +124,8 @@ class DataStructureConfig(BaseModel):
 
     region: Optional[RegionCodeListConfig] = Field(default_factory=RegionCodeListConfig)
     variable: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
+    model: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
+    scenario: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
 
     @field_validator("region", "variable", mode="before")
     @classmethod


danielhuppmann (commented):
I'm surprised that line 139 doesn't have to be changed as well?

nomenclature/config.py
```diff
@@ -124,8 +124,10 @@ class DataStructureConfig(BaseModel):
 
     region: Optional[RegionCodeListConfig] = Field(default_factory=RegionCodeListConfig)
     variable: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
+    model: Optional[CodeListConfig] = Field(default_factory=CodeListConfig)
```
danielhuppmann, line 127: Please always sort as `[model, scenario, region, variable]` for consistency.

korsbakken:
> I'm surprised that line 139 doesn't have to be changed as well?

Yes, it does for crossing t's and dotting i's, thanks for flagging that. It doesn't really affect the final `DataStructureDefinition` object, but it causes the `DataStructureConfig` object to lack those dimensions in its `repos` attribute. It doesn't look like that issue propagates to anywhere in this case, but it might cause problems in other circumstances?

Also, I added the extra dimensions in the `add_dimension` field validator. Again, I don't have a lot of pydantic expertise. But I assume not including them there might mean that formatting errors under `model` and `scenario` in the yaml file might not be caught.

phackstock (approved):
Thanks a lot for the useful addition @korsbakken. 
To address the two points you mentioned:
1. I agree that a test with all dimensions would be great. Normally I would insist on adding a test now, but since this feature expands on an existing, tested feature, it's fine to be merged from my side. We should open an issue though for adding a test so that we don't forget.
2. In the `DataStructureDefinition` class we are actually using what you are suggesting in that we create "dimension" attributes on the fly to allow for arbitrary dimensions. I am not super happy with this since it's potentially confusing for people reading the code and it's not compatible with type checkers. I prefer the explicit listing of the dimensions in the class attributes.
What you could do is replace the explicit listing in the validator with `"*"`. Although I'm also not sure that's the best idea. Following the zen of python "explicit is better than implicit", I'd be fine to keep it.

danielhuppmann (approved):
Thanks @korsbakken, looks good!

PR #418, opened 2024-10-16
Check for illegal characters in codelists

### LINKED ISSUE #401: Check for forbidden characters in attributes

dc-almeida:
Closes #401, updates #411

The way it's implemented, it does not check external repos for illegal characters, thus being specific to the repo where the config is defined.
This was tested previously with common-definitions, as some variables contain possibly problematic characters (see [here](https://github.com/IAMconsortium/common-definitions/blob/3b75bb946658e55affe4108ddcaac281058832cd/definitions/variable/sdg-indicators/sdg.yaml#L432)).

By checking if each code has a populated `repository` attribute, we identify and skip the external repo codes from the code lists.
I think the Pydantic validation of the `illegal_characters` field is straightforward, but can add a test for it if deemed necessary.

Caveat is that the stray tag check was previously a Pydantic validation, and now is (must) be called explicitly. If a better suggestion comes up, happy to implement it!


tests/data/codelist/illegal_chars/char_in_external_repo/nomenclature.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - variable
+repositories:
+  common-definitions:
+    url: https://github.com/IAMconsortium/common-definitions.git/
+    hash: 3b75bb9
+definitions:
+  variable:
+    repository:
+      - common-definitions
+illegal_characters: ['"' , ";"]  # these are known to be present in common-definitions variables
```
phackstock, line 11: I'm not sure that this should be a project specific configuration. I would just extend the `forbidden` list in the illegal character check.
Maybe there's a good reason to make it project specific though.

nomenclature/codelist.py
```diff
@@ -364,6 +335,35 @@ def read_excel(cls, name, source, sheet_name, col, attrs=None):
...
+                for v in value.values():
+                    _check_string(v)
+            elif isinstance(value, list):
+                for item in value:
+                    _check_string(item)
+
+        for code in self.mapping.values():
+            if not code.repository:
+                for attr in code.model_fields:
+                    if attr not in ["file", "repository"]:
```
phackstock, line 363: You're already excluding any codes with a not None repository attribute, so it should be fine to just filter out "file" here"
```suggestion
                    if attr != "file":
```

nomenclature/codelist.py
```diff
@@ -364,6 +335,35 @@ def read_excel(cls, name, source, sheet_name, col, attrs=None):
...
+                        value = getattr(code, attr)
+                        _check_string(value)
```
phackstock, line 365: This could be one line
```suggestion
                        _check_string(getattr(code, attr))
```

nomenclature/codelist.py
```diff
@@ -364,6 +335,35 @@ def read_excel(cls, name, source, sheet_name, col, attrs=None):
...
+
+        def _check_string(value):
+            if isinstance(value, str):
+                if any(char in value for char in forbidden):
+                    raise ValueError(
+                        f"Unexpected character in {self.name}: '{code.name}'."
+                        " Check for illegal characters and/or if tags were spelled correctly."
+                    )
+            elif isinstance(value, dict):
+                for v in value.values():
```
phackstock, line 354: I know you copied this over but now that I look at it, I'm thinking we might want to validate both keys and values of a dictionary.

nomenclature/codelist.py
```diff
@@ -364,6 +335,35 @@ def read_excel(cls, name, source, sheet_name, col, attrs=None):
...
+    ) -> Dict[str, Code]:
+        """Check that no illegal characters are left in codes after tag replacement"""
+        forbidden = ["{", "}"]
+        if config and config.illegal_characters:
+            forbidden += config.illegal_characters
+
+        def _check_string(value):
+            if isinstance(value, str):
+                if any(char in value for char in forbidden):
+                    raise ValueError(
```
phackstock, line 349: If you have a number of illegal errors, this can become extremely frustrating to fix because you get the errors one by one.
You can use the `ErrorCollector` class for this. There are a bunch of examples in nomenclature on how to use it.

nomenclature/codelist.py
```diff
@@ -364,6 +335,35 @@ def read_excel(cls, name, source, sheet_name, col, attrs=None):
 
         return cls(name=name, mapping=mapp)
 
+    def check_illegal_characters(
+        self, config: NomenclatureConfig = None
+    ) -> Dict[str, Code]:
+        """Check that no illegal characters are left in codes after tag replacement"""
+        forbidden = ["{", "}"]
```
phackstock, line 342: As mentioned below as well, I would just extend the list of forbidden characters directly. Unless, of course there is a need for project specific banned characters.

phackstock (commented):
Thanks for the PR @dc-almeida, couple of comments below.
Some more general questions from my side:
1. Should we extend the illegal characters to all attributes or just the name of a code? The reason I'm asking is that I think for the definition of a variable, it might be useful to use punctuation such as `:`, `;` or even quotes.
2. Should the illegal characters be project specific, i.e. be part of the nomenclature config? In my opinion, the point of nomenclature is to establish standards for variable names, etc... so having project specific illegal characters defeats that point.
3. I see the point of limiting the illegal character check for only the current repo. On the other hand, it would still affect the codelists so I'd vote for being more strict and search everything that's used in a project, including external repositories.

nomenclature/config.py
```diff
@@ -159,9 +159,15 @@ class NomenclatureConfig(BaseModel):
     repositories: dict[str, Repository] = Field(default_factory=dict)
     definitions: DataStructureConfig = Field(default_factory=DataStructureConfig)
     mappings: RegionMappingConfig = Field(default_factory=RegionMappingConfig)
+    illegal_characters: None | list[str] = None
```
phackstock, line 162: One more meta comment on the default value of iterable attributes (lists, dicts and sets). Recently, I've liked using an empty version instead of `None`. Reason being, that it's nicer to handle downstream since it saves you from implementing a bunch of checks if an attribute is None.
The whole config class is written around that idea which makes the code easier to read and write.
As an example, above you have the following check

```python
        if config and config.illegal_characters:
            forbidden += config.illegal_characters
```

if you used an empty list as a default you don't need the second check in the if:
```python
        if config and config.illegal_characters:
            forbidden += config.illegal_characters
```
since worst case you're adding an empty list.

As I said, the nomenclature config class is written around this idea of avoiding `None` as a default. This means that as a default for the config in the `check_illegal_characters` function, you can (and should in my opinion) use `NomenclatureConfig`.

This way your code can be further shortened to:

```python
    forbidden += config.illegal_characters
```
and you're actually saving yourself a level of nesting, making the code even nicer to read. 
If you look at how check_illegal_characters is called you can see that self.config can actually never be `None` so you don't even need a default value.

The reason that I use the `default_factory` option with pydantic has to do with the caveats of using mutable default values.

danielhuppmann:
FYI, the idea for this PR came from @orichters via https://github.com/IAMconsortium/common-definitions/issues/138. And the current implementation with making the illegal-characters project/repo-specific was to avoid having to go through a dozen legacy projects and manually cleaning up stuff...

phackstock:
@danielhuppmann, ah thanks for the clarification. This would mean though that we'd have to copy-paste these from now on standard illegal characters into every `nomenclature.yaml`, which is also not ideal.

phackstock:
We could flip it around, fix the illegal characters and introduce a `ignore illegal characters = False/True` (`False` being the default) switch. We would set this to `True` for all the legacy repos. This way going forward, we'd have a consistent style and don't have to keep on copy-pasting the same illegal characters to every repository.

danielhuppmann:
> We could flip it around, fix the illegal characters and introduce a `ignore illegal characters = False/True` (`False` being the default) switch. 

I like that idea but I’d again flip it around so that we can also set a specific list of characters (maybe we want to make $ illegal later).

So an argument “check-illegal-characters” that is True by default, can be set to False, or take a list of characters.

phackstock:
> So an argument “check-illegal-characters” that is True by default, can be set to False, or take a list of characters.

I'd split that into two attributes `check-illegal-characters: bool` and `illegal-characters: list = []`.
Variables that can take on many different data types usually make for ugly code that's hard to implement and hard to read.

danielhuppmann:
> I'd split that into two attributes check-illegal-characters: bool and illegal-characters: list[str] = [].
Variables that can take on many different data types and functions usually make for ugly code that's hard to implement and hard to read.

Fair enough. Sounds ok, @dc-almeida?

phackstock:
In addition, I would set the default `illegal-characters: list[str] = ["'", ":", ";"]` so that we don't need to explicitly copy over the illegal characters to every new config file.

nomenclature/config.py
phackstock: Opening this up again. @dc-almeida, it seems that you marked this conversation as resolved without implementing the suggestion or saying why you won't implement it.
Please either implement the suggestion or provide a reason for not implementing.

dc-almeida: Apologies, yesterday I did the commit but hadn't been pushed. Should be good now.

nomenclature/codelist.py
phackstock: Similar to my previous comment, you marked this as resolved but did not implement it or provide a reason why you did not implement it.

nomenclature/codelist.py
phackstock: Similar to my previous comment, you marked this as resolved but did not implement it or provide a reason why you did not implement it.

phackstock: Similar to my previous comment, you marked this as resolved but did not implement it or provide a reason why you did not implement it.

nomenclature/codelist.py
```diff
@@ -364,6 +335,37 @@ def read_excel(cls, name, source, sheet_name, col, attrs=None):
...
+                    _check_string(k)
+                for v in value.values():
+                    _check_string(v)
+            elif isinstance(value, list):
+                for item in value:
+                    _check_string(item)
+
+        for code in self.mapping.values():
+            if not code.repository:
+                for attr in code.model_fields:
```
phackstock, line 364: You can simplify this by using the `model_dump` method provided by pydantic (https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump):
```python
                for value in code.model_dump(exclude="file").values():
                    _check_string(value)
```
This way has two benefits, firstly you don't have to check if the current attribute is a file and secondly you don't have to explicitly fetch the value.

phackstock (changes requested):
Two smaller changes, using `ErrorCollector` and using model_dump to check the Code attributes, then good to merge from my side.

phackstock (approved):
Thanks for the quick changes @dc-almeida. Good to be merged from my side now.

danielhuppmann (approved):
This looks good to me, thanks @dc-almeida! Just please add empty lines at the end of the new yaml files for the tests.

Before merging, can you please go through all workflow-repos in production and add the line `illegal_characters: [ ]` If necessary.

nomenclature/config.py
```diff
@@ -161,9 +161,16 @@ class NomenclatureConfig(BaseModel):
     repositories: dict[str, Repository] = Field(default_factory=dict)
     definitions: DataStructureConfig = Field(default_factory=DataStructureConfig)
     mappings: RegionMappingConfig = Field(default_factory=RegionMappingConfig)
+    check_illegal_characters: bool = True
+    illegal_characters: list[str] = ["'", ":", ";"]
```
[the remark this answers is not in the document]
danielhuppmann: Well, useful when the PR hit sincerest repos…

PR #419, opened 2024-10-22
Add CLI command to validate scenario data file from definitions

### LINKED ISSUE #404: CLI to validate a scenario data file
To make it easier for non-expert users to check their data against a DataStructureDefinition, we need a simple CLI that takes an input file (IAMC scenario data) and validates it against a the definitions.

Basically, this should be a reduced version of `cli_run_workflow`, but only take an `input file` (required) and  a `definitions` folder (default to 'definitions').

dc-almeida:
First step towards #404

### DIFF
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -288,8 +288,7 @@ def cli_run_workflow(
     default="definitions",
 )
 def cli_validate_scenarios(input_file: Path, definitions: Path):
-    """Run a given input file through a DataStructureDefinition validation,
-    checking it against defined codelist(s)
+    """Validate a scenario file against the codelists of a project
 
     Parameters
     ----------
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -277,3 +277,29 @@ def cli_run_workflow(
     df = getattr(workflow, workflow_function)(IamDataFrame(input_file))
     if output_file is not None:
         df.to_excel(output_file)
+
+
+@cli.command("validate-scenarios")
+@click.argument("input_file", type=click.Path(exists=True, path_type=Path))
+@click.option(
+    "--definitions",
+    help="Optional name for definitions folder",
+    type=str,
+    default="definitions",
+)
+def cli_validate_scenarios(input_file: Path, definitions: Path):
+    """Run a given input file through a workflow function defined in a workflow.py
+
+    Parameters
+    ----------
+    input_file : Path
+        Input data file, must be IAMC format, .xlsx or .csv
+    definitions : str
+        Definitions folder inside workflow_directory, by default "definitions"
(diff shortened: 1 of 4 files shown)


phackstock (changes requested):
Nice addition @dc-almeida (sorry for jumping in here with a review).
Couple of comments in line below.

dc-almeida:
Thanks for the hawk-eye as always!

phackstock:
> Thanks for the hawk-eye as always!

You're welcome, always easier to spot these things in other people's work :D

phackstock (approved):
Thanks for the quick changes @dc-almeida.
Good to be merged from my side, in principle.
However, as mentioned in #419, I think the name of the command might be misleading and I think it should include an option to specify the dimensions that should be validated.

tests/data/cli/validate_scenarios/definitions/region/regions.yaml
```diff
@@ -0,0 +1,4 @@
+- common:
```
danielhuppmann, line 1: Can you use an existing definitions-folder in the test-data? We're adding quite a lot of very similar items here and it's hard to keep an overview...

nomenclature/cli.py
```diff
@@ -277,3 +277,30 @@ def cli_run_workflow(
...
+@click.argument("input_file", type=click.Path(exists=True, path_type=Path))
+@click.option(
+    "--definitions",
+    help="Optional name for definitions folder",
+    type=click.Path(exists=True, path_type=Path),
+    default="definitions",
+)
+def cli_validate_scenarios(input_file: Path, definitions: Path):
+    """Run a given input file through a DataStructureDefinition validation,
+    checking it against defined codelist(s)
```
danielhuppmann, line 292: Try to keep the docstring such that the first line is really only one line, per numpydoc standards.
```suggestion
    """Validate a scenario file against the codelists of a project
```

danielhuppmann (approved):
Thank you!

phackstock:
Before we go ahead with the merge if I understood our discussion in #419 correctly, this PR is still missing a dimensions option so that you can specify which dimension you want to validate, correct @danielhuppmann?

danielhuppmann:
I have a suggestion: we merge as is, and then @dc-almeida does a follow-up PR adding the dimensions for this CLI, and also extending the tests for the `validate-scenario` and `validate-project` to make sure that all three cases work as expected:
- explicit `dimensions` argument (this is currently tested in `test_cli_custom_dimensions_runs()`)
- if not given, use `dimensions` argument set from `nomenclature.yaml` (not sure if this is tested)
- if also not given, check all available folders

phackstock:
Sounds good to me.

PR #420, opened 2024-10-29
Add tier as named attribute for VariableCode

### LINKED ISSUE #409: Make "tier" a known attribute as integer
We often use `tier` as a variable-attribute to indicate importance of a variable - tier 1 is mandatory to be reported in a scenario-comparison project, tier 2 and 3 are optional.

This could be added as a standard attribute (as integer) to the VariableCode, but with a tag-list twist (a very special case of #402).

- If the Variable has a tier and the tag-list doesn't: easy
- If the Variable has a tier and the tag-list has a tier-attribute of the type "+1" - add the value to the tier.
- If the Variable doesn't have a tier: ignore

[I suggest the format "+1" to clearly indicate that this is added. If the tier in a tag-list is an integer, an error should be raised.]

This would help in cases like [here](https://github.com/IAMconsortium/common-definitions/blob/main/definitions/variable/energy/final-energy.yaml#L26) where `Final Energy|Non-Energy Use|Chemicals|Liquids` should have tier 2 but `Final Energy|Non-Energy Use|Chemicals|High Value Chemicals|Liquids|Biomass` should probably be tier 4. The of these variables are made of the same tag-lists (each combining different priorities), so having this flexibility to indicate priority within a tag list would be very helpful...

dc-almeida:
Closes #409 
Adds `tier` as a named attribute (integer) to `VariableCode`.
Adds special functionality for `tier` attribute in tags as described in #409.
After bilateral discussion, the valid notation for `tier` attributes for tags is "^1"/"^2", and integers for variables; errors raised otherwise.
For compatibility with repos that use the standard `"{tag}"` notation in variables, some tests were temporarily disabled for this PR. Repos like common-definitions require update to their `tier` attributes.


tests/data/codelist/tier_attribute/invalid/tag_fuel.yaml
```diff
@@ -0,0 +1,4 @@
+- Fuel:
+  - Coal:
+      description: coal
+      tier: +1
```
danielhuppmann, line 4:
```suggestion
      tier: +1

```

tests/data/codelist/tier_attribute/valid/tag_fuel.yaml
```diff
@@ -0,0 +1,4 @@
+- Fuel:
+  - Coal:
+      description: coal
+      tier: ^1
```
danielhuppmann, line 4: Suggest to test both cases here: no change and increment by 1
```suggestion
  - Coal|Lignite:
      tier: ^1

```

tests/data/codelist/tier_attribute/valid/variables.yaml
```diff
@@ -0,0 +1,11 @@
...
+    description: Total primary energy consumption
+    unit: EJ/yr
+- Primary Energy|{Fuel}|{Sector}:
+    description: Primary energy consumption of {Fuel}
+    unit: EJ/yr
+    tier: 1
+- Share|{Fuel}:
+    description: Share of {Fuel} in the total primary energy mix
+    unit: '%'
+    tier: 2
```
danielhuppmann, line 11: This variable should be called "Primary Energy|{Fuel} [Share]" according to the conventions, provided another suggestion above.
```suggestion
```

tests/data/codelist/tier_attribute/valid/variables.yaml
```diff
@@ -0,0 +1,11 @@
+- Primary Energy:
+    description: Total primary energy consumption
+    unit: EJ/yr
+- Primary Energy|{Fuel}|{Sector}:
+    description: Primary energy consumption of {Fuel}
+    unit: EJ/yr
+    tier: 1
```
danielhuppmann, line 7: Primary fuel consumption by definition does not have a sector.
```suggestion
- Primary Energy|{Fuel}:
    description: Primary energy consumption of {Fuel}
    unit: EJ/yr
    # this variable does not have a tier to test that no tier is added at initialization    
- Final Energy|{Fuel}:
    description: Final energy consumption of {Fuel}
    unit: EJ/yr
    tier: 1 
- Final Energy|{Fuel}|{Sector}:
    description: Final energy consumption of {Fuel} in {Sector}
    unit: EJ/yr
    tier: 1
```

tests/data/codelist/tier_attribute/valid/tag_sector.yaml
```diff
@@ -0,0 +1,6 @@
+- Sector:
+    - Agriculture:
+        description: agriculture sector including fishing
+        tier: ^1
+    - Industry:
+        description: industrial sector excluding non-energy use (e.g.feedstocks)
```
danielhuppmann, line 6: For better intuition, items on the same list should (generally) have the same tier - it makes more sense that different hierarchy-levels have different tiers.
```suggestion
    - Industry:
        description: industrial sector excluding non-energy use (e.g.feedstocks)
    - Industry|Chemicals:
        description: chemicals industry
        tier: ^1

```

nomenclature/code.py
```diff
@@ -125,12 +125,26 @@ def replace_tag(self, tag: str, target: "Code") -> "Code":
         def _replace_or_recurse(_attr, _value):
             # if the attribute is a string and contains "{tag}" replace
             if isinstance(_value, str) and "{" + tag + "}" in _value:
-                # if the the target has the corresponding attribute replace
+                # if the target has the corresponding attribute replace
```
danielhuppmann, line 128:
```suggestion
                # if the target has the attribute, replace the tag with the value
```

nomenclature/code.py
```diff
@@ -125,12 +125,26 @@ def replace_tag(self, tag: str, target: "Code") -> "Code":
...
                 if _attr in target.flattened_dict:
                     return _value.replace("{" + tag + "}", getattr(target, _attr))
                 # otherwise return the name
                 else:
                     return _value.replace("{" + tag + "}", getattr(target, "name"))
+            # if the attribute is an integer and "tier"
+            elif isinstance(_value, int) and _attr == "tier":
```
danielhuppmann, line 135:
```suggestion
            elif _attr == "tier" and isinstance(_value, int):
```

nomenclature/code.py
```diff
@@ -125,12 +125,26 @@ def replace_tag(self, tag: str, target: "Code") -> "Code":
...
+                if (tag_tier := getattr(target, _attr, None)) in {"^1", "^2"}:
+                    return _value + int(tag_tier[-1])
+                # if tag doesn't have tier attribute
+                elif not tag_tier:
+                    return _value
+                # else misformatted tier in tag
+                else:
+                    raise ValueError(
+                        f"Tag '{target.name}' includes misformatted 'tier' attribute. "
+                        "Allowed values for tier attributes in tags are '^1' or '^2'."
```
danielhuppmann, line 146:
```suggestion
                        f"Invalid 'tier' attribute in tag '{target.name}': {tag_tier}\n"
                        "Allowed values are '^1' or '^2'."
```

danielhuppmann (commented):
Looks good, a few details in the implementation and on the tests.

danielhuppmann (commented):
This looks very nice! One minor issue inline and one suggestion on how to fix the common-definitions conflict - so that I have time to fix common-definitions for one release cycle.

danielhuppmann:
Sweet!

danielhuppmann (approved):
This looks great, will be very helpful for common-definitions!

PR #421, opened 2024-10-31
Add dimensions option to CLI validate-scenarios

dc-almeida:
Contributes to #404
Adds a `--dimensions` option to the `validate-scenarios` command and sets this variable as described [here](https://github.com/IAMconsortium/nomenclature/pull/419#issuecomment-2431411781).
The additional test for implicit dimension setting is separate from original tests to preserve slightly cleaner code because of a technical detail on how the `dimensions` variable is set by the test CLI runner (`(None,)`/`('',)` instead of `()`).

Test harmonisation between `validate-project` and `validate-scenarios` will come in upcoming PR to ensure the tests: 1) check exit code, 2) check type of exception, 3) check error message.
Setting `dimensions` according to `nomenclature.yaml` is done in the CLI function(s) to override `DataStructureDefinition`'s standard order of explicit dimensions -> `nomenclature.yaml` -> default to `["region", "variable"]`. In this way, if `nomenclature.yaml` does not set `dimensions` (= `None`), it checks `definitions` subfolders (as previously implemented and now updated for `validate-project`)


tests/data/cli/structure_validation/nomenclature.yaml
```diff
@@ -0,0 +1,3 @@
+dimensions:
+  - region
+  - variable
```
danielhuppmann, line 3:
```suggestion
  - variable

```

nomenclature/testing.py
```diff
@@ -144,12 +145,16 @@ def assert_valid_structure(
...
+            dimensions = NomenclatureConfig.from_file(
+                path / "nomenclature.yaml"
+            ).dimensions
         if not dimensions:
-            raise FileNotFoundError(
-                f"`definitions` directory is empty: {path / definitions}"
-            )
+            dimensions = [x.stem for x in (path / definitions).iterdir() if x.is_dir()]
+        if not dimensions:
+            raise FileNotFoundError(f"`definitions` directory is empty: {definitions}")
```
danielhuppmann, line 156: All of this is basically already implemented directly within the **DataStructureDefinition**-initialiation, see https://github.com/IAMconsortium/nomenclature/blob/0ae378f418a260315f88c7e19a204e83a8416bcb/nomenclature/definition.py#L66

tests/test_cli.py
```diff
@@ -423,10 +423,15 @@ def test_cli_run_workflow(tmp_path, simple_df):
 
 
 @pytest.mark.parametrize(
-    "status, unit, exit_code", [("valid", "EJ/yr", 0), ("invalid", "EJ", 1)]
+    "status, unit, dimensions, exit_code",
+    [
+        ("valid_1", "EJ/yr", "region", 0),
```
danielhuppmann, line 428: Maybe test for a list here?
```suggestion
        ("valid_1", "EJ/yr", ["region", "variable"], 0),
```

tests/test_cli.py
```diff
@@ -441,11 +446,50 @@ def test_cli_valid_scenarios(status, unit, exit_code, tmp_path):
...
-            str(
-                MODULE_TEST_DATA_DIR
-                / "structure_validation_no_mappings"
-                / "definitions"
-            ),
+            str(MODULE_TEST_DATA_DIR / "structure_validation" / "definitions"),
+            "--dimension",
+            dimensions,
+        ],
+    )
```
danielhuppmann, line 453: You'll have to expand the list by adding something like
```python
dimension_args = []
for dim in dimensions:
    dimension_args.append("--dimension")
    dimension_args.append(dim)
```
Maybe this can be done more elegantly...?

```suggestion
       ] + dimension_args
    )
```

danielhuppmann (commented):
Looks good and works, but a few suggestions inline.

nomenclature/testing.py
dc-almeida: This is what I was pointing out initially about having to set dimensions outside the `DataStructureDefinition`. In order to have the order be:

>    * explicit `dimensions` argument (this is currently tested in `test_cli_custom_dimensions_runs()`)
>    * if not given, use `dimensions` argument set from `nomenclature.yaml` (not sure if this is tested)
>    * if also not given, check all available folders

I define it outside DSD, since within it, the order is: explicit `dimensions` -> `nomenclature.yaml` -> `["region", "variable"]`. The only way (besides changing the code in DSD) to enforce the last step is by doing the order before calling DSD and passing the dimensions. In its current implementation, the DSD won't check folders (hence why `validate-project` does that check in the CLI function too).

danielhuppmann: Ok, I think this is a misunderstanding - this function should just check whether a project can be instantiated correctly, and I think the extra code here is just from long ago when DataStructureDefinition wasn't as elegant as it is now.

danielhuppmann: I think we should go the other way and change the default-dimensions of a DataStructureDefinition to all sub-folders in `definitions`, not hard-coded to region and variable.

But this should be left to another PR, simply revert the changes here and we merge this PR?

dc-almeida: Agreed that it makes more sense to change the default dimensions in DSD! Will do the suggested changes.

danielhuppmann (approved):
Nice, thank you!

PR #423, opened 2024-11-14
Update installation.rst

gidden:
Update install docs to be consistent with pypi slug


danielhuppmann (commented):
Thanks but can you please implement the same change in the Readme?

gidden:
done

PR #430, opened 2024-11-21
Allow numpy 2x

### LINKED ISSUE #429: Check if nomenclature can use latest numpy
Currently we're restricting numpy to <2. 
I don't remember if there is a good reason for that or not.

phackstock:
Closes #429.

Tests are running locally. Seems we should be good allowing numpy v2


danielhuppmann (approved):
Thanks for looking into this!

PR #433, opened 2024-11-27
Format type hinting to 3.10+

dc-almeida:
Quick clean-up of type hinting to adhere to nomenclature's Python versions.
Replaced deprecated aliases, made type hinting consistent across files, applied recommended shorthand and removed unnecessary imports.

### DIFF
--- a/tests/test_codelist.py
+++ b/tests/test_codelist.py
@@ -358,7 +358,7 @@ def test_RegionCodeList_filter():
 
 
 def test_RegionCodeList_hierarchy():
-    """Verifies that the hierarchy method returns a List[str]"""
+    """Verifies that the hierarchy method returns a list[str]"""
 
     rcl = RegionCodeList.from_directory(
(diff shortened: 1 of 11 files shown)


tests/test_codelist.py
```diff
@@ -358,7 +358,7 @@ def test_RegionCodeList_filter():
 
 
 def test_RegionCodeList_hierarchy():
-    """Verifies that the hierarchy method returns a List[str]"""
+    """Verifies that the hierarchy method returns a list[str]"""
```
danielhuppmann, line 361:
```suggestion
    """Verifies that the hierarchy method returns a list"""

```

danielhuppmann (approved):
One minor suggestion, then good from my side

phackstock (approved):
Any PR that has more deletions than additions while preserving functionality, makes me happy. Thanks @dc-almeida!

PR #436, opened 2024-12-13
Allow easier import of NUTS regions

### LINKED ISSUE #435: Allow easier import of all NUTS regions
I tried to use NUTS regions in production for iDesignRES and I have three requests for simpler use:

1. Allow wildcard-selection: currently a user has to list all ISO2 codes individiually. It would be better to allow a wildcard import like
    ```yaml
    region:
        nuts:
          nuts-1: "*"
          nuts-2: "*"
          nuts-3: "*"
    ```
2. It would be easier for working with NUTS regions in the Scenario Explorer if the NUTS-level would be part of the hierarchy, so rather than importing all NUTS-regions as "NUTS 2021-2024", please change that to "NUTS 1 regions (2021-2024)".

dc-almeida:
Closes #435
Allows setting NUTS level config parameters as "true"

### DIFF
--- a/docs/user_guide/config.rst
+++ b/docs/user_guide/config.rst
@@ -126,12 +126,12 @@ By setting *definitions.region.nuts* (optional) in the configuration file:
       nuts:
         nuts-1: [ AT, BE, CZ ]
         nuts-2: [ AT ]
-        nuts-3: [ AT, BE ]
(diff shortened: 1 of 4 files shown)


danielhuppmann (approved):
Nice!

PR #440, opened 2024-12-18
Support directional region validation

danielhuppmann:
This PR implements validation for "directional regions" - when defining a region to report e.g. trade flows, this can be implemented as "China>Europe" using a > character. Both the source and the destination have to be defined in the codelist.

### DIFF
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -788,6 +788,18 @@ class RegionCodeList(CodeList):
                     )
                 )
             mapping[code.name] = code
+
+        for code in list(mapping):
+            if ">" in code:
+                origin, destination = code.split(">")
+                for region in [origin, destination]:
+                    if region not in mapping:
+                        errors.append(
+                            ValueError(
+                                f"Region '{region}' not defined for '{code}'"
+                            )
+                        )
+
         if errors:
             raise ValueError(errors)
         return cls(name=name, mapping=mapping)
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -801,11 +801,11 @@ class RegionCodeList(CodeList):
             if region.is_directional:
                 if region.origin not in v:
                     missing_regions.append(
-                        f"Region '{region.origin}' not defined for '{region.name}'"
+                        f"Origin '{region.origin}' not defined for '{region.name}'"
                     )
                 if region.destination not in v:
                     missing_regions.append(
-                        f"Region '{region.destination}' not defined for '{region.name}'"
+                        f"Destination '{region.destination}' not defined for '{region.name}'"
                     )
         if missing_regions:
             raise ValueError("\n".join(missing_regions))
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -795,9 +795,7 @@ class RegionCodeList(CodeList):
                 for region in [origin, destination]:
                     if region not in mapping:
                         errors.append(
-                            ValueError(
-                                f"Region '{region}' not defined for '{code}'"
-                            )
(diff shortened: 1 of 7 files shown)


dc-almeida (approved):
That clarifies why I had seen regions with the '>' before!

nomenclature/codelist.py
```diff
@@ -788,6 +788,16 @@ def from_directory(
                     )
                 )
             mapping[code.name] = code
+
+        for code in list(mapping):
```
phackstock, line 792: To me this does not belong into the constructor of the class but rather is a classic application of a pydantic validator.
I have built this on the branch so if you don't mind, I'd push my changes.

phackstock (changes requested):
One change requested, then good to merge.
If it's ok, I'd push my proposed change and you can take a look.

nomenclature/codelist.py
danielhuppmann: Yes, please push, thank you!

nomenclature/codelist.py
```diff
@@ -788,10 +788,29 @@ def from_directory(
...
+        for region in v.values():
+            if region.is_directional:
+                if region.origin not in v:
+                    missing_regions.append(
+                        f"Region '{region.origin}' not defined for '{region.name}'"
+                    )
+                if region.destination not in v:
+                    missing_regions.append(
+                        f"Region '{region.destination}' not defined for '{region.name}'"
+                    )
```
danielhuppmann, line 809: Maybe slightly more concise?
```suggestion
                for _region in [region.origin, region.destination]:
                    if _region not in v:
                        missing_regions.append(
                            f"Region '{_region}' not defined for '{region.name}'"
                        )
```

danielhuppmann:
Thanks @phackstock for the improvement, very nice!

I made one minor suggestion to reduce the duplicated validation-statement, feel free to decide whether that should be merged. Over to you to approve.

nomenclature/codelist.py
phackstock: I see the argument against duplication but I consciously opted against it for readability. 
To me it's easier to grasp what we're doing when we explicitly have `region.origin` and `region.destination` in the messages.

phackstock (approved):
I'd keep the two checks for origin and destination explicit for better readability.

nomenclature/codelist.py
```diff
@@ -788,10 +788,29 @@ def from_directory(
...
 
+    @field_validator("mapping")
+    @classmethod
+    def check_directional_regions(cls, v: dict[str, RegionCode]):
+        missing_regions = []
+        for region in v.values():
+            if region.is_directional:
+                if region.origin not in v:
+                    missing_regions.append(
+                        f"Region '{region.origin}' not defined for '{region.name}'"
```
danielhuppmann, line 804:
```suggestion
                        f"Origin '{region.origin}' not defined for '{region.name}'"
```

nomenclature/codelist.py
```diff
@@ -788,10 +788,29 @@ def from_directory(
...
+                    )
+                if region.destination not in v:
+                    missing_regions.append(
+                        f"Region '{region.destination}' not defined for '{region.name}'"
```
danielhuppmann, line 808:
```suggestion
                        f"Destination '{region.destination}' not defined for '{region.name}'"
```

danielhuppmann (commented):
Fair enough @phackstock, but then we can at least make the error messages more specific - what do you think?

phackstock:
> Fair enough @phackstock, but then we can at least make the error messages more specific - what do you think?

Good point, suggestions are merged.

PR #442, opened 2024-12-18
Add a test for unexpected failing of validation with multiple units

### LINKED ISSUE #441: One of multiple allowed units not passing validation
As just reported by @EmiFej, nomenclature scenario-validation currently fails if multiple units are allowed.

Error message:
```
 - 'Primary Energy|Coal' - expected: one of '['EJ/yr', 'GWh/yr']', found: 'GWh/yr'
```

FYI @phackstock @dc-almeida

danielhuppmann:
Closes #441

### DIFF
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -622,7 +622,7 @@ class VariableCodeList(CodeList):
             (variable, unit, self.mapping[variable].unit)
             for variable, unit in unit_mapping.items()
             if variable in self.variables
-            and not set(unit) == set(self.mapping[variable].unit)
+            and unit not in self.mapping[variable].unit
         ]:
             lst = [
                 f"'{v}' - expected: {'one of ' if isinstance(e, list) else ''}"
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -7,7 +7,7 @@ import numpy as np
 import pandas as pd
 import yaml
 from pyam import IamDataFrame
-from pyam.utils import is_list_like, write_sheet, pattern_match
+from pyam.utils import is_list_like, write_sheet, pattern_match, to_list
 from pydantic import BaseModel, ValidationInfo, field_validator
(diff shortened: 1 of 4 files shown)


danielhuppmann:
No idea why there was no test and why the submissions started failing now...

Due to some (weirdly) failing test, I noticed that the check "illegal characters in common definitions" was actually now testing for characters that were sanitized out of common-definition - so I added a quickfix for that (using a character that appears there) on top of the unit-quickfix.

nomenclature/codelist.py
```diff
@@ -622,7 +622,7 @@ def validate_units(
             (variable, unit, self.mapping[variable].unit)
             for variable, unit in unit_mapping.items()
             if variable in self.variables
-            and not set(unit) == set(self.mapping[variable].unit)
+            and unit not in to_list(self.mapping[variable].unit)
```
phackstock, line 625: The `to_list(self.mapping[variable].unit)` already exists as property `units` of `VariableCode`:
```suggestion
            and unit not in self.mapping[variable].units
```

nomenclature/codelist.py
```diff
@@ -7,7 +7,7 @@
 import pandas as pd
 import yaml
 from pyam import IamDataFrame
-from pyam.utils import is_list_like, write_sheet, pattern_match
+from pyam.utils import is_list_like, write_sheet, pattern_match, to_list
```
phackstock, line 10:
```suggestion
from pyam.utils import is_list_like, write_sheet, pattern_match
```

phackstock (changes requested):
Two small changes, then good to be merged.
I'd also suggest to rename the PR because you're not just adding a test, you're fixing the multiple unit issue itself.

danielhuppmann:
Thanks for the suggestion, implemented your changes.

phackstock (approved):
Thanks for implementing my suggestions, good to be merged.

PR #447, opened 2024-12-19
Support pyam v3.0 and fix incompatible test

danielhuppmann:
This PR bumps the pyam dependency to support the latest version, and it fixes a test that was failing due to the no-sorting behavior in pyam such that tests will pass both with pyam v2.3 and v3.0


phackstock (approved):
Perfect, thanks for the fix.

PR #448, opened 2024-12-19
Fix wildcard aggregation by skipping

### LINKED ISSUE #444: Region aggregation fails for wildcard variables
This dict lookup fails because we don't have an exact match:

https://github.com/IAMconsortium/nomenclature/blob/c3ad7daf699906d4e9ec8ed538434033194e7ffc/nomenclature/codelist.py#L598-L605

### LINKED ISSUE #445: test_illegal_char_ignores_external_repo does not clean up after itself
The test `test_illegal_char_ignores_external_repo` should have the following lines at the end:
```python
    finally:
        clean_up_external_repos(nomenclature_config.repositories)
```

danielhuppmann:
This PR is an alternative solution to #446, assuming that wildcard-variables should not be aggregated.

So this PR implements the following:
- Require that wildcard-variables have an explicit attribute "skip-region-aggregation: True" to avoid confusion and be forward-compatible in case we ever decide to change this behavior
- Skip common-region-aggregation for any variables that are not explicitly listed in the VariableCodeList (i.e., wildcard variables)
- Deactivate aggregate-check for any variables that are not explicitly listed in the VariableCodeList (i.e., wildcard variables)
 
In the process, I also noticed that the test `test_region_processing_skip_aggregation` passed even though the mapping-file was misspelled (m_a instead of model_a), but because the test looked at skipping aggregation, skip-aggregation and no mapping actually had the same effect. I fixed it anyway

closes #444 
closes #445

### DIFF
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -601,7 +601,9 @@ class VariableCodeList(CodeList):
         return [
             var
             for var in variables
-            if var in self.keys() and not self[var].agg_kwargs and not self[var].skip_region_aggregation
+            if var in self.keys()
+            and not self[var].agg_kwargs
+            and not self[var].skip_region_aggregation
         ]
 
     def vars_kwargs(self, variables: list[str]) -> list[VariableCode]:
@@ -610,7 +612,9 @@ class VariableCodeList(CodeList):
         return [
             self[var]
             for var in variables
-            if var in self.keys() and self[var].agg_kwargs and not self[var].skip_region_aggregation
+            if var in self.keys()
+            and self[var].agg_kwargs
+            and not self[var].skip_region_aggregation
         ]
 
     def validate_units(
--- a/nomenclature/codelist.py
+++ b/nomenclature/codelist.py
@@ -595,11 +595,11 @@ class VariableCodeList(CodeList):
             )
         return v
 
-    def vars_default_args(self, variables: list[str]) -> list[VariableCode]:
+    def vars_default_args(self, variables: list[str]) -> list[str]:
         """return subset of variables which does not feature any special pyam
         aggregation arguments and where skip_region_aggregation is False"""
         return [
-            self[var]
+            var
             for var in variables
             if not self[var].agg_kwargs and not self[var].skip_region_aggregation
         ]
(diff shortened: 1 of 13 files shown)


danielhuppmann:
At least this issue made me realize that I hadn't set skip-region-aggregation where I should have done so...
https://github.com/IAMconsortium/common-definitions/pull/249

tests/test_core.py
```diff
@@ -296,6 +296,51 @@ def test_region_processing_skip_aggregation(model_name, region_names):
     assert_iamframe_equal(obs, exp)
 
 
+@pytest.mark.parametrize(
+    "model_name, region_names",
+    [("model_a", ("region_A", "region_B")), ("model_b", ("region_A", "region_b"))],
+)
+def test_region_processing_wildcard_skip_aggregation(model_name, region_names):
```
phackstock, line 303: I think this test misses a crucial case where the code would currently still fail. The model mappings contain no aggregation instructions. I think they have to since most, if not all, real world examples contain aggregation instructions.
Because it skips all aggregation it also skips over a point where the code would still fail (I think).
I've added comments above where I think that would be.

nomenclature/codelist.py
```diff
@@ -610,7 +612,9 @@ def vars_kwargs(self, variables: list[str]) -> list[VariableCode]:
         return [
             self[var]
             for var in variables
-            if self[var].agg_kwargs and not self[var].skip_region_aggregation
+            if var in self.keys()
+            and self[var].agg_kwargs
```
phackstock, line 616: If we attempted aggregation with wildcard variables (which is the real world example because we need to look at them to decide if we skip or not) this code would throw and error because the lookup for: a variable called `Variable|1` would fail if there's a code `Variable*`

nomenclature/codelist.py
```diff
@@ -595,13 +595,15 @@ def check_weight_in_vars(cls, v):
...
+    def vars_default_args(self, variables: list[str]) -> list[str]:
         """return subset of variables which does not feature any special pyam
         aggregation arguments and where skip_region_aggregation is False"""
         return [
-            self[var]
+            var
             for var in variables
-            if not self[var].agg_kwargs and not self[var].skip_region_aggregation
+            if var in self.keys()
+            and not self[var].agg_kwargs
```
phackstock, line 605: If we attempted aggregation with wildcard variables (which is the real world example because we need to look at them to decide if we skip or not) this code would throw and error because the lookup for: a variable called Variable|1 would fail if there's a code Variable*

phackstock (changes requested):
As it stands right now, I think this PR does not fix the problem.
See details below.

PR #452, opened 2025-01-08
Remove redundant illegal character check attribute

dc-almeida:
The boolean attribute is unnecessary when the list of illegal characters can be empty


danielhuppmann (approved):
Makes sense to me, thanks!

phackstock (approved):
Sensible defaults make more readable code :+1: 
Very nice, thanks @dc-almeida :+1:

PR #459, opened 2025-01-15
Add `from_external_repository` attribute to `Code`

### LINKED ISSUE #458: Rename `repository` attribute in `Code` for readability
Not part of this PR, but I'd suggest renaming this attribute to `from_external_repository`. This would make the code easier to read.

_Originally posted by @phackstock in https://github.com/IAMconsortium/nomenclature/pull/456#discussion_r1916221928_

dc-almeida:
Closes #458

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -28,7 +28,7 @@ class Code(BaseModel):
     description: str | None = None
     file: str | Path | None = None
     extra_attributes: dict[str, Any] = {}
-    repository: str | None = None
+    from_external_repository: str | None = None
 
     def __eq__(self, other) -> bool:
         return self.model_dump(exclude="file") == other.model_dump(exclude="file")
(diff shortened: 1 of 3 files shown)


phackstock (changes requested):
Sorry, I did incorrectly think that the `repository` attribute is a boolean that's set to true if the code is from an external repository. 
Since it instead contains the actual name of the repository, I'd actually keep the attribute as is including the name. 
Instead I'd introduce a property that makes it easier to assess if a code is from an external repository. 

```python
    @property
    def from_external_repository(self) -> bool:
        return self.repository is not None
```

I'd use this property then in #456.

phackstock (approved):
Thanks @dc-almeida, good to be merged from my side. 
Then you can rebase https://github.com/IAMconsortium/nomenclature/pull/456, make use of the `from_external_repo` property and also merge that one.

PR #462, opened 2025-01-22
Add generic `Aggregator` processor

danielhuppmann:
This PR is in relation to a request by @jkikstra for a CEDS-emissions-processor that renames and aggregates emissions-variables from IAM results to match the CEDS schema - see the related PR at https://github.com/IAMconsortium/common-definitions/pull/255

The idea is to have a mapping like
```yaml
- dimension: variable
- aggregate:
  - Target Code:
      - Variable A
      - Variable B
```

that can then be called like 
```python
ceds_mapping = AggregationMapping.from_yaml("ceds-mapping/ceds-variable-mapping.yaml")
ceds_aggregator = Aggregator(mapping=ceds_mapping)

ceds_aggregator.apply(df)
```
to apply the aggregation-mapping onto an IamDataFrame `df`.

The method uses the pyam `rename()` method because that actually works (with summation) on all dimensions. This could be extended later for different aggregation-methods, etc.

This could be useful for @lisahligono as part of the scenario-transfer-routine, where the rename-mapping for regions is implemented as an external yaml file and then just called like shown above as part of the processing-and-transfer.


danielhuppmann:
Yes, I will still add tests but wanted to hear any comments about feature-design first...

phackstock:
As per bilateral discussion, I'd say this is a good way forward.
I'd include the dimension attribute in the yaml file.

dc-almeida:
Makes sense to me!

danielhuppmann:
Ready for review @phackstock, I have a few questions from your pedantic wisdom:
I defined an attribue `codes` for the `validate_with_definition()` method, but it seems I cannot use `codes` in the field-validators (even with after-mode)?

Apart from that, I'm not sure about a good way to include the file name in a validation-warning - but maybe we can also leave that for a future clean-up and standardization effort across the package, we are probably not 100% consistent. There are also a few overlaps with the region-processing feature, but I wanted to focus on this PR for now.

I will also add a "rename"-config alternative to work with what @lisahligono implemented in https://github.com/IAMconsortium/common-definitions/pull/272.

I will add documentation in a follow-up PR once #470 is merged.

nomenclature/processor/generic.py
```diff
@@ -0,0 +1,153 @@
...
+        _codes = list()
+        for item in self.mapping:
+            _codes.append(item.name)
+            _codes.extend(item.components)
+        return _codes
+
+    def validate_with_definition(self, dsd: DataStructureDefinition) -> None:
+        error = None
+        # check for codes that are not defined in the codelists
+        codelist = getattr(dsd, self.dimension, None)
```
phackstock, line 98: I'd let the error occur right here if the dimension is not found:
```suggestion
        codelist = getattr(dsd, self.dimension)
```

nomenclature/processor/generic.py
```diff
@@ -0,0 +1,153 @@
...
+    @property
+    def codes(self):
+        _codes = list()
+        for item in self.mapping:
+            _codes.append(item.name)
+            _codes.extend(item.components)
+        return _codes
+
+    def validate_with_definition(self, dsd: DataStructureDefinition) -> None:
+        error = None
```
phackstock, line 96: If you raise the error below directly, you don't need this
```suggestion
```

nomenclature/processor/generic.py
```diff
@@ -0,0 +1,153 @@
...
+        # check for codes that are not defined in the codelists
+        codelist = getattr(dsd, self.dimension, None)
+        # no validation if codelist is not defined or filter-item is None
+        if codelist is None:
+            error = f"Dimension '{self.dimension}' not found in DataStructureDefinition"
+        elif invalid := codelist.validate_items(self.codes):
+            error = (
```
phackstock, line 103:
```suggestion
            raise ValueError(f"The following {self.dimension}s are not defined in the "
                "DataStructureDefinition:\n - " + "\n - ".join(invalid) + "\nin " + str(self.file) + "")

```

nomenclature/processor/generic.py
```diff
@@ -0,0 +1,153 @@
...
+            dimension=mapping_input["dimension"],
+            mapping=mapping_list,  # type: ignore
+            file=get_relative_path(file),
+        )
+
+
+def _validate_items(items, info, _type):
+    duplicates = [item for item, count in Counter(items).items() if count > 1]
+    if duplicates:
+        raise PydanticCustomError(
```
phackstock, line 146: Even though I introduced them, I'd actually say we should move away from using custom errors. The way that I see it, they don't really provide any additional benefit and just make the code harder to read. 
We don't catch specific errors at any higher level and even if we did, they get wrapped in some pydantic error anyway

nomenclature/processor/generic.py
```diff
@@ -0,0 +1,153 @@
...
+from nomenclature.error import custom_pydantic_errors
+from nomenclature.processor import Processor
+from nomenclature.processor.utils import get_relative_path
+
+logger = logging.getLogger(__name__)
+
+here = Path(__file__).parent.absolute()
+
+
+class AggregationItem(BaseModel):
```
phackstock, line 21: I'm not sure this class is actually needed. 
I think you could just use: `mapping: dict[str, list[str]]`. This way you also don't need the `rename_mapping` property below since you can use `mapping` directly.

nomenclature/processor/generic.py
```diff
@@ -0,0 +1,153 @@
...
+        elif invalid := codelist.validate_items(self.codes):
+            error = (
+                f"The following {self.dimension}s are not defined in the "
+                "DataStructureDefinition:\n - " + "\n - ".join(invalid)
+            )
+        if error:
+            raise ValueError(error + "\nin " + str(self.file) + "")
+
+    @classmethod
+    def from_file(cls, file: Path | str):
```
phackstock, line 111: If you rename the `mapping` attribute to `aggregate` and use `dict[str, list[str]]` in favor of `AggregationItem`, I think you should be able to remove this function almost entirely.

phackstock (commented):
I would say that there's potential to simplify and shorten the code.
I have left some comments where exactly to do so in line.

phackstock:
> I defined an attribue codes for the validate_with_definition() method, but it seems I cannot use codes in the field-validators (even with after-mode)?

That might not be possible, I'll play around with the code a bit more to see.

nomenclature/processor/generic.py
phackstock, line 1: Before I forget it, I'd suggest to rename to `aggregator.py`

nomenclature/processor/generic.py
danielhuppmann: I like the current implementation better because this way, I can show the name of the offending file directly in the error message, making it easier to debug in repositories that have many files.

nomenclature/processor/generic.py
danielhuppmann: As above.

nomenclature/processor/generic.py
danielhuppmann: As above.

nomenclature/processor/generic.py
```diff
@@ -0,0 +1,153 @@
...
+        codelist = getattr(dsd, self.dimension, None)
+        # no validation if codelist is not defined or filter-item is None
+        if codelist is None:
+            error = f"Dimension '{self.dimension}' not found in DataStructureDefinition"
+        elif invalid := codelist.validate_items(self.codes):
+            error = (
+                f"The following {self.dimension}s are not defined in the "
+                "DataStructureDefinition:\n - " + "\n - ".join(invalid)
+            )
+        if error:
```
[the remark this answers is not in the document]
danielhuppmann: As above.

nomenclature/processor/generic.py
```diff
@@ -0,0 +1,153 @@
...
+            raise ValueError(error + "\nin " + str(self.file) + "")
```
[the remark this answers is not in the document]
danielhuppmann: As above.

nomenclature/processor/generic.py
danielhuppmann: Well, one benefit of the custom errors is that the code in the actual files is shorter, and the concatenation of a readable string with filename, etc. is moved elsewhere.

I agree that the error messages should be standardized and reused, but this PR is not the right place for that discussion.

nomenclature/processor/generic.py
```diff
@@ -0,0 +1,153 @@
...
+
+    name: str
+    components: list[str]
+
+
+class Aggregator(Processor):
+    """Aggregation or renaming of an IamDataFrame on a `dimension`"""
+    file: FilePath
+    dimension: str
+    mapping: list[AggregationItem]
```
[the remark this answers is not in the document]
danielhuppmann: Good suggestion, thanks.

nomenclature/processor/generic.py
danielhuppmann: I'd prefer follow the current implementation in the RegionProcessor, and then see how we can streamline or refactor to use common utils methods/classes...

nomenclature/processor/generic.py
danielhuppmann: Thanks, done!

danielhuppmann:
Thanks for your comments @phackstock!

I renamed file names and modules to **aggregator** and implemented some suggestions, but I'd push larger refactoring like using standardized error messages to a follow-up PR.

nomenclature/processor/generic.py
phackstock: Yes, exactly

nomenclature/processor/generic.py
phackstock: Fair point, but in this case the error does not really have anything to do with the aggregator processor but rather with the missing dimension in the DataStructueDefinition.

phackstock (approved):
Thanks for incorporating the changes @danielhuppmann. 
Agreed that we should push for a larger refactor and agreed that this PR is not the time and place for that.
So good to be merged from my side.

PR #463, opened 2025-01-28
Check if constituents of common regions are in native regions

### LINKED ISSUE #443: Check native regions belong to the same model in mappings
Following up on [this common-definitions PR](https://github.com/IAMconsortium/common-definitions/pull/247), check if the constituent regions used are the same native regions listed in the mappings file

dc-almeida:
Closes #443
Constituent regions used for region aggregation should also be native regions, otherwise, unexpected errors may appear (as caught in [this PR](https://github.com/IAMconsortium/common-definitions/pull/247).

### DIFF
--- a/nomenclature/error.py
+++ b/nomenclature/error.py
@@ -40,6 +40,10 @@ pydantic_custom_error_config = {
         "region_not_defined",
         "Region(s)\n{regions}\nin {file}\nnot found in RegionCodeList",
     ),
+    "ConstituentsNotNativeError": (
+        "constituents_not_native",
+        "Constituent region(s)\n{regions}\nin {file} not found in RegionCodeList",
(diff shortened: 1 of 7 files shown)


dc-almeida:
The two tests failing are caused by the common-definitions hash pointing to a version where the error found [here](https://github.com/IAMconsortium/common-definitions/pull/247) remains:
https://github.com/dc-almeida/nomenclature/blob/04cf36a2fd31407b3be6329c51150a981daac6ef/tests/data/region_processing/external_repo_test/nomenclature.yaml#L4-L7

phackstock:
@dc-almeida, yes please go ahead and update the hash.

danielhuppmann (approved):

PR #466, opened 2025-01-29
Remove None defaults from RegionAggregationMapping

### LINKED ISSUE #465: Get rid of None as default where possible
Currently, we have `None` as the default value in a lot of places. This makes for code that is hard to read, write and maintain.
One example is class variables that are lists that can be empty. The default should be an empty list rather than None. This way, worst case, you iterate over an empty list rather than needing a guard clause before the iteration.

phackstock:
Closes #465.

This PR removes the `None` defaults for `native_regions`, `common_regions` and `exclude_regions` and replaces it with empty lists.
This allows us to get rid of some if guards that we had before attempting to iterate over the above lists. As a result, `_apply_region_processing` is now less nested and easier to read.
In addition I also cleaned up `RegionAggregationMapping` a bit making use of some more pydantic native features. The two clean ups are:
* Use `model_config = ConfigDict(extra="forbid")` to prevent additional attributes instead of checking explicitly 
* Refactor `to_yaml` to use serializers for more complex fields.

### DIFF
--- a/nomenclature/processor/region.py
+++ b/nomenclature/processor/region.py
@@ -396,28 +396,35 @@ class RegionAggregationMapping(BaseModel):
     def __eq__(self, other: "RegionAggregationMapping") -> bool:
         return self.model_dump(exclude={"file"}) == other.model_dump(exclude={"file"})
 
+    @field_serializer("model", when_used="json")
+    def serialize_model(self, model) -> str | list[str]:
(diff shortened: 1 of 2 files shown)


dc-almeida (approved):
Thanks! Once merged, I will rebase and refactor my code in #463 to reflect the change.

Issue #471, opened 2025-02-04
Add a "range" validator


danielhuppmann:
In addition to the datavalidator-options of explicit upper & lower bounds and the value-plus-tolerance config, please add an option to specify a "range" keyword argument that sets both an upper and lower bound in one line, like used [here](https://github.com/IAMconsortium/common-definitions/blob/6bf8d67d2ddf8d7033a27ca6eb90f5917f58cdaf/definitions/variable/techno-economic/electricity-generation.yaml#L25)

```yaml
- Efficiency|Electricity|{Electricity Source}|*:
    description: Conversion efficiency per unit of primary energy of a newly installed
      plant to generate electricity from {Electricity Source} for a specific technology
    unit: "%"
    range: [ 0,100 ]
```

Issue #469, opened 2025-01-30
Refactor DataValidator


phackstock:
The current implementation of DataValidator and its associated classes is rather convoluted. Before we add more functionality, we should streamline and simplify.

PR #473, opened 2025-02-07
Add `range` validation

dc-almeida:
#471
Adds a `range` attribute to set lower/upper bounds in validation criteria.
Adds a simple test showing it works equivalently to bounds.
Missing test(s) to raise validation error.

### DIFF
--- a/nomenclature/processor/data_validator.py
+++ b/nomenclature/processor/data_validator.py
@@ -7,7 +7,7 @@ import yaml
 from pandas import concat
 from pyam import IamDataFrame
 from pyam.logging import adjust_log_level
-from pydantic import computed_field, field_validator, model_validator
+from pydantic import computed_field, field_validator, model_validator, Field
 
 from nomenclature.definition import DataStructureDefinition
 from nomenclature.error import ErrorCollector
@@ -86,9 +86,49 @@ class DataValidationCriteriaBounds(DataValidationCriteria):
         )
 
 
+class DataValidationCriteriaRange(DataValidationCriteria):
+    range: list[float] = Field(..., min_length=2, max_length=2)
+
+    @model_validator(mode="after")
+    def check_range_is_valid(self):
+        if self.range[0] > self.range[1]:
+            raise ValueError("Validation range is invalid: " + str(self.criteria))
+        return self
+
+    @computed_field
+    def upper_bound(self) -> float:
+        return self.range[1]
+
+    @computed_field
+    def lower_bound(self) -> float:
+        return self.range[0]
+
+    @property
+    def validation_args(self):
+        """Attributes used for validation (as bounds)."""
+        return self.model_dump(
+            exclude_none=True,
+            exclude_unset=True,
+            exclude=["warning_level", "range"],
+        )
+
+    @property
+    def criteria(self):
+        return self.model_dump(
+            exclude_none=True,
+            exclude_unset=True,
+            exclude=["warning_level", "lower_bound", "upper_bound"],
+        )
+
+
 class DataValidationCriteriaMultiple(IamcDataFilter):
     validation: (
-        list[DataValidationCriteriaValue | DataValidationCriteriaBounds] | None
+        list[
+            DataValidationCriteriaValue
+            | DataValidationCriteriaBounds
+            | DataValidationCriteriaRange
+        ]
+        | None
     ) = None
 
     @model_validator(mode="after")
--- a/tests/data/validation/validate_data/validate_warning_range.yaml
+++ b/tests/data/validation/validate_data/validate_warning_range.yaml
@@ -1,0 +1,11 @@
+ - variable: Primary Energy
+   year: 2010
+   validation:
+    - range: [ 1, 5 ]
+    - warning_level: low
+      upper_bound: 2.5
+      lower_bound: 1
+ - variable: Primary Energy|Coal
+   year: 2010
+   upper_bound: 5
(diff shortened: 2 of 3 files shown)


nomenclature/processor/data_validator.py
```diff
@@ -86,9 +86,49 @@ def criteria(self):
...
+            exclude_unset=True,
+            exclude=["warning_level", "lower_bound", "upper_bound"],
+        )
+
+
 class DataValidationCriteriaMultiple(IamcDataFilter):
     validation: (
-        list[DataValidationCriteriaValue | DataValidationCriteriaBounds] | None
+        list[
+            DataValidationCriteriaValue
```
phackstock, line 127: I would change this to:
```suggestion
            DataValidationCriteria
```

nomenclature/processor/data_validator.py
```diff
@@ -86,9 +86,49 @@ def criteria(self):
...
+            | DataValidationCriteriaBounds
```
phackstock, line 128:
```suggestion
```

nomenclature/processor/data_validator.py
```diff
@@ -86,9 +86,49 @@ def criteria(self):
...
+            | DataValidationCriteriaRange
```
phackstock, line 129:
```suggestion
```

nomenclature/processor/data_validator.py
```diff
@@ -86,9 +86,49 @@ def criteria(self):
         )
 
 
+class DataValidationCriteriaRange(DataValidationCriteria):
+    range: list[float] = Field(..., min_length=2, max_length=2)
+
+    @model_validator(mode="after")
+    def check_range_is_valid(self):
```
phackstock, line 93: This validator represents a design choice in my opinion. We for sure want the range to be ordered and the way I see it, is we have two options to ensure this:
1. (implemented here) we check and throw and error if the range is not ordered
2. We take the input value for `range`, sort it and only then save it into the attribute

The same design question presents itself when talking about the ordering of the severity levels. Currently we through an error if they are not sorted by the most severe first. Alternatively, we could also take the input, sort it and then save it.

The question does not (and probably should not) be answered, I just wanted to raise the point to be discussed at some stage. My (slight) personal preference would be to take whatever the user provides us, sort it and then work with it. This way we could potentially reduce the amount of frustrated nomenclature users.

nomenclature/processor/data_validator.py
```diff
@@ -86,9 +86,49 @@ def criteria(self):
...
+        if self.range[0] > self.range[1]:
+            raise ValueError("Validation range is invalid: " + str(self.criteria))
```
phackstock, line 95: This error message does not contain enough information for the user to fix the issue:
```suggestion
            raise ValueError("Validation range must be ordered with the smaller value first: " + str(self.criteria))
```

nomenclature/processor/data_validator.py
```diff
@@ -86,9 +86,49 @@ def criteria(self):
...
+        return self
+
+    @computed_field
```
phackstock, line 98: Nice use of the `computed_field` decorator :+1:  To make things more explicit and follow the pydantic docs (https://docs.pydantic.dev/2.10/concepts/fields/#the-computed_field-decorator), I'd adding the property decorator as well for this and `lower_bound`.

phackstock (commented):
Thanks for adding this quickly @dc-almeida :+1: 
A few comments and suggested changes in line. 
One more general comment is that I'm not sure we need `DataValidationCriteriaRange` as a new class. In terms of functionality it's entirely identical to using upper and lower bound, just in a different notation in the yaml file.
Thinking about it, I'm not sure we should even offer two ways to do the same thing. I'd suggest to only support either `range` or `upper_bound` and `lower_bound`. I'd suggest only supporting the more compact `range`.
On a related note, the classes `IamcDataFilter` and `DataValidationCriteria` can and should be combined in my opinion. `IamcDataFilter` is never used standalone and only `DataValidationCriteria` inherits from it. That's probably better addressed though when tackling #469.

danielhuppmann:
> One more general comment is that I'm not sure we need DataValidationCriteriaRange as a new class. In terms of functionality it's entirely identical to using upper and lower bound, just in a different notation in the yaml file.
Thinking about it, I'm not sure we should even offer two ways to do the same thing.

Disagree, there are actually two distinct use cases: have both upper and lower bounds (i.e., a range), vs. having only one of those.

phackstock:
> Disagree, there are actually two distinct use cases: have both upper and lower bounds (i.e., a range), vs. having only one of those.

Fair point, still, having upper and lower bound entirely covers the use of range. If you want to provide users the opportunity to use both option then I think having two classes is the cleanest way to implement that. 
I'd still prefer to give users one, and only one way (upper and lower bound in this case), but I'm fine merging the PR. We are after all in the dangerous realm of personal preference :D

nomenclature/processor/data_validator.py
phackstock: > The design-choice on checking the order of severity levels was my suggestion - I think it's better to force users to follow an easy-to-read convention. And it also avoids the ordering as one extra step in the code, improving readability.

Fair point in making that design choice and sticking to it here as well, works for me. You don't really save any code though since in one case you'd be sorting the inputs, in the other case you're checking that they arrive sorted, six and one half dozen.

danielhuppmann:
Closing in favor of #477

PR #476, opened 2025-02-19
Use logging-config instead of basicConfig

danielhuppmann:
Finally took time to get to the bottom of why my scenario-processing scripts show all httpx-requests-logging... 

This PR harmonizes the logging-setup with ixmp4 and pyam.

FYI @lisahligono


phackstock (approved):
Ah very nice addition, thanks a lot. That'll align the logging output nicely for scse-processing. We'll still need to set the log level accordingly there but that should be a good start.

PR #482, opened 2025-03-04
Add mypy support when using nomenclature as part of another project

### LINKED ISSUE #481: Nomenclature does not support mypy
When importing nomenclature into an external project, mypy cannot read the type hints because we are missing `py.typed`.

phackstock:
Closes #481


danielhuppmann (approved):
Nice!

PR #484, opened 2025-03-07
Add option to output `DataValidator` results in xlsx

### LINKED ISSUE #438: Add `DataValidator` option for warning log as single DataFrame
Current implementation of warning log for `DataValidator` prints a DataFrame with flagged rows for each criterion.
Suggested by @danielhuppmann to join all warnings into a single DataFrame, ordered by warning level (descending).
Question: what is the best format for the criteria messages (which would not be joined too, I presume)?

### LINKED ISSUE #478: Validation warning order incorrect
When reviewing #477, I noticed that the current implementation of `WarningEnum` is incorrect.
We have an order of severity being "error", "high", "medium", "low" that we want to enforce. With the `WarningEnum` being implemented this way:

```python
class WarningEnum(str, Enum):
    error = "error"
    high = "high"
    medium = "medium"
    low = "low"
```

we do this by sorting the enum. As the enum is a string the ordering is alphabetical and is accidentally almost correct: "error", "high", "low", "medium".
I'd suggest to replace this with an int enum, the same way it's being done for logging in the python standard library: https://docs.python.org/3/library/logging.html#logging-levels.

dc-almeida:
Closes #438. Adds option to specify a file path for `DataValidator` to output an xlsx file that compiles the flagged failed validation data rows.
Also refactors some of the code, as suggested by @phackstock [here](https://github.com/IAMconsortium/nomenclature/pull/461/files#r1935447861), and incorporates the enum fixes by @phackstock from [here](https://github.com/danielhuppmann/nomenclature/pull/1) (closes #478).

### DIFF
--- a/nomenclature/processor/data_validator.py
+++ b/nomenclature/processor/data_validator.py
@@ -228,9 +228,6 @@ class DataValidator(Processor):
     def from_file(
         cls, file: Path | str, output_path: Path | str = None
     ) -> "DataValidator":
-        if output_path is not None:
-            if not isinstance(output_path, Path):
-                output_path = Path(output_path)
         with open(file, "r", encoding="utf-8") as f:
             content = yaml.safe_load(f)
         criteria_items = []


nomenclature/processor/data_validator.py
```diff
@@ -175,6 +189,33 @@ def filter_args(self):
     def __str__(self):
         return ", ".join([f"{key}: {value}" for key, value in self.filter_args.items()])
 
+    def apply(
```
phackstock, line 192: I'd strongly suggest against using `fail_list` and `output_list` as input values that we append to and not explicitly return. This is programming by side effect and is not recommended since it makes it harder to understand what is going on.
Instead I'd recommend to change the function to either return a tuple of a dictionary that contains `[error, fail_list, output_list]` and work with it further down the line.

nomenclature/processor/data_validator.py
```diff
@@ -204,7 +245,7 @@ def from_file(cls, file: Path | str) -> "DataValidator":
             criteria_items.append(item)
         return cls(file=file, criteria_items=criteria_items)  # type: ignore
 
-    def apply(self, df: IamDataFrame) -> IamDataFrame:
+    def apply(self, df: IamDataFrame, output_path: Path | str = None) -> IamDataFrame:
```
phackstock, line 248: The signature of this function won't work in production. The reason for this is that in order to run a number of `Processor` objects, one after the other in a streamlined way, they have to implement a .apply() method that **only** take a pyam dataframe and an input and return a pyam data frame. This way you can nicely stack processors like so:

```python
processors = [
    UsefulEnergyProcessor(),
    RegionProcessor.from_directory(path=here / "mappings", dsd=dsd),
    RequiredDataValidator.from_file(
        here / "required_data" / "IAM_requirements.yaml"
    ),
    RequiredDataValidator.from_file(
        here / "required_data" / "Downscaling_requirements.yaml"
    ),
    RequiredDataValidator.from_file(
        here / "required_data" / "Damages_requirements.yaml"
    ),
]
data = pyam.IamDataFrame()
for processor in processors:
    data = processor.apply(data)
```

In order to be able to keep the above pattern, I'd suggest to move `output_path` as an attribute of `DataValidator` and allow it to be set in the file or directly in the `from_file` function, like so:

```python
DataValidator.from_file("data_validation_file.yaml", output_path= Path("results.xlsx"))
```

then `DataValidator` can use the apply function in the same way as all other `Processor`s with just a data frame as an input.

nomenclature/processor/data_validator.py
```diff
@@ -175,15 +189,48 @@ def filter_args(self):
...
     criteria_items: list[DataValidationItem]
     file: Path
+    output_path: Path | None = None
 
     @classmethod
-    def from_file(cls, file: Path | str) -> "DataValidator":
+    def from_file(
+        cls, file: Path | str, output_path: Path | str = None
+    ) -> "DataValidator":
+        if output_path is not None:
```
phackstock, line 231: You don't need this check. Pydantic will automatically convert a string to a path. It automatically attempts any type conversion if the input type is not equal to the expected type.

phackstock (changes requested):
One small change, after that, good to be merged from my side without an additional review.
Thanks @dc-almeida.

nomenclature/processor/data_validator.py
dc-almeida: Okay, wasn't sure the str to Path coercion was possible, good to know!

phackstock (approved):

PR #487, opened 2025-03-26
Add datetime validation

dc-almeida:
Adds datetime validation, based on an initial implementation in [openENTRANCE](https://github.com/openENTRANCE/openentrance/blob/main/definitions/subannual/README.md).

### DIFF
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -245,6 +245,11 @@ class RegionMappingConfig(BaseModel):
         return v
 
 
+class TimeConfig(BaseModel):
+    year: bool = True
+    datetime: str = Field(pattern=r"^UTC([+-])(1[0-4]|0?[0-9]):([0-5][0-9])$")
+
+
 class DimensionEnum(str, Enum):
     model = "model"
     scenario = "scenario"
@@ -259,6 +264,7 @@ class NomenclatureConfig(BaseModel):
     definitions: DataStructureConfig = Field(default_factory=DataStructureConfig)
     mappings: RegionMappingConfig = Field(default_factory=RegionMappingConfig)
     illegal_characters: list[str] = [":", ";", '"']
+    time: None | TimeConfig = None
 
     model_config = ConfigDict(use_enum_values=True)
 
--- a/nomenclature/definition.py
+++ b/nomenclature/definition.py
@@ -16,6 +16,7 @@ from nomenclature.codelist import (
     MetaCodeList,
 )
 from nomenclature.config import NomenclatureConfig
+from nomenclature.error import ErrorCollector
 
 logger = logging.getLogger(__name__)
 SPECIAL_CODELIST = {
@@ -111,6 +112,50 @@ class DataStructureDefinition:
             for dimension in (dimensions or self.dimensions)
         ):
             raise ValueError("The validation failed. Please check the log for details.")
+
+    def validate_datetime(self, df: IamDataFrame) -> None:
+        """Validate datetime coordinates against allowed format and/or timezone.
+
+        Parameters
+        ----------
+        df : :class:`pyam.IamDataFrame`
+            Scenario data to be validated against the datetime format.
+
+        Returns
+        -------
+        None
+
+        Raises
+        ------
+        ValueError
+            If `df` fails validation against the required datetime format.
+        """
+        errors = ErrorCollector()
+        if self.config.time:
+            year = "2020-" if self.config.time.year else ""
+            dt = "subannual" if self.config.time.year else "time"
+
+            if dt in df.dimensions:
+                # collect datetime items not covered by subannual codelist
+                _datetime = [s for s in df[dt] if s not in self.subannual]
+                # if datetime is coerced in IamDataFrame, include seconds
+                dt_format = "%Y-%m-%d %H:%M" if year else "%Y-%m-%d %H:%M:%S"
+                for d in _datetime:
+                    try:
+                        _dt = datetime.strptime(f"{year}{d}", dt_format + "%z")
+                        # casting to datetime with timezone was successful
+                        if not _dt.tzname() == self.config.time.datetime:
+                            errors.append(ValueError(f"Invalid timezone: {d}"))
+                    except ValueError:
+                        try:
+                            _dt = datetime.strptime(f"{year}{d}", dt_format)
+                        except ValueError:
+                            errors.append(
+                                ValueError(f"Invalid subannual timeslice: {d}")
+                            )
+                        errors.append(ValueError(f"Missing timezone: {d}"))
+        if errors:
+            raise ValueError(errors)
 
     def check_aggregate(self, df: IamDataFrame, **kwargs) -> None:
         """Check for consistency of scenario data along the variable hierarchy
--- a/tests/data/data_structure_definition/subannual/subannual/months.yaml
+++ b/tests/data/data_structure_definition/subannual/subannual/months.yaml
@@ -1,0 +1,24 @@
+- January:
+    duration: 31 / 365
+- February:
+    duration: 28 / 365
+- March:
+    duration: 31 / 365
+- April:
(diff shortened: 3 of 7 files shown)


dc-almeida:
Some questions on implementation:
- currently, it makes use of `ErrorCollector` to point out multiple problems in the data. Considering the potential size of a DataFrame, this might lead to a too large error message, and considering the datetime format is expected to probably be consistent in the same file, maybe it's best to just fail fast on the first misformatted case?
- should it be added to `dsd.validate(df)`, or be a standalone function? If merged, I'm not sure how best to format the error/log messages, as `.validate()` outputs to log a list of forbidden codes, whereas datetime validation is not against a codelist. It can just raise its standard datetime format error separately before/after the rest of the validation.

I will update the docs after.

phackstock (changes requested):
Hopefully this was not still a work in progress.
I'd say this PR is still a little rough around the edges. It took me quite a long time to wrap my head around what exactly the code does and what options there are. Maybe it's just a complicated topic though.

phackstock:
Sorry @dc-almeida only saw your questions after my review:

> * currently, it makes use of ErrorCollector to point out multiple problems in the data. Considering the potential size of a DataFrame, this might lead to a too large error message, and considering the datetime format is expected to probably be consistent in the same file, maybe it's best to just fail fast on the first misformatted case?
> * should it be added to dsd.validate(df), or be a standalone function? If merged, I'm not sure how best to format the error/log messages, as .validate() outputs to log a list of forbidden codes, whereas datetime validation is not against a codelist. It can just raise its standard datetime format error separately before/after the rest of the validation.

Regarding the first point, I've seen people get very creative in using multiple different errors throughout a file, I'd dump everything since you have to go through everything anyways.
For the second point, I'd very much integrate it into `dsd.validate()`. For the variables we already test for units and so we've already broken the pattern of only outputting forbidden codes. Making sure that the output for the subannual is nicely formatted is a of course a good idea.

dc-almeida:
Incorporated some of the suggested changes, but considering this comment:

> I'm not sure I love this special treatment of the subannual column. Maybe we could change this to putting the pattern of the datetime into the codelist, something like:
> ```yaml
> - "%Y-%m-%d %H:%M%z":
>   pattern: true
> ```
> this would also make it explicit for the modelers on what to submit in the datetime column. We would create a Subannual codelist and put the validation logic in there. Then dsd.validate() would still work just as normal.

Thoughts, @danielhuppmann?

nomenclature/config.py
```diff
@@ -245,6 +245,11 @@ def convert_to_set_of_repos(cls, v):
         return v
 
 
+class TimeConfig(BaseModel):
+    year: bool = True
```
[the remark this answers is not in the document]
danielhuppmann: Agree, if this is the config to distinguish whether the data should be yearly or datetime, „TimeDomain“ (or maybe TimeDomainConfig) would be clearer.

danielhuppmann:
> For the second point, I'd very much integrate it into `dsd.validate()`. For the variables we already test for units and so we've already broken the pattern of only outputting forbidden codes. Making sure that the output for the subannual is nicely formatted is a of course a good idea.

Agree

tests/data/data_structure_definition/subannual/subannual/months.yaml
```diff
@@ -0,0 +1,24 @@
...
+- August:
+    duration: 31 / 365
+- September:
+    duration: 30 / 365
+- October:
+    duration: 31 / 365
+- November:
+    duration: 30 / 365
+- December:
+   duration: 31 / 365
```
danielhuppmann, line 24:
```suggestion
   duration: 31 / 365

```

tests/test_validation.py
```diff
@@ -108,3 +111,55 @@ def test_wildcard_match(simple_df):
...
+    definition.config.time = nomenclature.config.TimeConfig(
+        year=True,
+        datetime="UTC+01:00",
+    )
+    df = IamDataFrame(simple_df._data, subannual=subannual)
+    if not expected_error:
+        assert definition.validate_datetime(df) is None
+    else:
+        with pytest.raises(ValueError):
+            definition.validate_datetime(df)
```
danielhuppmann, line 138: Please remove this test case or change to only use the examples like months shown above. Using truncated datetime-strings in the subannual column was an hacky quickfix and shouldn't be supported in the future.

nomenclature/definition.py
```diff
@@ -112,6 +113,48 @@ def validate(self, df: IamDataFrame, dimensions: list | None = None) -> None:
...
+
+        Raises
+        ------
+        ValueError
+            If `df` fails validation against the required datetime format.
+        """
+        if self.config.time is None:
+            return
+        errors = ErrorCollector()
+        time_column_name = "subannual" if self.config.time.year else "time"
```
danielhuppmann, line 136: "subannual" should not be covered in this example but use the standard codelist-validation.

danielhuppmann (changes requested):
My apologies, I think that I sent you down a rabbit hole with the legacy examples. The "subannual" column should only be used for items like months or seasons, so this should not need any special treatment.

For the validation of the datetime format, this should be quite easy with the **IamDataFrame** attributes `time_domain` and `time` (which gives the list of date time-values used in the data as a **pandas.Index**.

nomenclature/definition.py
```diff
@@ -112,6 +113,50 @@ def validate(self, df: IamDataFrame, dimensions: list | None = None) -> None:
...
+            year = "2020-" if self.config.time.year else ""
+            dt = "subannual" if self.config.time.year else "time"
+
+            if dt in df.dimensions:
+                # collect datetime items not covered by subannual codelist
+                _datetime = [s for s in df[dt] if s not in self.subannual]
+                # if datetime is coerced in IamDataFrame, include seconds
+                dt_format = "%Y-%m-%d %H:%M" if year else "%Y-%m-%d %H:%M:%S"
+                for d in _datetime:
+                    try:
```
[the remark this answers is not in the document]
phackstock: Fair point @dc-almeida, I still find it a bit hard to read but you're correct. This seems to be the easiest way to achieve the functionality that we want.

dc-almeida:
Thanks for the clarification. Playing around with pyam, `time_domain` is set according to time column name, but there are some caveats: a column `year` doesn't enforce year format, so I can create an IamDataFrame with both years and datetimes in the same column, name it `year`, and its `time_domain` will be `year`, instead of `mixed`. In the same process, a column named `time` with only years will be `mixed`. Only the `datetime` time domain explicitly checks for a `DatetimeIndex`.
It's still easy to implement the datetime validation expecting years and/or datetimes, but I'm not sure if the above is desired behaviour in pyam.

dc-almeida:
Took some time because I initially looked into pyam to see if I could change the time domain column name setting, but realised that is more complex than initially seemed.
This PR currently assumes well-named time columns.
I would still argue like @phackstock that the datetime format could also be enforced with a codelist; it feels like this implementation may be making a new form of validation that could fit the standard one.

tests/data/data_structure_definition/subannual/subannual/months.yaml
```diff
@@ -0,0 +1,24 @@
+- January:
```
danielhuppmann, line 1: I think that this test doesn't have any effect (yet), so I suggest to remove and only add when categorical time-domain validation is added explicitly.

danielhuppmann (commented):
I tried to test this but not sure if I configured it incorrectly or misunderstood the implementation. Let's sit together tomorrow and discuss how to proceed.

dc-almeida:
Points 1. and 2. added and tested. For 3., I refactored the existing test and added the another test, both using `definition.validate(df)` but with a nomenclature config file setting the validation parameters (default or otherwise). Let me know if something else was in mind.

danielhuppmann (approved):
Looks good to me, thanks!

PR #489, opened 2025-04-02
Add filter to `CodeList`

### LINKED ISSUE #417: Allow filtering by list of values and wildcard
The current implementation of **CodeList.filter()** allows to select by a tag-value combination - however, it is not possible to select by a list, and it is not possible to do wildcard-matching (like pyam).

For example, it would be useful to support the following
```python
RegionCodeList.filter(hierarchy=["R5", "R9"])
VariableCodeList.filter(variable="Final Energy|*")
```

dc-almeida:
Closes #417
Allows filtering directly from `CodeList` objects. AND logic for multiple code attributes, OR logic for multiple same-attribute values.
Refactors filter function implemented in #396 to avoid code duplication.


phackstock (changes requested):
I'd suggest the following changes:
1. Move `nomenclature.utils.filter_codes` into `Codelist.filter_codes(codes: list[code], include: None|dict=None, exclude: None|dict=None) -> list[code]` as a staticmethod.
2. Remove `CodeListFromRepository.filter_list_of_codes`
3. Change line 213 and 575 in codelist.py from `code_list.extend(repo.filter_list_of_codes(repository_code_list))` to `code_list.extend(cls.filter_list_of_codes(repository_code_list, repo.include, repo.exclude))`

This way you don't need utils.py and can remove code from config.py.

phackstock (approved):
Good to be merged, thanks @dc-almeida for all the changes :+1:

PR #491, opened 2025-04-03
Don't show index-column for data-validator xlsx output

danielhuppmann:
For a nicer look, this PR removes the index-column from the xlsx-output of the **DataValidator**.


dc-almeida (approved):
The initial thought was that the index could help locate the flagged rows easier, but if the output is large, it becomes less relevant than just listing the failed criteria or exporting the xlsx.
Makes sense to me!

PR #493, opened 2025-04-09
Automate region `countries` list for `parse_model_registration`

### LINKED ISSUE #359: Automate region descriptions from Excel to YAML
The `Region-Country-Mapping` sheet from model registration Excel files maps regions to their constituent countries. This information could be processed automatically to build the list of countries in each region's description in the YAML files.

https://github.com/IAMconsortium/nomenclature/blob/bc5a1337317a97dd6fdeb9bf2fdcec745d4c5e9a/nomenclature/__init__.py#L50-L88

@phackstock @danielhuppmann

dc-almeida:
Closes #359. Allows `parse_model_registration` to automatically create a `countries` attribute in region definition YAML files by parsing the `Region-Country-Mapping` sheet. If not present or empty, defaults to format without `countries`.
Also experimented with using GitHub Copilot code review after initial PR commit.


nomenclature/__init__.py
```diff
@@ -63,24 +64,55 @@ def parse_model_registration(
...
+    native = "Native region (as reported by the model)"
+    constituents = "Country name"
+    try:
+        region_country_mapping = pd.read_excel(
+            model_registration_file,
+            sheet_name="Region-Country-Mapping",
+            header=2,
+            usecols=[native, constituents],
+        )
+        region_country_mapping = region_country_mapping.where(
```
phackstock, line 88: I'm pretty sure that these three lines can be removed. In line 92, you drop all the na values anyway so the replacement with `None` does not have any effect. I quickly tested it locally and without those lines, the test still passes.
```suggestion
```

nomenclature/__init__.py
```diff
@@ -63,24 +64,55 @@ def parse_model_registration(
     """
     if not isinstance(output_directory, Path):
         output_directory = Path(output_directory)
-
-    region_aggregregation_mapping = RegionAggregationMapping.from_file(
+    # parse Common-Region-Mapping
+    region_aggregation_mapping = RegionAggregationMapping.from_file(
```
phackstock, line 68: Good catch on the typo :+1:

tests/data/region_processing/region_aggregation/excel_model_registration.xlsx
phackstock, line 1: I'd suggest adding at least one more constituent country to `Region 1` so that you can test if truly all the constituent regions end up in the list.

nomenclature/__init__.py
```diff
@@ -63,24 +64,55 @@ def parse_model_registration(
...
+            usecols=[native, constituents],
+        )
+        region_country_mapping = region_country_mapping.where(
+            pd.notnull(region_country_mapping), None
+        )
+        region_country_mapping = (
+            region_country_mapping.dropna().groupby(native)[constituents].apply(list)
+        )
+    except ValueError:
+        region_country_mapping = pd.DataFrame()
```
phackstock, line 95: Are you anticipating a specific error here and is there a reason for letting the error pass silently?
I'd suggest at least raising a warning in case something did not work so that the user is not surprised when the countries are missing. You can always cast the error to a string and emit is as a log message with the appropriate level, probably warning would be good.

phackstock (changes requested):
Thanks for the useful addition @dc-almeida. Couple of comments below.

nomenclature/__init__.py
dc-almeida: Yeah, to catch excel templates without the region-country sheet. I guess it would also catch some errors of an existing sheet but misformatted, so the warning can clarify that no such sheet was processed.

phackstock: Ah ok, then I'd suggest two different things:
1. Check explicitly for the presence of the `"Region-Country-Mapping"` sheet. I agree, that it's ok if it's missing so I'd just write out a warning or even just info level log message.
2. If the `"Region-Country-Mapping"` sheet is present, it needs to be correct. I'd suggest removing the try except clause and failing if there's any issue.

phackstock (approved):
Thanks for the quick updates @dc-almeida. Good to be merged from my side.

danielhuppmann (approved):
Nice work!

PR #495, opened 2025-04-17
Fix region aggregation skip

### LINKED ISSUE #483: skip-region-aggregation: true is ignored for single consituent common regions
I just encountered a case for a project where we had a variable where we decided to turn off region aggregation:

```yaml
- Power Generation Technologies|Coal fired:
    unit: '% Shares in Electricity production'
    description: Share of coal fired in electricity production
    skip-region-aggregation: true
...
``` 

in the model mapping we have two types of common regions, single- and multi-constituent:

```yaml
  - Reforming Economies (R5):
    - RUS
  - Asia (R5):
    - CHN
    - IND
    - KOR
    - IDN
```

in the processing results we now find no `Power Generation Technologies|Coal fired` for `Asia (R5)`. This is expected and desired since we turned regional aggregation off. 
For `Reforming Economies (R5)`, however, since the aggregation in this case is just a trivial renaming, we keep the `Power Generation Technologies|Coal fired` leading to some unexpected variables in some regions in the result.

@danielhuppmann, while I think this behavior is not technically wrong, I'd say it breaks the principle of least surprise and therefore we should also skip "region-aggregation" in single-constituent common regions, what do you think?

dc-almeida:
Closes #483.

According to the documentation of `process`, the expected results are:
> Region-processing, which can consist of three parts:
> * Model native regions not listed in the model mapping will be dropped
> * Model native regions can be renamed
> * Aggregation from model native regions to "common regions"

`test_region_processing_skip_aggregation` had as expected results:
```
# Testing two cases:
# * model "model_a" renames native regions and the world region is skipped
# * model "model_b" renames single constituent common regions
```

```yaml
model: model_a
native_regions:
  - region_A
  - region_B
common_regions:
  - World:
    - region_A
    - region_B
```

For `model_a`, the expected result is skipping the World region, _no_ native renaming,
but keeping the native regions in the data.

For `model_b`, mappings are:
```yaml
model: model_b
common_regions:
  - region_A:
    - region_A
  - region_B:
    - region_b
```
We should expect:
1. no native regions (as they are not listed in the mappings);
2. no renamed native regions (because of 1.);
3. aggregation to common regions (in this case, single constituent common regions,
effectively acting as a renaming).

Given `skip-region-aggregation: true` for `Primary Energy`, the final DataFrame
shouldn't include any data for this variable for common regions. Thus, the expected
result is an empty DataFrame.

As it was previously implemented, it would return a DataFrame of renamed regions,
thus breaking the principle of least surprise, as pointed out by @phackstock.

This PR fixes `skip-region-aggregation` behaviour for RegionProcessor,
excluding common regions in the final data when aggregation is skipped, by:
* making the aggregation pipeline the same for single and multiple constituent
common regions
* excluding single constituent common regions with the same name as their native
region

This is likely a breaking change for workflow repos, since, previously, it
preserved native region data, assuming it to be single constituent common region
data (if the names were the same). Now, that data is excluded, and can lead to a
`ValueError` for an empty dataset.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -260,7 +260,8 @@ def test_region_processing_weighted_aggregation(folder, exp_df, args, caplog):
 def test_region_processing_skip_aggregation(model_name, region_names):
     # Testing two cases:
     # * model "model_a" renames native regions and the world region is skipped
-    # * model "model_b" aggregates single constituent common regions, which are skipped
+    # * model "model_b" aggregates single constituent common regions, which raises
+    # a ValueError due to an empty dataset
 
     test_df = IamDataFrame(
         pd.DataFrame(
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -257,10 +257,10 @@ def test_region_processing_weighted_aggregation(folder, exp_df, args, caplog):
     "model_name, region_names",
     [("model_a", ("region_A", "region_B")), ("model_b", ("region_A", "region_b"))],
 )
-def test_region_processing_skip_aggregation(model_name, region_names):
+def test_region_processing_skip_aggregation(model_name, region_names, caplog):
     # Testing two cases:
     # * model "model_a" renames native regions and the world region is skipped
-    # * model "model_b" renames single constituent common regions
+    # * model "model_b" aggregates single constituent common regions, which are skipped
 
     test_df = IamDataFrame(
         pd.DataFrame(
@@ -284,16 +284,19 @@ def test_region_processing_skip_aggregation(model_name, region_names):
     )
     add_meta(exp)
 
-    obs = process(
-        test_df,
-        dsd := DataStructureDefinition(
-            TEST_DATA_DIR / "region_processing/skip_aggregation/dsd"
-        ),
-        processor=RegionProcessor.from_directory(
-            TEST_DATA_DIR / "region_processing/skip_aggregation/mappings", dsd
-        ),
+    dsd = DataStructureDefinition(
+        TEST_DATA_DIR / "region_processing/skip_aggregation/dsd"
     )
-    assert_iamframe_equal(obs, exp)
+    processor = RegionProcessor.from_directory(
+        TEST_DATA_DIR / "region_processing/skip_aggregation/mappings", dsd
+    )
+    if model_name == "model_b":
+        with pytest.raises(ValueError):
+            obs = process(test_df, dsd, processor=processor)
+            assert "returned an empty dataset" in caplog.text
+    else:
+        obs = process(test_df, dsd, processor=processor)
+        assert_iamframe_equal(obs, exp)
 
 
 @pytest.mark.parametrize(
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -257,7 +257,7 @@ def test_region_processing_weighted_aggregation(folder, exp_df, args, caplog):
     "model_name, region_names",
     [("model_a", ("region_A", "region_B")), ("model_b", ("region_A", "region_b"))],
 )
-def test_region_processing_skip_aggregation(model_name, region_names, caplog):
+def test_region_processing_skip_aggregation(model_name, region_names):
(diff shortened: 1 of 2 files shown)


danielhuppmann:
Thanks @dc-almeida, but I'm confused why this would raise an error. Shouldn't we expect that an empty dataframe is ok?

danielhuppmann:
> The error raising is not added in the PR, it is the default behaviour of the region processor whenever processing results in an empty dataframe. I can change it to log a warning and return an empty dataframe.

Ok, keep as is for now, and then let's discuss with @phackstock whether that design behavior makes sense.

tests/test_core.py
```diff
@@ -284,16 +285,19 @@ def test_region_processing_skip_aggregation(model_name, region_names):
...
-            TEST_DATA_DIR / "region_processing/skip_aggregation/mappings", dsd
-        ),
+    dsd = DataStructureDefinition(
+        TEST_DATA_DIR / "region_processing/skip_aggregation/dsd"
     )
-    assert_iamframe_equal(obs, exp)
+    processor = RegionProcessor.from_directory(
+        TEST_DATA_DIR / "region_processing/skip_aggregation/mappings", dsd
+    )
+    if model_name == "model_b":
```
phackstock, line 294: I think this makes the test rather hard to read having to follow those two different branches. I'd suggest, in analogy to many other tests in nomenclature, to split this into two tests. One that expects a certain data frame back and the other one that expects the error.

phackstock (changes requested):
Sorry for the long delay, finally now my review.
Thanks for the implementation @dc-almeida and taking into account all the different single constituent region possibilities. I we're currently making our lives a little harder than they need to be. The case of a single constituent common region that is being "aggregated" to a region of the same name is one that I don't think exists outside of the test case (that I admittedly originally wrote). 
My proposal to simplify the implementation in `region.py` would therefore be to truly treat single and multi constituent regions the same for aggregation while disallowing the case of:
```yaml
...
common_regions:
  - region_A:
    - region_A
```

if we now have a model with a `region_A` in the above example, we just treat is as a model native region. Which I'd argue it is anyway at this point.
Happy to discuss further on that.

Regarding the question of whether or not an error should be raised if the processing returns an empty frame, I am in favor of that. Admittedly it's a little on the cautious side but it prevents the case where an upload on a Scenario Explorer has a "green" status but no data has been uploaded.

dc-almeida:
Committed changes change single constituent logic, to skip common regions which have equal native region (they are excluded from aggregate df). Updated tests as requested and added one to clarify this skip.

danielhuppmann:
Thanks @dc-almeida, that a useful illustration of a corner case.

My suggestion would be the following, also going in the direction of @phackstock's intuition:
1. Explicitly disallow the situation where a common-region has a mapping where the source and the target are identical (as a new separate PR)

Then rebase this PR and implement the following logic
2. Keep the current logic that if data is provided for a common-region, that is indeed kept as is, as the intended override-functionality
3. Filter out any skip-region-aggregation-variables during the renaming of a single-constituent common-region processing

danielhuppmann:
Thanks @dc-almeida for following this rabbit hole all the way to the end!

I would go for option 2 (adding `region_a` to the code list and adding the following comments to the test)

```yaml
model: model_c
native_regions:
  # this is a native region that is kept as-is during processing (no renaming)
  - region_a
  # this is a common region that is reported as-is by the model output
  - region_B
common_regions:
  # this is a common region that is renamed from the native-region model output 
  # variables marked as `skip-region-aggregation` are removed during region-processing
  - region_A: [region_a]
exclude_regions:
  - region_C
```

dc-almeida:
As suggested by @phackstock, changed one of the tests' mappings to ensure it still yields an equivalent result as before.

danielhuppmann (approved):
Thank you, looks great! Please still add empty lines at the end of the new/updated yaml files, then good to merge (in spite of failing tests, make a release, then update common-definitions.

PR #496, opened 2025-05-12
Add `rename` to `Aggregator`

### LINKED ISSUE #492: Add a `rename` option for the Aggregator
The **Aggregator** class currently accepts only a config-mapping like 

```yaml
dimension: variable
aggregate:
  - Primary Energy:
    - Primary Energy|Coal
    - Primary Energy|Biomass
  - Final Energy:
    - Final Energy|Electricity
    - Final Energy|Heat
```

For some simple use cases (like the scenario-migration-tool by @lisahligono), we should also support a simple rename-feature, like
```yaml
dimension: unit
rename:
  - "?C": "°C"
  - "??C": "°C"
  - "?C-days": "°C-days"
```

[I was planning on doing this myself, but I realize that I won't get to it before the summer]

dc-almeida:
Closes #492
Adds a `rename` option to the `Aggregator` config file as specified in the issue.
Adds sanity check to avoid conficts between aggregation and renaming.
Updates tests accordingly.

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -211,6 +211,7 @@ class VariableCode(Code):
             return v
 
     @field_validator("unit", mode="before")
+    @classmethod
     def convert_none_to_empty_string(cls, v):
         return v if v is not None else ""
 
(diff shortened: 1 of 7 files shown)


danielhuppmann (approved):
Thank you, sorry for the long delay with the review. I tested some corner cases locally and it behaves just as expected.

PR #504, opened 2025-07-17
Fix illegal characters validation

dc-almeida:
Fix field definition that was overriding option to ignore illegal characters.
Added test to reflect this case.

### DIFF
--- a/tests/test_codelist.py
+++ b/tests/test_codelist.py
@@ -294,12 +294,24 @@ def test_stray_tag_fails(subfolder, match):
         code_list.check_illegal_characters(NomenclatureConfig(dimensions=["variable"]))
 
 
-def test_illegal_char_fails():
-    """Check that illegal character raises expected error."""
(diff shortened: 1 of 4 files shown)


tests/test_codelist.py
```diff
@@ -294,12 +294,24 @@ def test_stray_tag_fails(subfolder, match):
...
 
 
-def test_illegal_char_fails():
-    """Check that illegal character raises expected error."""
-    match = r"Illegal character\(s\) '\"' in info of variable 'Primary Energy\|Coal'"
-    with raises(ValueError, match=match):
-        DataStructureDefinition(
-            MODULE_TEST_DATA_DIR / "illegal_chars" / "char_in_str" / "definitions"
+@pytest.mark.parametrize("subfolder", ["char_in_str", "no_chars"])
+def test_illegal_chars_raise_or_ignore(subfolder):
```
phackstock, line 298: This should be two different tests to make it easier to see what you are testing for.

phackstock (changes requested):
One small change other than that, good to be merged

PR #505, opened 2025-07-21
Add "name" attribute to DataValidationItem

danielhuppmann:
In preparation for a new feature to set meta indicators from data-validation, this PR went down some rabbit holes to clean up the code and tests so that the actual new feature will be easier to review.
1. Add an attribute "name" to the `DataValidationItem` (which doesn't yet have any effect)
2. Rename the existing tests for `validate_data` to a consistent input file structure like `validate_fail_[warning_][type].yaml` where "warning" is for those tests that have non-error-conditions, i.e., warning-level other than error.
3. Move the "ranges" test from `test_DataValidator_validate_fail_with_warning` to `test_DataValidator_apply_fails` because it does not have non-error warning levels.


phackstock (approved):
Always a big fan of cleaning up :+1:, good to be merged from my side.

danielhuppmann:
FYI @phackstock & @dc-almeida, the final commit extends the range-validation test case so that is identical to the bounds- and value-tests, so that some of the changes in the test aren't necessary anymore.

PR #509, opened 2025-07-24
Rename "time" to "time-domain" in `NomenclatureConfig` and update tests

dc-almeida:
For consistency with pyam naming convention


dc-almeida:
Ah thanks, well spotted. Was missing a change in a configuration option in `ConfigDict`. Committing that needed change where it was missing

phackstock (approved):
Looks good to me, thanks @dc-almeida.

PR #519, opened 2025-09-17
Limit click to <=8.1.8

phackstock:
As I was working on removing some warnings, I saw that the latest version of click breaks the tests. The reason is that the root cli command `nomenclature` now no longer returns 0 but an error.
For now I've pinned click to <=8.1.8


danielhuppmann (approved):
Thank you!

PR #532, opened 2025-10-14
Create a DataValidator from a DataStructureDefinition

danielhuppmann:
This PR implements a class-method to create a **DataValidator** from the validation-attributes of a **VariableCodeList**. These attributes should only be used for general properties of a variable, like a range being between 0 and 100 or a variable like GDP being non-negative.

In preparation for a follow-up PR, this PR also renames `validate_data()` to `validate_df`, because this checks that the codes used for variables, regions, ... of an **IamDataFrame** are in line with the codelists, not actually checking the data.

~Any thoughts whether we should rename the `file` property of the **DataValidator** to something more generic, like `source`?~ 
UPDATE: will leave that for a follow-up PR.

### DIFF
--- a/nomenclature/processor/data_validator.py
+++ b/nomenclature/processor/data_validator.py
@@ -291,7 +291,7 @@ class DataValidator(Processor):
         ]
         return cls(
             file="definitions", criteria_items=criteria_items, output_path=output_path
-        )  # type: ignore
+        )
 
     def apply(self, df: IamDataFrame) -> IamDataFrame:
         """Validates data in IAMC format according to specified criteria.
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -290,7 +290,7 @@ class VariableCode(Code):
     @property
     def has_validation_args(self) -> bool:
         return any(
-            [attribute in VALIDATION_ARGS for attribute in self.extra_attributes]
+            attribute in VALIDATION_ARGS for attribute in self.extra_attributes
         )
 
     @property


nomenclature/processor/data_validator.py
```diff
@@ -276,6 +286,32 @@ def from_file(
...
         return cls(file=file, criteria_items=criteria_items, output_path=output_path)  # type: ignore
 
+    @classmethod
+    def from_definition(
+        cls, definition: DataStructureDefinition, output_path: Path | str | None = None
+    ) -> "DataValidator":
+        criteria_items = []
+
+        for name, variable in definition.variable.items():
+            if any([i in VALIDATION_ARGS for i in variable.extra_attributes]):
```
phackstock: Personal preference against since letter variable names:

```suggestion
            if any([attribute in VALIDATION_ARGS for attribute in variable.extra_attributes]):
```

nomenclature/codelist.py
```diff
@@ -700,14 +700,14 @@ def validate_units(
...
-    def validate_data(
+    def validate_df(
         self,
         df: IamDataFrame,
         dimension: str,
         project: str | None = None,
     ) -> bool:
         # validate variables
-        all_variables_valid = super().validate_data(df, dimension, project)
+        all_variables_valid = super().validate_df(df, dimension, project)
```
phackstock:
```suggestion
        all_variables_valid = super().validate_dataframe(df, dimension, project)
```

danielhuppmann:
Thanks @phackstock for the review.

> One thing that might be nice would be to directly include the DataValidator in the VariableCodeList

Added that, but named it ` data_validator`  in line with the module name. Or do you think the version without an underscore is much better?

About the new workflow (per your suggestion), I think that running the DataValidator as part of ` definition.validate()` should be the default option anyway (rather than adding it manually to every workflow).

nomenclature/processor/data_validator.py
danielhuppmann: Done

nomenclature/processor/data_validator.py
```diff
@@ -276,6 +286,32 @@ def from_file(
...
+                criteria_items.append(
+                    dict(
+                        variable=name,
```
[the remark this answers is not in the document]
danielhuppmann: Done.

nomenclature/codelist.py
```diff
@@ -91,7 +91,7 @@ def keys(self):
     def values(self):
         return self.mapping.values()
 
-    def validate_data(
+    def validate_df(
```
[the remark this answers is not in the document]
danielhuppmann: Disagree that this is easier to read, and it may add confusion because *dataframe* is primarily associated with pandas, not pyam. In the codebase, we usually use ` df`  for an **IamDataFrame**, so I prefer to keep as is.

nomenclature/codelist.py
danielhuppmann: As above

nomenclature/processor/data_validator.py
```diff
@@ -276,6 +287,37 @@ def from_file(
...
+                                    for key, value in variable.extra_attributes.items()
+                                    if key in VALIDATION_ARGS
+                                ]
+                            )
+                        ],
+                    )
+                )
+        return cls(
+            file="definitions", criteria_items=criteria_items, output_path=output_path
+        )  # type: ignore
```
phackstock: I think the reason the type checker complains is because the type hint in the function inputs is incorrect. I would fix that rather than tell the type checker to ignore it.

PR #535, opened 2025-11-06
Add check for infinite values in `process()`

### LINKED ISSUE #387: Check if data contains inf values
We could implement a validator that checks if a given data set contains any inf values.

dc-almeida:
Closes #387
A simple check for infinite values in the `process` function used for workflow repos.


phackstock (changes requested):
Thanks for the quick PR @dc-almeida. Two small changes required due to a misunderstanding. Other than that good to be merged from my side.

PR #540, opened 2025-11-13
Allow scse-toolkit <0.16.0

phackstock:
This is required for to https://github.com/iiasa/scse-processing/issues/113 since https://github.com/iiasa/scse-toolkit/pull/47 introduced the required impersonation token feature. This is only available from toolkit version 0.15


danielhuppmann (approved):
Looks ok to me.

PR #543, opened 2025-11-24
Add check of `VariableCodeList.data_validator()` to project-validation

### LINKED ISSUE #542: Potential issue with common-definitions as validation criteria
Not sure about the exact source but this is the error that I currently get when trying to use latest nomenclature with the common-definitions.

```console
criteria_items.0.validation.0.DataValidationValue.value
  Field required [type=missing, input_value={'range': '0, 100'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
criteria_items.0.validation.0.DataValidationValue.range
  Extra inputs are not permitted [type=extra_forbidden, input_value='0, 100', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
criteria_items.0.validation.0.DataValidationRange.range
  Input should be a valid list [type=list_type, input_value='0, 100', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/list_type
criteria_items.0.validation.0.function-after[check_validation_multiple_criteria(), function-after[check_validation_criteria_exist(), DataValidationBounds]]
  Value error, No validation criteria provided: {'range': '0, 100'} [type=value_error, input_value={'range': '0, 100'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/value_error
```

danielhuppmann:
Closes #542.
This PR adds initializing the **VariableCodeList.data_validator** when running the CLI `nomenclature validate-project .` to guard against malformatted validation items.

Not sure why the test returns a SystemExist (2) - any ideas @phackstock?

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -176,8 +176,7 @@ def test_cli_wrong_definitions_name():
 
 
 def test_cli_variable_validation_item_invalid():
-    """Check that CLI raises expected error when a non-existing non-default directory
-    is given"""
+    """Check that CLI raises expected error when malformatted validation item is given"""
     result = runner.invoke(
         cli,
         [


nomenclature/testing.py
```diff
@@ -144,6 +144,7 @@ def assert_valid_structure(
         )
 
     dsd = DataStructureDefinition(path / definitions, dimensions)
+    dsd.variable.data_validator
```
phackstock: I think this will fail if we don't have the variable dimension. Apart from this, I think the line technically works but only in a somewhat hard to read way. I'd make it more explicit and therefor easier to read, something like:
```python
if dsd.variable:
    DataValidator.from_codelist(dsd.variable)
```

tests/data/cli/variable_invald_validation_item/definitions/variable/variables.yaml
phackstock: The path of the file should be `tests/data/cli/variable_invalid_validation_item/definitions/variable/variables.yaml`

tests/data/cli/variable_invalid_validation_item/definitions/variable/variables.yaml
danielhuppmann: Thanks for spotting this!

tests/test_cli.py
```diff
@@ -175,6 +175,19 @@ def test_cli_wrong_definitions_name():
...
+    """Check that CLI raises expected error when a non-existing non-default directory
+    is given"""
+    result = runner.invoke(
+        cli,
+        [
+            "validate-project",
+            str(MODULE_TEST_DATA_DIR / "variable_invalid_validation_item"),
+        ],
+    )
+    assert result.exit_code == 2
```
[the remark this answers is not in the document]
danielhuppmann: Works now, and I added the validation of the expected string

phackstock:
> Maybe to be tackled in another PR, but the VariableCodeList is defined in codelist.py, which is already quite long, and the DataValidator inherits from Processor and it makes sense that all such classes are defined in the processor subfolder...

Fair point, that is a topic that could get more complicated rather quickly...

Issue #547, opened 2025-11-29
Support attribute-tag-replacement with `int` or None


danielhuppmann:
Replacing a tag in an attribute currently only works with strings, so

```yaml
- Emissions|{Level-1 Species}:
    description: Emissions of {Level-1 Species} to the atmosphere
    unit: "{Level-1 Species}"
    tier: 1
    lower_bound: "{Level-1 Species}"
```

works with 
```yaml
- Level-1 Species:
    - CO2:
        description: carbon dioxide (CO2)
        unit: Mt CO2/yr
        lower_bound: some string
```

but not if *lower_bound* in the tag list is an integer or None.

I suggest the following approach: if an attribute of a code is only the tag (like in the example above), directly use the value from the tag-list for that attribute.

Required for https://github.com/IAMconsortium/common-definitions/issues/339

PR #548, opened 2025-11-29
Skip **DataValidatonItem** in a VariableCodeList when values are None

danielhuppmann:
Currently,, having an item in a **VariableCodeList** with a validation-arg as `None` like 
```yaml
- Emissions|CO2:
    description: Emissions of CO2 to the atmosphere
    unit: Mt CO2/yr
    lower_bound:
```
fails even though **pyam.IamDataFrame.validate()** (which is called downstream) would accept it and simply not apply a lower bound.

This PR adds a quickfix so that no **DataValidationItem** is created in the **VariableCodeList._datavalidator** if all `VALIDATION_ARGS` are None, and extends the test.

Related to #547

### DIFF
--- a/tests/data/processor/data_validator/init_from_codelist/variable.yaml
+++ b/tests/data/processor/data_validator/init_from_codelist/variable.yaml
@@ -9,4 +9,4 @@
 - Emissions|CO2:
     description: Emissions of carbon dioxide (CO2)
     unit: Mt CO2/yr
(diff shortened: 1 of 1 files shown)


phackstock:
Ah, got it. You have to include the `lower_bound` attribute in the for the `CO2` tag even though it is `None` because otherwise the tag replacement would fail. Sounds good to me.

PR #550, opened 2025-12-10
Add model registration parsing to cli

### LINKED ISSUE #549: Turn `parse_model_registration` into CLI command
This https://github.com/IAMconsortium/nomenclature/blob/b43ee539fff354d4e064e1047743ec3fecc807ba/nomenclature/__init__.py#L68C5-L68C29 should be a cli command.

phackstock:
Closes #549.

This PR also moves `convert_xlsx_to_yaml` to the cli and renames it to `convert_xlsx_codelist_to_yaml`. Because we are using typer now for the cli, you can still import the functions and run them as part of a script in addition to the cli.
You can do something like:

```python
from pathlib import Path
from nomenclature.cli import parse_model_registration

parse_model_registration(Path("model_registration_file.xlsx"))
```

in addition to:

```console
nomenclature parse_model_registration model_registration_file.xlsx
```

I also replaced the spaces in the file names with underscores.

### DIFF
--- a/tests/test_region_aggregation.py
+++ b/tests/test_region_aggregation.py
@@ -331,7 +331,7 @@ def test_reverse_region_aggregation():
 
 
 def test_model_mapping_from_excel():
-    excel_file = TEST_FOLDER_REGION_AGGREGATION / "excel_model_registration.xlsx"
+    excel_file = TEST_DATA_DIR / "model_registration" / "excel_model_registration.xlsx"
     obs = RegionAggregationMapping.from_file(excel_file)
     exp = RegionAggregationMapping(
         model=["Model 1.1"],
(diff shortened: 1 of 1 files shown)


dc-almeida (approved):
Definitely makes more sense than how it was set up before, and great that typer's syntax means the function is accessible besides its cli command form.

nomenclature/cli.py
```diff
@@ -269,3 +271,118 @@ def validate_scenarios(
...
+    ),
+    mappings_path: Annotated[Path, typer.Option(exists=True)] = Path.cwd() / "mappings",
+) -> None:
+    """Parses a model registration file and writes the definitions & mapping yaml files
+
+    Parameters
+    ----------
+    model_registration_file : path, file-like object
+        Path to xlsx model registration file.
+    region_path : path
```
danielhuppmann: Suggest to rename for clarity.
```suggestion
    definition_path : path
```

PR #562, opened 2026-02-24
Fix repository URL handling to re-clone on URL change

### LINKED ISSUE #541: Changing git-url is ignored if folder already cloned
Changing the url of a GitHub-repo in nomenclature.yaml is ignored if the folder already exists with a valid cloned git repo.

dc-almeida:
Closes #541

### DIFF
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -119,7 +119,7 @@ class Repository(BaseModel):
                     f"Re-cloning repository to '{to_path}'..."
                 )
                 repo.close()  # Close repo before removing directory
-                repo.__del__()  # Force cleanup of git objects
+                del repo  # Delete reference to allow garbage collection
                 gc.collect()  # Force garbage collection to release file handles
 
                 rmtree(to_path, onerror=handle_remove_readonly)


nomenclature/config.py
```diff
@@ -109,7 +110,16 @@ def fetch_repo(self, to_path):
             repo = Repo.clone_from(self.url, to_path)
         else:
             repo = Repo(to_path)
-            repo.remotes.origin.fetch()
+            # If the URL has changed, remove existing directory and re-clone
+            if origin_url := repo.remotes.origin.url != self.url:
```
phackstock: `origin_url` will always resolve to a boolean so the logging info below will read "_Repository URL changed from True ..._" instead of reporting the original repository url:

```suggestion
            if repo.remotes.origin.url != self.url:
```

nomenclature/config.py
```diff
@@ -109,7 +110,16 @@ def fetch_repo(self, to_path):
...
+                logger.info(
+                    f"Repository URL changed from '{origin_url}' to '{self.url}'. "
```
```suggestion
                    f"Repository URL changed from '{repo.remotes.origin.url}' to '{self.url}'. "
```

phackstock (changes requested):
Thanks for the quick PR @dc-almeida, some comments below to fix the logging information. If you want to be thorough, I'd also suggest adding a test.

PR #569, opened 2026-03-10
Add `NutsProcessor` for NUTS region aggregation

### LINKED ISSUE #563: RegionProcessor for NUTS regions
I wanted to hold back this idea until more of the refactoring is done, but because the EXPANSE model requested a registration for this in the NEWPATHWAYS project...

To avoid having to add mapping files with instructions for the NUTS regions, I think that we should have a new **NutsRegionAggregationMapping** class that parses the **pysquirrel** package to instantiate a **RegionAggregationMapping** from NUTS3 to country-level and EU27 aggregate.

My initial hunch would be that this can be activated in the `nomenclature.yaml` file like
```yaml
mappings:
  repository:
    name: common-definitions
    include:
      - <some-model>
  nuts-processor:
    - <model-with-NUTS-regions>
```

dc-almeida:
Closes #563. `NutsProcessor` performs aggregation based on the NUTS regions available in the dataframe and defined in the project configuration.
If there are regions in the data that were not previously defined in `nomenclature.yaml`, it raises (e.g.: if NUTS regions for Belgium are in the data, but BE is not listed in the config file).
If regions within a NUTS hierarchy are missing (e.g.: for AT11, AT111 is present but AT112 is missing), aggregation runs with the existing ones.

- Extracted some of the aggregation logic in `RegionProcessor` into helper functions that are shared with `NutsProcessor`
- Added API documentation and a brief guide


nomenclature/config.py
```diff
@@ -60,6 +72,12 @@ def repository_dimension_path(self) -> str:
 
 
 class RegionCodeListConfig(CodeListConfig):
+    """
+    Configuration for a region's codelist.
```
phackstock: I find the doc string here slightly misleading. Since it inherits from `CodeListConfig`, I'd argue it's mainly used to select which regions to import from an external repository.
```suggestion
    Configuration for a region codelist.
```

nomenclature/config.py
```diff
@@ -150,13 +169,16 @@ def check_external_repo_double_stacking(self):
...
-    """A class for configuration of a DataStructureDefinition
+    """
+    Configuration class for the data structure definition.
 
-    Attributes
-    ----------
-    region : RegionCodeListConfig
-        Attributes for configuring the RegionCodeList
+    This class defines the configuration for the main IAMC dimensions:
+    - model
```
phackstock: Looking at the git blame, I realize that I introduced this but we don't have ModelCodeLists so this should be removed.

nomenclature/config.py
```diff
@@ -150,13 +169,16 @@ def check_external_repo_double_stacking(self):
...
+    - scenario
+    - region
+    - variable
 
+    Each dimension can be configured with its own code list and repository sources.
     """
 
     model: CodeListConfig = Field(default_factory=CodeListConfig)
```
phackstock: Since `ModelCodeList` is not a thing this should be removed.
```suggestion
```

nomenclature/config.py
```suggestion
            for dimension in ("scenario", "region", "variable")
```

```suggestion
    @field_validator("scenario", "region", "variable", mode="before")
```

nomenclature/config.py
```diff
@@ -337,6 +380,15 @@ def check_definitions_repository(
                 raise ValueError((f"Unknown repository {unknown_repos} in '{use}'."))
         return v
 
+    @model_validator(mode="after")
+    @classmethod
+    def check_nuts_consistency(cls, v: "NomenclatureConfig") -> "NomenclatureConfig":
+        if v.processor and v.processor.nuts and not v.definitions.region.nuts:
```
phackstock: One of the reasons that I started using the `default_factory` pattern is so that all attributes exist and you don't have to explicitly check that in this case `v.processor` exists:
```suggestion
        if v.processor.nuts and not v.definitions.region.nuts:
```

nomenclature/config.py
```diff
@@ -337,6 +380,15 @@ def check_definitions_repository(
...
+            raise ValueError(
+                "`nuts` region processor set but no NUTS regions in `definitions`."
```
phackstock: It would be nice if the error message contains the exact fix, something like:
```suggestion
                "`nuts` region processor set but no NUTS regions in `definitions`. To fix, set definitions.regions.nuts to True."
```

nomenclature/core.py
```diff
@@ -56,8 +59,30 @@ def process(
 
     dimensions = dimensions or dsd.dimensions
 
+    # Auto-instantiate processors declared in nomenclature.yaml under 'processors'
+    # Explicit processors take precedence; config-based ones are appended after.
+    if dsd.config.processor.region_processor:
+        if any(isinstance(p, RegionProcessor) for p in processor):
+            logger.info(
```
phackstock: I'm inclined to convert this to a proper error reason being that it's easy enough to skip reading the log when an upload is successful and a situation like this might break the principle of least surprise.

nomenclature/core.py
```diff
@@ -56,8 +59,30 @@ def process(
...
+                "Config declares 'region-processor: true' but an explicit "
+                "RegionProcessor was provided -- skipping config-defined processor."
+            )
+        else:
+            processor = processor + [
```
phackstock: Suggest to use the `.append()` method directly:
```suggestion
            processor.append(RegionProcessor.from_directory(dsd.project_folder / "mappings", dsd))
```

nomenclature/core.py
```diff
@@ -56,8 +59,30 @@ def process(
...
+                RegionProcessor.from_directory(dsd.project_folder / "mappings", dsd)
+            ]
+
+    if dsd.config.processor.nuts is not None:
+        if any(isinstance(p, NutsProcessor) for p in processor):
+            logger.info(
```
phackstock: Same suggestion as above, I'd raise an error if we're trying to overwrite a `NutsProcessor`

tests/data/nuts_processing/dsd/definitions/region/regions.yaml
```diff
@@ -0,0 +1,32 @@
+- Country:  # Minimal region list - NUTS regions will be auto-imported
```
phackstock: For the test you can use `region.country = True`, then you don't have to redefine all the countries here.

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+            raise ValueError("No models configured for NUTS processor")
+        return cls(
+            variable_codelist=dsd.variable, region_codelist=dsd.region, models=models
+        )
+
+    @property
+    def nuts_codelist(self):
+        return RegionCodeList(
+            name="NUTS",
+            mapping={
```
phackstock, line 103: The `CodeList` class implements a filter method you can use directly. Since you control how the nuts regions are added to the codelist and can set arbitrary attributes, I'd suggest adding a `nuts` attribute that you set to `true`. 
Then following should work:

```suggestion
            mapping=self.region_codelist.filter(nuts=True).mapping
```
This has the added benefit of avoiding the somewhat brittle regex and being easier to read.
You'd have to add a filter option for `bool` though in `CodeList.filter_codes.matches_filter.check_attribute_match`.

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+
+        return result
+
+    def _apply_nuts_processing(
+        self,
+        model_df: IamDataFrame,
+        return_aggregation_difference: bool = False,
+        rtol_difference: float = 0.01,
+    ):
+        if len(model_df.model) != 1:
```
phackstock: Since you're only feeding the data one model at a time anyway, I'd say you can remove this check.

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+    variable_codelist: VariableCodeList
+    region_codelist: RegionCodeList
+    models: list[str]
+
+    model_config = ConfigDict(hide_input_in_errors=True)
+
+    @classmethod
+    def from_definition(cls, dsd: DataStructureDefinition):
+        models = dsd.config.processor.nuts
+        if not models:
```
phackstock, line 93: This check means that the nuts processor is only ever intended to be "instantiated" from the `nomenclature.yaml` file. To initialize an instance of `NutsProcessor` I'll need to set `config.processor.nuts` to a list of models. Since this list of models then exists, the check in `core.py` on line 76 will be true. I'm not necessarily against that, just wanted to make sure that this is intended.

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+        rtol_difference: float = 0.01,
+    ):
+        if len(model_df.model) != 1:
+            raise ValueError(
+                f"Must be called for a unique model, found: {model_df.model}"
+            )
+        model = model_df.model[0]
+
+        # Check for NUTS regions not listed in the configuration
+        all_nuts = {r.code for r in nuts.get(level={1, 2, 3})}
```
phackstock: My type checker complains here that level should be a `list` not a `set`:
```suggestion
        all_nuts = {r.code for r in nuts.get(level=[1, 2, 3])}
```

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+        if unaccounted_nuts := (set(model_df.region) & all_nuts) - set(
```
phackstock: Since `self.nuts_codelist` is a fully functioning `RegionCodeList`, I'd suggest to use the `.validate_df()` method directly:

```suggestion
        self.nuts_codelist.validate_df(model_df.filter(region=all_nuts), "region")
```
I would also put this check at the beginning of the `.apply` function, there's no need to wait this long to check for unaccounted nuts regions.

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+                f"Did not find NUTS region(s) {unaccounted_nuts} in 'region.nuts' configuration."
+            )
+
+        _df = model_df.copy()
+        _processed_data: list[pd.Series] = []
+
+        # Silence pyam's empty filter warnings
+        with adjust_log_level(logger="pyam", level="ERROR"):
+            # NUTS3 > NUTS2 aggregation
+            if nuts3_in_data := (set(_df.region) & {r.code for r in nuts.get(level=3)}):
```
phackstock: I'd suggest to use use filter function again:
```suggestion
            if nuts3_in_data := _df.filter(region=nuts.get(level=3)):
```

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+                _processed_data = self._aggregate_nuts_level(_df, nuts3_in_data, 4)
+                # Keep NUTS3, add aggregated NUTS2
+                _df = pyam.concat([_df, IamDataFrame(pd.concat(_processed_data))])
+
+            # NUTS2 > NUTS1 aggregation (uses original NUTS2 + aggregated NUTS2)
+            if nuts2_in_data := (set(_df.region) & {r.code for r in nuts.get(level=2)}):
```
phackstock:
```suggestion
            if nuts2_in_data := _df.filter(region=nuts.get(level=2)):
```

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+                _processed_data = self._aggregate_nuts_level(_df, nuts2_in_data, 3)
+                # Keep NUTS2, add aggregated NUTS1
+                _df = pyam.concat([_df, IamDataFrame(pd.concat(_processed_data))])
+
+            # NUTS1 > Country aggregation (uses original NUTS1 + aggregated NUTS1)
+            if nuts1_in_data := (set(_df.region) & {r.code for r in nuts.get(level=1)}):
```
phackstock:
```suggestion
            if nuts1_in_data := _df.filter(region=nuts.get(level=1)):
```

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+        _df = model_df.copy()
+        _processed_data: list[pd.Series] = []
+
+        # Silence pyam's empty filter warnings
+        with adjust_log_level(logger="pyam", level="ERROR"):
+            # NUTS3 > NUTS2 aggregation
+            if nuts3_in_data := (set(_df.region) & {r.code for r in nuts.get(level=3)}):
+                _processed_data = self._aggregate_nuts_level(_df, nuts3_in_data, 4)
+                # Keep NUTS3, add aggregated NUTS2
+                _df = pyam.concat([_df, IamDataFrame(pd.concat(_processed_data))])
```
phackstock: I don't see a need to save `_processed_data` separately as a variable, plus I'd change the return type of `_aggregate_nuts_level` to `IamDateFrame`, then you don't have to `pd.concat` and cast every time you call `_aggregate_nuts_level`:
```suggestion
                _df = pyam.concat([_df, self._aggregate_nuts_level(_df, nuts3_in_data, 4)])
```

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+            raise UnknownRegionError(not_defined_regions)
+
+        return res
+
+    def _aggregate_nuts_level(
+        self,
+        model_df: IamDataFrame,
+        source_regions: list[str],
+        parent_prefix_length: int,
+    ) -> list[pd.Series]:
```
phackstock: Suggest to change the return type to `IamDataFrame` for easier downstream handling.
```suggestion
    ) -> IamDataFrame:
```

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+            )
+            aggregated = aggregate_region_with_variable_rules(
+                model_df,
+                parent,
+                constituents,
+                self.variable_codelist,
+            )
+            aggregated_data.extend(aggregated)
+
+        return aggregated_data
```
phackstock:
```suggestion
        return IamDataFrame(pd.concat(aggregated_data))
```

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+        parent_groups = defaultdict(list)
+        for source_region in source_regions:
+            parent = source_region[:parent_prefix_length]
+            parent_groups[parent].append(source_region)
+
+        # Aggregate each parent from its constituents
+        for parent_code, constituents in parent_groups.items():
+            parent = (
+                countries.get(alpha_2=parent_code).name
+                if len(parent_code) == 2  # If NUTS 1 > country, use name
```
phackstock: I'd swap the if around so that the option that's used most is covered first:
```python
parent = (
                parent_code 
                if len(parent_code) > 2  # If NUTS 1 > country, use name
                else countries.get(alpha_2=parent_code).name
            )
```

nomenclature/processor/region.py
```diff
@@ -694,90 +698,43 @@ def _apply_region_processing(
...
-                # treat as rename that filters skip-region-aggregation variables
-                non_skip_vars = [
-                    var.name
-                    for var in self.variable_codelist.values()
-                    if not var.skip_region_aggregation
-                ]
+                # Single constituent common regions are a special rename case
+                # (technically aggregated, so aggregation-skipped variables are excluded)
                 if common_region.is_single_constituent_region:
+                    non_skip_vars = [
```
danielhuppmann: It could be useful to make this an attribute of the **VariableCodeList** such that it is computed when first called, something like

```python
@property
def region_aggregation_variables():
    if self._common_regions_variables is not None:
        return self._common_regions_variables
    self._common_regions_variables = []
    return self._common_regions_variables 
```

docs/api/nuts.rst
```diff
@@ -23,8 +23,78 @@ The full list of NUTS regions is accessible via the Eurostat website (`xlsx, 500
...
+
+.. code:: yaml
+
+  dimensions:
+    - region
+    - variable
+  definitions:
+    region:
+      nuts:
+        nuts-3: [ AT ]
```
phackstock: If my understanding of how the `NutsProcessor` works in conjunction with `nomenclature.yaml` is correct, this example is not ideal. By providing nuts-3, you aggregate to nuts-2, nuts-1 and finally country level. Since your `RegionCodeList` only contains nuts-3, you'll end up dumping the regions at the end of the nuts aggregation process (https://github.com/dc-almeida/nomenclature/blob/ebef5f3f183acdeb5c3477101901d2b34ad15e69/nomenclature/processor/nuts.py#L304-L305). You're running through the entire aggregation process just to dump the results again, with the exception of the final country level.
Maybe in the example here you could emphasize the fact that all the nuts levels that are desired to have in the final result have to be listed here.

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,307 @@
...
+    def apply(self, df: IamDataFrame):
+        processed_dfs: list[IamDataFrame] = []
+
+        for model in df.model:
+            model_df = df.filter(model=model)
+
+            # Skip unlisted models
+            if model not in self.models:
+                logger.info(
+                    f"Skipping NUTS region aggregation for model '{model}' (no region processing mapping)"
```
phackstock:
```suggestion
                    f"Skipping NUTS region aggregation for model '{model}' (no nuts aggregation mapping)"
```

nomenclature/processor/region.py
```diff
@@ -793,6 +750,127 @@ def revert(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
...
+    return aggregated_data
+
+
+def merge_with_preaggregated_data(
+    model_df: IamDataFrame,
+    aggregated_data: list[pd.Series],
+    target_regions: list[str],
+    variable_codelist: VariableCodeList,
+    rtol_difference: float = 0.01,
+    return_aggregation_difference: bool = False,
```
phackstock, line 816: Not part of this PR and my own fault after consulting git blame but this variable does not do what it claims. The aggregation difference is always returned, the only thing that it does is that if the variable is set to `False` it additionally issues a log message detailing how to get the difference, I'm opening an issue.

phackstock (changes requested):
Thanks for adding a lot of functionality @dc-almeida. The PR looks good to me in principle. There a number of comments below but most of them are minor changes to make the code easier to read and maintain.

nomenclature/processor/nuts.py
dc-almeida: I set it up this way since the processor will always need the definitions to run, but the models could be passed separately from the yaml file and still work.

docs/user_guide/config.rst
```diff
@@ -174,3 +176,42 @@ the filtering for definitions.
...
+
+  processors:
+    region-processor: true
+
+This is equivalent to calling:
+
+.. code:: python
+
+  from nomenclature.processor import RegionProcessor
+  processor = RegionProcessor.from_directory("mappings", dsd)
```
danielhuppmann, line 207: Should this also "run" the processor as part of `process()`?

docs/user_guide/config.rst
```diff
@@ -174,3 +176,42 @@ the filtering for definitions.
...
+^^^^^^^^^^^^^^
+
+Setting *processors.nuts* to a list of model names will automatically create a
+:class:`NutsProcessor` and apply NUTS hierarchical aggregation (NUTS3 → NUTS2 →
+NUTS1 → Country → EU27) for those models:
+
+.. code:: yaml
+
+  processors:
+    nuts: [ Model A, Model B ]
```
danielhuppmann: For consistency with the section above? 
```suggestion
    nuts-processor: [ Model A, Model B ]
```

nomenclature/processor/nuts.py
```diff
@@ -0,0 +1,357 @@
...
+        )
+
+    @property
+    def nuts_codelist(self):
+        return RegionCodeList(
+            name="NUTS",
+            mapping={
+                code.name: code
+                for code in self.region_codelist.mapping.values()
+                if re.search(r"NUTS \d regions \(2024 edition\)", code.hierarchy)
```
danielhuppmann: Maybe a simpler approach to filtering?
```suggestion
                if code.hierarchy.startswith("NUTS ")
```

tests/data/nuts_processing/dsd/definitions/variable/variables.yaml
```diff
@@ -0,0 +1,3 @@
+- Primary Energy:
+    definition: Total primary energy consumption
+    unit: EJ/yr
```
danielhuppmann:
```suggestion
    unit: EJ/yr

```

tests/data/nuts_processing/dsd/nomenclature.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - variable
+definitions:
+  region:
+    country: true
+    nuts:
+      nuts-1: [AT]
+      nuts-2: [AT]
+      nuts-3: [AT]
+processors:
+  nuts: [model_a]
```
danielhuppmann:
```suggestion
  nuts: [model_a]

```

docs/api/nuts.rst
```diff
@@ -23,8 +23,92 @@ The full list of NUTS regions is accessible via the Eurostat website (`xlsx, 500
...
+
+The data is aggregated for the applicable variables, creating the common region
+``Austria`` (AT) from its constituent NUTS subregions.
+The country-level regions must be defined in a region definition file or by setting
+*definitions.region.country* as *true* in the configuration file
+(see :ref:`adding-countries`).
+
+.. note::
+
+   Only NUTS regions explicitly listed under ``definitions.region.nuts`` are present
```
danielhuppmann:
```suggestion
   Only NUTS regions explicitly listed under ``definitions.region.nuts`` are added to
```

docs/api/nuts.rst
```diff
@@ -23,8 +23,92 @@ The full list of NUTS regions is accessible via the Eurostat website (`xlsx, 500
...
+   in the output. The :class:`NutsProcessor` always aggregates through all levels,
+   but intermediate levels are **dropped** from the result unless they are listed
+   in the configuration. In the example above, all three levels (NUTS1, NUTS2, NUTS3)
```
danielhuppmann:
```suggestion
   but intermediate levels are **dropped** from the result unless if they are not listed
   in the configuration. In the example above, all three levels (NUTS1, NUTS2, NUTS3)
```

danielhuppmann (approved):
Looks good to me in principle, a few suggestions inline.

docs/api/nuts.rst
dc-almeida: Isn't this the opposite behaviour? If all NUTS levels are listed in the config file, then all are present in the final output?

danielhuppmann: Right, I wanted to remove the word "unless" as well. I was confused by the wording "dropped unless" and wanted to clarify...

PR #573, opened 2026-03-17
Add validation for include filters in CodeList

### LINKED ISSUE #551: Ensure that items included via `nomenclature.yaml` exist
In `nomenclature.yaml` we can "include" regions (e.g by hierarchy) and mappings from other repos, usually common-definitions.

We should add a test to ensure that if there is an "include" statement for an import, such an item should actually exist.

This is to guard against typos in `nomenclature.yaml`.

dc-almeida:
Closes #551. Introduce validation for `include` filters used in _nomenclature.yaml_ for codelists and mappings to ensure that all specified filters correspond to existing codes.
For codelists, `include` filters by code attributes, for mappings it filters by model name.


danielhuppmann (commented):
I used this feature this morning when implementing https://github.com/iiasa/prisma-workflow/pull/18, very useful! But the error messages could be more precise, see some suggestion inline. And it's not clear to me how models are validated.

dc-almeida:
Revisited this PR after the whole "bug" that wasn't a bug in adding new regions/mappings to ELEVATE/NEWPATHWAYS and made new commits to clean up some things and address some pending suggestions.

nomenclature/config.py
```diff
@@ -51,6 +51,18 @@ class CodeListConfig(BaseModel):
     )
     model_config = ConfigDict(validate_by_name=True, validate_by_alias=True)
 
+    @model_validator(mode="before")
+    @classmethod
+    def check_no_filter_at_dimension_level(cls, v):
+        if isinstance(v, dict):
```
phackstock, line 59: I'm afraid my pydantic is a little rusty these days, what's the purpose of the `if` clause here?

nomenclature/codelist.py
```diff
@@ -511,9 +514,29 @@ def filter(self, **kwargs) -> "CodeList":
...
+                f"Importing {dimension}s from external repository '{repository}' failed",
+                errors,
+            )
+
     @staticmethod
     def filter_codes(
-        codes: list[Code], include: dict | None = None, exclude: dict | None = None
+        codes: list[Code],
+        include: dict | list[dict[str, Any]] | None = None,
+        exclude: list[dict[str, Any]] | None = None,
```
phackstock: Is there a specific reason for not allowing the `exclude` input also to be a single `dict`? Like so:

```suggestion
        exclude: dict | list[dict[str, Any]] | None = None,
```

plus then the required recasting to a list below

phackstock (approved):
Thanks for the good work @dc-almeida especially since this PR opened up some unexpected complexities, nicely done 👍 
Changes are good to be merged regardless of the two comments below.

nomenclature/config.py
dc-almeida: This validator is specifically to catch the case of misplacing the `include` and `exclude` filters in the YAML file (outside of `repository` instead of inside it) so it checks the dict keys. But since it's a before validator, I need to ensure it is specifically a dict (not None, etc) otherwise it will raise another error looking for keys in the wrong type.

danielhuppmann (approved):
Very useful, thank you!

phackstock:
@dc-almeida, since #584 is now fixed, could you rebase this PR onto main and once the tests run through go ahead with the merge? Then we can issue a new release today, thanks.

PR #582, opened 2026-04-29
Add validation for missing components in variable codelist

### LINKED ISSUE #134: Check that components variables exist
Similar to #132 & #133 following the merging of #131, per [a suggestion](https://github.com/IAMconsortium/nomenclature/pull/131#pullrequestreview-1027377375) by @phackstock: implement a validation that all variables listed in "components" must be defined in the variable codelist.

dc-almeida:
Closes #134
Checks variables listed as `components` are present in codelist.
Currently only implemented for variables where `check-aggregate: true`, as it would break other tests that use `common-definitions` (for example, the case where Level-1 Species have Level-2 Species components that follow the variable hierarchy but are not reported in the data).


dc-almeida:
> Currently only implemented for variables where `check-aggregate: true`, as it would break other tests that use `common-definitions` (for example, the case where Level-1 Species have Level-2 Species components that follow the variable hierarchy but are not reported in the data).

@danielhuppmann should this validation encompass `check-aggregate: false` or can I move it to final review?

danielhuppmann:
Yes, having the `check-aggregate`-restriction now is a good idea to not have any surprises, and then I will work through **common-definitions** and activate it step-by-step.

nomenclature/codelist.py
```diff
@@ -774,6 +775,34 @@ def check_weight_in_vars(cls, v):
...
+            components: list[str] = (
+                var.components
+                if isinstance(var.components, list)
+                else [c for group in var.components.values() for c in group]
+            )
+            missing.extend(
+                (var.name, comp, var.file) for comp in components if comp not in v
+            )
+        if missing:
+            raise MissingComponentError(
```
danielhuppmann: Wondering if this shouldn't be an **UnknownVariable** error? (This also applies to the **MissingWeightError**). "Missing" to me implies that it is missing from the data, which is different than not defined in the codelist.

danielhuppmann (approved):
Looks good to me, thanks, except for one suggestion to rename the errors.

nomenclature/codelist.py
dc-almeida: Would it be a simple UnknownVariable error, or a subclass of it, which would allow for a clearer error message?

danielhuppmann: Right, maybe "UnknownVariableComponent"? So that the message can state the component that is not defined and the variable in which it is referenced.

dc-almeida: The subclass approach wasn't very feasible and overcomplicated for the situation, so I did a simple renaming.

PR #583, opened 2026-05-06
Close backlog of issues

### LINKED ISSUE #115: Allow single value input for DataStructureDefinition
When creating a new `DataStructureDefinition` object it would be nice to specify a single dimension just as a string and not a list which only contains one string, i.e. instead of:

```python
dsd = DataStructureDefinition(..., dimensions=["region"])
```

it would be nice to do this:

```python
dsd = DataStructureDefinition(..., dimensions="region")
```

The same goes for `DataStructureDefinition().validate(df, dimensions=['region'])`.

### LINKED ISSUE #154: Refine cli error tests
As for right now the tests for the cli that are supposed to fail only check _that_ the command failed and not _how_.
As part of improving error reporting the testing should test for the error message.

### LINKED ISSUE #181: Validate "directional" data
In the openENTRANCE project, we introduced a `>` character to define "directional" variable, for example trade flows from Country A to Country B. The region code then has the structure: `Austria>Germany`, see [here](https://github.com/openENTRANCE/openentrance/tree/main/definitions/region#directional-timeseries-data) for more info.

It would be useful that the nomenclature validation also checks that for any "directional" code in the region list, there can only be two components (`Austria>Germany>France`) is invalid and both components must be valid regions.

See [here](https://github.com/openENTRANCE/openentrance/blob/d28a9e61d1e594b20da121e3495a833990520d46/workflow.py#L37) for the current prototype in the openENTRANCE workflow.

### LINKED ISSUE #182: Summary errors of trailing whitespaces
Currently, the validation against trailing whitespaces only shows the first occurrence of such an error (not all). It would be more user-friendly to show all errors as part of the pydantic-ValidationError...

### LINKED ISSUE #183: CLI help options too crowded
When using the CLI and consulting the `--help` option for `nomenclature validate-project`, I noticed that the output is currently a little crowded:

```console
nomenclature validate-project --help
Usage: nomenclature validate-project [OPTIONS] PATH

  Assert that `path` is a valid project nomenclature

  Parameters ---------- path : Path     Project directory to be validated
  definitions : str, optional     Name of the definitions folder, defaults to
  "definitions" mappings : str, optional     Name of the mappings folder,
  defaults to "mappings" (if this folder exists) dimensions : List[str],
  optional     Dimensions to be checked, defaults to all sub-folders of
  `definitions`

  Example ------- $ nomenclature validate-project .
  --definitions <def-folder> --mappings <map-folder>
  --dimensions "['<folder1>', '<folder2>', '<folder3>']"

  Note ---- This test includes three steps:

  1. Test that all yaml files in `definitions/` and `mappings/` can be
  correctly read    as yaml files. This is a formal check for yaml syntax
  only. 2. Test that all files in `definitions/` can be correctly parsed as a
  :class:`DataStructureDefinition` object comprised of individual codelists.
  3. Test that all model mappings in `mappings/` can be correctly parsed as a
  :class:`RegionProcessor` object. This includes a check that all regions
  mentioned    in a model mapping are defined in the region codelist.

Options:
  --definitions TEXT  Optional name for definitions folder
  --mappings TEXT     Optional name for mappings folder
  --dimensions TEXT   Optional list of dimensions
  --help              Show this message and exit.

```

It seems that click is mashing up multiple doc strings.
Should be easy enough to clear that up.

### LINKED ISSUE #301: Explicit CLI check does not work on Windows as part of GH action
As discovered in #300, trying to explicitly call `nomenclature` as a subprocess does not work out of the box in Windows.

It might be related to the fact that we are using poetry and as detailed here: https://github.com/marketplace/actions/install-poetry-action#running-on-windows there are some extra steps necessary to get poetry to play nice with Windows in the first place.

It seems though as if the issue is not with installing `nomenclature` as a cli option itself.

### LINKED ISSUE #333: Exclude defaults when writing to yaml
Currently we include default values when writing to yaml which can make for somewhat unwieldy files.
Instead of:

```
- Some Variable:
    description: Some basic variable
    unit:
    skip-region-aggregation: false
```

we should do:

```
- Some Variable:
    description: Some basic variable
    unit:
```

### LINKED ISSUE #346: Add option to sort CodeLists

### LINKED ISSUE #350: Document model mapping conversion from Excel to YAML

### LINKED ISSUE #498: Add a `fail_ok` option to the Processor class
For a number of projects we now have processors that are allowed to fail and we still want to otherwise continue with the processing. Currently this is implemented "by hand" on a per workflow basis. One additional attribute in the `Processor` class and some logic in the `process` function should make this a lot cleaner.

### LINKED ISSUE #499: Use quotes around unknown variables/regions etc... to emphasize trailing whitespace
Just had the issue that a user tried to upload variables with trailing white spaces, nomenclature produced the following:

```console
- Bubble Investments|Biomass       
- Bubble Investments|CCS coal      
- ...
```

with using quotes it can be made more obvious what's going wrong:

```console
- "Bubble Investments|Biomass       "
- "Bubble Investments|CCS coal      "
- ...
```

### LINKED ISSUE #517: Invalid "include" logic should raise an error
Specifying `nomenclature.yaml` like 

```yaml
  region:
    include:
        - hierarchy: common
```
simply ignores the filter, because it should be formatted as

```yaml
  region:
    repository:
      name: common-definitions
      include:
        - hierarchy: common
```

It would be useful to check against the invalid filtering when validating the yaml files.

### LINKED ISSUE #538: Remove special handling of countries without alpha_3 code
When initializing a **RegionCodeList** from a directory with `countries: True`, there is currently a special handling for Kosovo which doesn't have an assigned ISO3166-1 code - but we manually added that in **nomenclature.countries**, so the try-except can probably be removed.

 https://github.com/IAMconsortium/nomenclature/blob/c0c425ba2ccbc4d2e492c41bf6708a82fdc38e9f/nomenclature/codelist.py#L801

dc-almeida:
Closes #115, closes #538, closes #181, closes #182, closes #517, closes #154, closes #183, closes #301, closes #333, closes #346, closes #350, closes #498, closes #499

Apologies for the large PR but most commits were quite minute, so looking at them one by one is fairly straightforward. Thus, I decided to condense them all into a single branch to merge.
This PR addresses pending issues in the backlog, reformats some documentation, docstrings, and code comments for consistency.

### DIFF
--- a/docs/user_guide/model-registration.rst
+++ b/docs/user_guide/model-registration.rst
@@ -24,6 +24,36 @@ Please use the `Excel template`_ and send it to the project managers by email.
 
 .. _`Excel template`: https://raw.githubusercontent.com/IAMconsortium/nomenclature/main/templates/model-registration-template.xlsx
 
+Converting Excel template to YAML
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you have filled out the Excel model registration template and need to convert it
+to YAML format, you can use the :ref:`cli` command ``parse-model-registration``:
+
+.. code-block:: bash
+
+    nomenclature parse-model-registration registration.xlsx
+
+This command reads the Excel file and generates two sets of YAML files:
+
+- **Region definitions**: Created in ``definitions/region/`` (default location)
+- **Model mappings**: Created in ``mappings/`` (default location)
+
+You can specify custom output folders using the ``--definition-path`` and
+``--mappings-path`` options:
+
+.. code-block:: bash
+
+    nomenclature parse-model-registration registration.xlsx \
+        --definition-path ./custom/definitions/region \
+        --mappings-path ./custom/mappings
+
+The command automatically:
+
+- Parses the "Common-Region-Mapping" sheet to generate model mapping YAML files
+- Parses the "Region-Country-Mapping" sheet (if present) to generate region definitions with country lists
+- Handles R5 region conventions with automatic World region generation
+
 Option 2) Registration using a GitHub pull request
 --------------------------------------------------
 
--- a/docs/user_guide/model-registration.rst
+++ b/docs/user_guide/model-registration.rst
@@ -52,7 +52,7 @@ The command automatically:
 
 - Parses the "Common-Region-Mapping" sheet to generate model mapping YAML files
 - Parses the "Region-Country-Mapping" sheet (if present) to generate region definitions with country lists
-- Handles R5 region conventions with automatic World region generation
+- Adds an aggregation to "World" from R5 regions (if present)
 
 Option 2) Registration using a GitHub pull request
 --------------------------------------------------
--- a/docs/user_guide/region.rst
+++ b/docs/user_guide/region.rst
@@ -11,16 +11,18 @@ structure is required:
    - Hierarchy 1:
      - Region 1:
          description: Some short explanation
-         iso3_codes: [ABC, DEF, ...]
+         countries: [Country 1, Country 2, ...]
      - Region 2:
          description: Some short explanation
-         iso3_codes: GHI
+         iso3_codes: ABC
    - Hierarchy 2:
      - ...
 
-Regions can have attributes, for example a description or ISO3-codes. If the attribute
-`iso3_codes` is provided, the item(s) are validated against a list of valid codes taken
-from the `pycountry <https://github.com/flyingcircusio/pycountry>`_ package.
+Regions can have attributes, for example a description, a list of countries, or ISO3 codes.
+If the attributes `countries` and/or `iso3_codes` are provided, the item(s) are 
+validated against a list of valid ISO3 country names/codes taken from the
+`pycountry <https://github.com/flyingcircusio/pycountry>`_ package.
+More details on the list of countries can be found :ref:`here <countries>`.
 
 Common regions
 --------------
--- a/docs/user_guide/region.rst
+++ b/docs/user_guide/region.rst
@@ -20,7 +20,7 @@ structure is required:
 
 Regions can have attributes, for example a description, a list of countries, or ISO3 codes.
 If the attributes `countries` and/or `iso3_codes` are provided, the item(s) are 
-validated against a list of valid ISO3 country names/codes taken from the
+validated against a list of valid ISO3 country names/codes adapted from the
 `pycountry <https://github.com/flyingcircusio/pycountry>`_ package.
 More details on the list of countries can be found :ref:`here <countries>`.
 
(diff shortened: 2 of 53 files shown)


docs/user_guide/model-registration.rst
```diff
@@ -24,6 +24,36 @@ Please use the `Excel template`_ and send it to the project managers by email.
...
+
+    nomenclature parse-model-registration registration.xlsx \
+        --definition-path ./custom/definitions/region \
+        --mappings-path ./custom/mappings
+
+The command automatically:
+
+- Parses the "Common-Region-Mapping" sheet to generate model mapping YAML files
+- Parses the "Region-Country-Mapping" sheet (if present) to generate region definitions with country lists
+- Handles R5 region conventions with automatic World region generation
```
danielhuppmann:
```suggestion
- Adds an aggregation to "World" from R5 regions (if present)
```

docs/user_guide/region.rst
```diff
@@ -11,16 +11,18 @@ structure is required:
...
    - Hierarchy 2:
      - ...
 
-Regions can have attributes, for example a description or ISO3-codes. If the attribute
-`iso3_codes` is provided, the item(s) are validated against a list of valid codes taken
-from the `pycountry <https://github.com/flyingcircusio/pycountry>`_ package.
+Regions can have attributes, for example a description, a list of countries, or ISO3 codes.
+If the attributes `countries` and/or `iso3_codes` are provided, the item(s) are 
+validated against a list of valid ISO3 country names/codes taken from the
+`pycountry <https://github.com/flyingcircusio/pycountry>`_ package.
```
danielhuppmann, line 24: This is actually validated against nomenclature countries list, which deviates slightly from the original **pycountry** package.

nomenclature/processor/aggregator.py
```diff
@@ -131,15 +131,16 @@ def codes(self):
...
         codelist = getattr(dsd, self.dimension, None)
-        # no validation if codelist is not defined or filter-item is None
+        # No validation if codelist is not defined or filter-item is None
         if codelist is None:
             error = f"Dimension '{self.dimension}' not found in DataStructureDefinition"
         elif invalid := codelist.validate_items(self.codes):
             error = (
                 f"The following {self.dimension}s are not defined in the "
-                "DataStructureDefinition:\n - " + "\n - ".join(invalid)
+                "DataStructureDefinition:\n - "
```
danielhuppmann: Maybe streamline to the following?
```
f"The following {self.dimension}s are not defined:\n - "
```

nomenclature/processor/region.py
```diff
@@ -887,7 +887,8 @@ def _compare_and_merge(
 ) -> tuple[IamDataFrame, pd.DataFrame]:
     """Compare and merge original and aggregated results"""
 
-    # compare processed (aggregated) data and data provided at the common-region level
+    # Compare aggregated (processed) data and original (pre-aggregated) data
+    # provided at the common-region level
```
danielhuppmann:
```suggestion
    # Compare aggregated (processed) and original data at the common-region level
```

danielhuppmann (approved):
Looks great, just a few minor suggestions inline

PR #585, opened 2026-05-19
Fix failing external repo test

### LINKED ISSUE #584: Nightly tests currently (May 19 2026) broken
Something related to the cleanup of the external repositories. 

Example: https://github.com/IAMconsortium/nomenclature/actions/runs/26085647579

phackstock:
Fixes #584

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -21,10 +21,10 @@ jobs:
     name: py${{ matrix.python-version }}
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python-version }}
     #----------------------------------------------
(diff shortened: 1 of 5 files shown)


dc-almeida:
A note that `rmtree`'s `onexc` is 3.12+ only, so that would mean no 3.11 support

phackstock:
Hence why I added this piece of code:
```python
            if sys.version_info[1] >= 12:
                shutil.rmtree(repository.local_path, onexc=remove_readonly)
            else:
                shutil.rmtree(repository.local_path, onerror=remove_readonly)
```

danielhuppmann (approved):
Not sure what magic is happening here, but happy that the tests pass again...

phackstock:
> Not sure what magic is happening here, but happy that the tests pass again...

The reason why the tests failed in the first place was due to errors when deleting the external repositories after a test. The way that I ended up fixing it is by not fetching them in the first place since it was not actually needed for any of the tests and in one case leaving the cleanup to pytest. 
So in short, I'm also not 100% sure what exactly the issue was and I solved it by circumventing the problem.
As a little bonus, the tests should now run a little bit faster since we're now avoiding cloning and deleting external repos a bunch of times.

PR #588, opened 2026-06-11
Update to pyam-iamc v3.4.0

phackstock:
This is more or less a pro forma PR just to explicitly check the compatibility of the latest pyam release, v3.4.0 with the current latest version of nomenclature.


danielhuppmann (approved):
Thanks for checking

PR #590, opened 2026-06-12
Update minimum pysquirrel to 1.3

phackstock:
This is required for the iDesignRes project (iiasa/idesignres-workflow#33)


dc-almeida (approved):
The test assertion difference is due to pysquirrel now excluding Extra-Regio codes by default, thus excluding one of the regions expected in the test. Good to go once that test is updated

phackstock:
Thanks for the quick review and clarification @dc-almeida. Just for my understanding what are "Extra-Regio codes"?

dc-almeida:
It's the NUTS terminology for placeholder regions, like ATZ, ATZZ, ATZZZ.

phackstock:
Ah I see, I've just updated the test accordingly by reducing the expected number of regions. That should be fine @dc-almeida, correct?

PR #592, opened 2026-06-15
Add `CountryProcessor`

### LINKED ISSUE #587: Add `CountryProcessor`
As suggested [here](https://github.com/IAMconsortium/common-definitions/pull/410#issuecomment-4595822378). Aggregation from country level regions to R5/R9/R10

dc-almeida:
Closes #587. Adds a `CountryProcessor` to aggregate country-level data into regional levels (R5, R9, R10).
Currently implemented with importing common-definitions as a dependency.


docs/api/countryprocessor.rst
```diff
@@ -0,0 +1,44 @@
...
+   definitions:
+     region:
+       country: true
+       repository:
+         name: common-definitions
+         include:
+           - hierarchy: [R5, R9, R10]
+
+   processors:
+     countries-processor: [Model A, Model B]
```
danielhuppmann: It's a bit confusing that the class is called **CountryProcessor** but the attribute is `countries-processor`.

nomenclature/processor/countries.py
```diff
@@ -0,0 +1,277 @@
...
+        models: list[str] | None = None,
+    ):
+        """Instantiate from a :class:`DataStructureDefinition`.
+
+        Parameters
+        ----------
+        dsd : DataStructureDefinition
+            Project data structure definition.
+        models : list[str], optional
+            Models to apply country aggregation to. Defaults to the list configured
```
danielhuppmann:
```suggestion
            Models for which to apply country aggregation. Defaults to the list configured
```

nomenclature/processor/countries.py
```diff
@@ -0,0 +1,277 @@
...
+                    # Convert ISO3 codes to country names
+                    for iso3 in code.iso3_codes:
+                        country = countries.get(alpha_3=iso3)
+                        if country:
+                            constituent_countries.append(country.name)
+                        else:
+                            logger.warning(
+                                f"ISO3 code '{iso3}' in region '{code.name}' "
+                                "not found in countries database"
+                            )
```
danielhuppmann: I suggest to simplify the code so that the R5/R9/R10 regions have to have the countries-attribute.

nomenclature/processor/countries.py
```diff
@@ -0,0 +1,277 @@
...
+        UnknownRegionError
+            If the result contains regions not defined in the region codelist.
+        """
+        processed_dfs: list[IamDataFrame] = []
+
+        for model in df.model:
+            model_df = df.filter(model=model)
+
+            # Skip unlisted models
+            if model not in self.models:
```
danielhuppmann: To avoid very verbose log-output (for all models that have custom mappings), I would skip this logging.

tests/test_country_aggregation.py
```diff
@@ -0,0 +1,193 @@
...
+
+        china_r10 = result.filter(region="China+ (R10)")
+        assert len(china_r10) > 0
+        assert china_r10.filter(year=2005)["value"].values[0] == 10.0
+        assert china_r10.filter(year=2010)["value"].values[0] == 12.0
+
+        pacific_oecd_r10 = result.filter(region="Pacific OECD (R10)")
+        assert len(pacific_oecd_r10) > 0
+        assert pacific_oecd_r10.filter(year=2005)["value"].values[0] == 3.0
+        assert pacific_oecd_r10.filter(year=2010)["value"].values[0] == 4.0
```
danielhuppmann: I think that this assertion can be simplified by defining the expected data and use `pyam.testing.assert_frame_equal()`.

danielhuppmann (commented):
Looks nice, a first round of suggestions and comments below.

danielhuppmann:
Is this ready for review yet?

dc-almeida:
Yes, I'll open it for final review. I left it as a draft because I was going to look into the minimum countries for aggregation question, but meanwhile looked at other tasks so best to close this one and tackle that a bit later.

docs/api/countryprocessor.rst
```diff
@@ -0,0 +1,44 @@
...
+       url: https://github.com/IAMconsortium/common-definitions.git/
+
+   definitions:
+     region:
+       country: true
+       repository:
+         name: common-definitions
+         include:
+           - hierarchy: [R5, R9, R10]
+
```
danielhuppmann: I've so far always done the structure "first the repo, then the other args"
```suggestion
       repository:
         name: common-definitions
         include:
           - hierarchy: [R5, R9, R10]
       country: true
```

docs/api/countryprocessor.rst
```diff
@@ -0,0 +1,44 @@
...
+   processors:
+     country-processor: [Model A, Model B]
+
+In this setup, all country codes are added to the region codelist, the R5/R9/R10
```
danielhuppmann: Country codes (alpha3) or names?

nomenclature/processor/countries.py
```diff
@@ -0,0 +1,264 @@
...
+from nomenclature.exceptions import UnknownRegionError
+from nomenclature.countries import countries
+
+logger = logging.getLogger(__name__)
+
+here = Path(__file__).parent.absolute()
+
+
+class CountryProcessor(Processor):
+    """Country to regional aggregation (R5/R9/R10) for scenario processing.
```
danielhuppmann:
```suggestion
    """Aggregation from country-level results to R5/R9/R10 regions
```

nomenclature/processor/countries.py
```diff
@@ -0,0 +1,264 @@
...
+            raise ValueError("No models configured for Country processor")
+        return cls(
+            variable_codelist=dsd.variable,
+            region_codelist=dsd.region,
+            models=models,
+        )
+
+    @property
+    def regional_aggregates_by_level(self) -> dict[str, dict[str, list[str]]]:
+        """Return regional aggregates grouped by hierarchy level."""
```
danielhuppmann: For consistency...
```suggestion
        """Return regional aggregates grouped by hierarchy level"""
```

nomenclature/processor/countries.py
```diff
@@ -0,0 +1,264 @@
...
+        aggregates_by_level: dict[str, dict[str, list[str]]] = {}
+        for code in self.region_codelist.mapping.values():
+            # Check if this is a regional aggregate (R5/R9/R10)
+            hierarchy = code.hierarchy or ""
+            level = next(
```
danielhuppmann: Is the iterator and the default `None` (and then checking if it is `None` in the next line really necessary?

nomenclature/processor/countries.py
```diff
@@ -0,0 +1,264 @@
...
+            # Skip unlisted models
+            if model not in self.models:
+                processed_dfs.append(model_df)
+            else:
+                logger.info(f"Applying country processing for model '{model}'")
+                processed_dfs.append(self._apply_country_processing(model_df)[0])
+
+        res = pyam.concat(processed_dfs)
+        if not_defined_regions := self.region_codelist.validate_items(res.region):
+            raise UnknownRegionError(not_defined_regions)
```
danielhuppmann: This should not happen, right? Because we only aggregate to R5, R9 and R10...

danielhuppmann (commented):
A few comments inline, and one general question about the architecture - wouldn't it make sense to have a constructor that builds a **RegionProcessor** class and then applies that class?

Maybe add a **RegionProcessor.from_codelist(hierarchy)** constructor?

nomenclature/processor/countries.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @property
+    def regional_aggregates_by_level(self) -> dict[str, dict[str, list[str]]]:
+        """Return regional aggregates grouped by hierarchy level."""
+        aggregates_by_level: dict[str, dict[str, list[str]]] = {}
+        for code in self.region_codelist.mapping.values():
+            # Check if this is a regional aggregate (R5/R9/R10)
+            hierarchy = code.hierarchy or ""
+            level = next(
+                (pattern for pattern in ["R5", "R9", "R10"] if pattern in hierarchy),
```
phackstock: I'd suggest to create a function that allow you to filter a RegionCodeList for a hierarchy.
Something long the lines of:
```python
self.region_codelist.by_hierachy("R5") -> list[RegionCode]
```
that would make this part much easier to read and understand

phackstock:
> A few comments inline, and one general question about the architecture - wouldn't it make sense to have a constructor that builds a RegionProcessor class and then applies that class?
>
> Maybe add a RegionProcessor.from_codelist(hierarchy) constructor?

I'd agree with that approach. The current implementation produces quite a lot of code duplication which increases the chance of failure and makes future maintenance harder. 
I'd suggest to either go with something along the lines of `RegionProcessor.create_country_processor_from_codelist` or making a `CountryProcessor` class that inherits from `RegionProcessor` to get all of the common functionality. As part of this, I'd suggest to directly create `RegionAggregationMapping`s for the models that are specified in the config.

nomenclature/processor/countries.py
```diff
@@ -0,0 +1,264 @@
...
+            if level:
+                # Get constituent countries (except for "Other" regions)
+                if not code.countries and not code.name.startswith("Other"):
+                    raise ValueError(
+                        f"List of constituent countries for region '{code.name}' "
+                        "not found in codelist."
+                    )
+                constituent_countries = code.countries
+
+                if constituent_countries:
```
phackstock: As outlined in my general comment in the PR, I think this would benefit from a re-write to use the existing functionality from `RegionProcessor`.

nomenclature/core.py
```diff
@@ -63,7 +63,7 @@ def process(
 
     # Auto-instantiate processors declared in nomenclature.yaml under 'processors'
     # Raise error if both explicit and config-based processors exist.
-    if getattr(dsd.config.processor, "region_processor", False):
+    if getattr(dsd.config.processor, "region", False):
```
phackstock: You don't need the safeguard here, since the config class is written to provide defaults of all fields, you can directly check:
```suggestion
    if dsd.config.processor.region:
```

as a bonus, this is more pleasant to read.

nomenclature/core.py
phackstock, line 1: I think this file is missing the actual usage of the country processor. You're checking for a nuts and a region processor but from what I read setting `config.processor.country = True` does not do anything currently.

phackstock (changes requested):
In terms for functionality, I'd say this PR is where we want it to be with the exception of checking a minimum set of countries. 
Implementation wise, I want to echo what @danielhuppmann said earlier in the discussion in this PR, `CountryProcessor` should either be implemented as a subclass of `RegionProcessor` or generated through a dedicated constructor as part of the `RegionProcessor` class. The RegionCodeList and model information should be used to generate model mappings on-the-fly that are then applied to incoming data.

nomenclature/processor/countries.py
dc-almeida: Initially implemented this, but realised the `filter` codelist method and the `hierarchy` region codelist property suffice

dc-almeida:
Applied review suggestions and:
- added a `RegionProcessor.from_country_codelist` factory method that creates a processor/aggregator for a list of hierarchies with `country` attributes
- added a `create_country_processor` function as a specific use case of the method above, for R5/R9/R10 regions
- initially added `by_hierarchy` RegionCodeList method but prioritised use of already existing `filter` CodeList method
- updated docs to have the country processor guide in `RegionProcessor`

docs/api/regionprocessor.rst
```diff
@@ -1,7 +1,60 @@
-.. currentmodule:: nomenclature
+.. currentmodule:: nomenclature.processor
 
 **RegionProcessor**
 ===================
 
 .. autoclass:: RegionProcessor
    :members: from_directory, validate_with_definition, apply, check_region_aggregation, revert, get_common_region_country_mapping, get_native_region_country_mapping
+
+Country to region processor
```
danielhuppmann:
```suggestion
Country-to-region processor
```

docs/api/regionprocessor.rst
```diff
@@ -1,7 +1,60 @@
...
+       country: true
+
+   processors:
+     country: [Model A, Model B]
+
+In this setup, all country names are added to the region codelist, the R5/R9/R10
+region definitions are imported from ``common-definitions``, and the country
+processor is applied to the listed models.
+
+Creating a country processor
```
danielhuppmann:
```suggestion
Creating a country-to-region processor
```

nomenclature/processor/region.py
```diff
@@ -619,6 +621,100 @@ def from_directory(cls, path: DirectoryPath, dsd: DataStructureDefinition):
...
+            NativeRegion(name=country) for country in all_countries_in_codelist
+        ]
+
+        # Create common regions for aggregation
+        common_regions = [
+            CommonRegion(name=region_name, constituent_regions=constituent_countries)
+            for region_name, constituent_countries in regional_aggregates.items()
+        ]
+
+        # Create the same mapping for all models
```
danielhuppmann:
```suggestion
        # Create a mapping for all models
```

nomenclature/processor/country.py
```diff
@@ -0,0 +1,46 @@
...
+
+from nomenclature.definition import DataStructureDefinition
+from nomenclature.processor.region import RegionProcessor
+
+logger = logging.getLogger(__name__)
+
+here = Path(__file__).parent.absolute()
+
+
+def create_country_processor(
```
danielhuppmann: I think it is more confusing than convenient to have two similarly named functions.

nomenclature/processor/region.py
```diff
@@ -619,6 +621,100 @@ def from_directory(cls, path: DirectoryPath, dsd: DataStructureDefinition):
             variable_codelist=dsd.variable,
         )
 
+    @classmethod
+    @validate_call(config={"arbitrary_types_allowed": True})
+    def from_country_codelist(
```
danielhuppmann: I know that @phackstock suggested this approach to have a `from_country_codelist()` constructor, but I think it would be more elegant to have a `CountryProcessor(Regionprocessor)` subclass with a `from_codelist()` constructor (so that this is consistent with the **DataValidator**).

This would also keep the code for teh **RegionProcessor* shorter.

danielhuppmann (commented):
Thanks, tests and implementation look good, but as detailed below, I would go for the proper subclass-strategy instead of a utility function.

tests/test_country_aggregation.py
```diff
@@ -0,0 +1,162 @@
...
+    dsd = DataStructureDefinition(COUNTRY_TEST_DIR / "all" / "definitions")
+    try:
+        # Load DSD and apply RegionProcessor
+        processor = CountryProcessor.from_codelist(dsd=dsd, models=["model_a"])
+
+        result = processor.apply(test_df)
+
+        # Check that aggregated regions are present
+        print(expected_values.keys())
+        print(set(result.region))
```
danielhuppmann:
```suggestion
```

tests/test_country_aggregation.py
```diff
@@ -0,0 +1,162 @@
...
+        assert set().issubset(set(result.region))
```
danielhuppmann: What does this assertion do? Seems like it checks that an empty set is a subset of the region-result...

danielhuppmann (approved):
Two minor comments, good from my side to merge after a review by @phackstock

nomenclature/processor/country.py
```diff
@@ -0,0 +1,121 @@
...
+            A CountryProcessor instance with generated mappings.
+
+        Raises
+        ------
+            If regions with the specified hierarchies lack constituent country information.
+        """
+        models = models or dsd.config.processor.country
+        if not models:
+            raise ValueError("No models configured for country processor")
+        skip_patterns = ["Other (R*)"] + (skip_patterns or [])
```
phackstock, line 64: This looks like it's doubled with the default value for `skip_patterns`

nomenclature/processor/country.py
```diff
@@ -0,0 +1,121 @@
...
+        available_hierarchies = set(hierarchies) & set(dsd.region.hierarchy)
```
phackstock: Suggest to make `hierarchies` a set from the get go. Then you don't need to cast to set.

```suggestion
        available_hierarchies = hierarchies & set(dsd.region.hierarchy)
```

nomenclature/processor/country.py
```diff
@@ -0,0 +1,121 @@
...
+    for hierarchies from the region codelist. Defaults to R5/R9/R10 regions.
+    """
+
+    @classmethod
+    @validate_call(config={"arbitrary_types_allowed": True})
+    def from_codelist(
+        cls,
+        dsd: DataStructureDefinition,
+        models: list[str],
+        hierarchies: list[str] = ["R5", "R9", "R10"],
```
phackstock: Suggest to make this a set directly, then you don't need to recast later.
```suggestion
        hierarchies: set[str] = {"R5", "R9", "R10"},
```

nomenclature/processor/country.py
```diff
@@ -0,0 +1,121 @@
...
+        models = models or dsd.config.processor.country
+        if not models:
+            raise ValueError("No models configured for country processor")
+        skip_patterns = ["Other (R*)"] + (skip_patterns or [])
+        available_hierarchies = set(hierarchies) & set(dsd.region.hierarchy)
+
+        # Extract regional aggregates from codelist for given hierarchies
+        regional_aggregates = {}
+        for hierarchy in available_hierarchies:
+            for code in dsd.region.filter(hierarchy=hierarchy).items():
```
phackstock: Suggest to use `.values()` here for easier readability. 
```suggestion
            for regioncode in dsd.region.filter(hierarchy=hierarchy).values():
```
This way in the following, you use `regioncode.name` and `regioncode.countries`. I think that easier to read than working with the key, value tuple.

nomenclature/processor/country.py
```diff
@@ -0,0 +1,121 @@
...
+                    regional_aggregates[code[0]] = code[1].countries
+
+        if not regional_aggregates:
+            logger.warning(
+                f"No regional aggregates for hierarchies {hierarchies} found in "
+                "region codelist."
+            )
+
+        # Get all countries from the codelist
+        country_names = set(countries.names)
```
phackstock, line 91: Not technically part of this PR but I'd suggest to make `countries.names` a set since country names in this collection should be unique anyway.

nomenclature/processor/country.py
```diff
@@ -0,0 +1,121 @@
...
+        all_countries_in_codelist = [
```
phackstock: I think you can directly use the `CodeList.filter` function here and then use `RegionCode.name` in line 100 below.
```suggestion
        all_countries_in_codelist = dsd.region.filter(name=all_countries_in_codelist)
```

nomenclature/processor/country.py
phackstock, line 1: Thanks a lot for implementing this change @dc-almeida, this is now much cleaner and much easier to read, nice job 👍

phackstock (approved):
Some minor cleanup comments from my side but good to be merged either way.

PR #595, opened 2026-06-19
Add subannual validation to TimeDomainConfig and corresponding tests

### LINKED ISSUE #594: Raise error for "subannual" column unless explicitly allowed
While doing some testing for another project using idesignres as a starting point, I noticed that using `datetime` format raises an exception
> nomenclature.exceptions.TimeDomainError: Invalid time domain - `mixed` found, but not allowed.
if it is not explicitly configured in `nomenclature.yaml`.

But... Using data with `year` plus `subannual` passes validation if no "subannual" codelist is defined. This could lead to a lot of headaches if users accidentally or otherwise submit subannual data to projects where that isn't expected.

So I suggest to raise a similar error when the subannual column exists but is not explicitly configured in the project nomenclature.yaml.

dc-almeida:
Closes #594

### DIFF
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -328,8 +328,21 @@ class TimeDomainConfig(BaseModel):
                 "The following datetime values are invalid:", errors
             )
 
-    def validate_datetime(self, df: IamDataFrame) -> None:
-        """Validate datetime coordinates against allowed format and/or timezone."""
+    def validate_datetime(self, df: IamDataFrame, dimensions: list[str] | None = None) -> None:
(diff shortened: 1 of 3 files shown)


dc-almeida:
Could opt to rephrase error message to refer to "dimension" instead of time domain for precision's sake

danielhuppmann (approved):
Nice, works as expected in my earlier test case.

I'd keep the error message as is for now, the subannual time-domain should probably be better reflected on the pyam side (like extending the `time_domain` attribute).

PR #598, opened 2026-07-03
Add long_description

phackstock:
As mentioned in iiasa/scse-update#46.

### DIFF
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -313,7 +313,7 @@ class VariableCode(Code):
         """Return a long description of the variable code, including validation args."""
         if self.description is None:
             return self.description
-        return self.description + f"; [{self.unit}]"
+        return self.description + f" [{self.unit}]"
 
 
 class RegionCode(Code):
@@ -412,14 +412,12 @@ class RegionCode(Code):
     def long_description(self) -> str | None:
         """Return a long description of the region code, including countries and ISO3 codes."""
         if self.description is None:
-            return self.description
+            return
         description = self.description
         if self.countries:
             description += f"; Countries: {', '.join(self.countries)}"
-        if self.iso3_codes and isinstance(self.iso3_codes, list):
-            description += f"; ISO3 codes: {', '.join(to_list(self.iso3_codes))}"
-        if self.iso3_codes and isinstance(self.iso3_codes, str):
-            description += f"; ISO3 code: {self.iso3_codes}"
+        if self.iso3_codes:
+            description += f" [{', '.join(to_list(self.iso3_codes))}]"
         return description
 
 
--- a/nomenclature/code.py
+++ b/nomenclature/code.py
@@ -18,7 +18,6 @@ from typing_extensions import Self
 
 from nomenclature.countries import countries
 
-
 # This must be kept in sync with the subtypes of `DataValidationCriteria`
 VALIDATION_ARGS = [
     "upper_bound",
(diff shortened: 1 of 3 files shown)


nomenclature/code.py
```diff
@@ -304,6 +308,13 @@ def validation_args(self) -> dict:
             if key in VALIDATION_ARGS and value is not None
         }
 
+    @property
+    def long_description(self) -> str | None:
+        """Return a long description of the variable code, including validation args."""
+        if self.description is None:
+            return self.description
+        return self.description + f"; [{self.unit}]"
```
danielhuppmann: Too much visual separation
```suggestion
        return self.description + f" [{self.unit}]"
```

nomenclature/code.py
```diff
@@ -397,6 +408,20 @@ def prefix(self) -> str:
     def has_prefix(self) -> bool:
         return "|" in self.name
 
+    @property
+    def long_description(self) -> str | None:
+        """Return a long description of the region code, including countries and ISO3 codes."""
+        if self.description is None:
+            return self.description
```
danielhuppmann: More obvious...
```suggestion
            return
```

tests/test_code.py
```diff
@@ -199,3 +199,35 @@ def test_code_with_multi_key_dict_raises():
...
+
+def test_code_long_description():
+    code = Code(name="code1", description="A code")
+    assert code.long_description == "A code"
+    assert code.long_description == code.description
+
+
+def test_variablecode_long_description():
+    variablecode = VariableCode(name="var1", description="A variable", unit="unit1")
+    assert variablecode.long_description == "A variable; [unit1]"
```
danielhuppmann:
```suggestion
    assert variablecode.long_description == "A variable [unit1]"
```

nomenclature/code.py
```diff
@@ -397,6 +408,20 @@ def prefix(self) -> str:
...
+        description = self.description
+        if self.countries:
+            description += f"; Countries: {', '.join(self.countries)}"
+        if self.iso3_codes and isinstance(self.iso3_codes, list):
+            description += f"; ISO3 codes: {', '.join(to_list(self.iso3_codes))}"
+        if self.iso3_codes and isinstance(self.iso3_codes, str):
+            description += f"; ISO3 code: {self.iso3_codes}"
```
danielhuppmann: The term "ISO3 code" isn't really correct, it should be ISO3166-1 alpha-3 - but that is too long...

So I suggest to simplify this
```suggestion
        if self.iso3_codes:
            description += f" [{', '.join(to_list(self.iso3_codes))}]"
```

nomenclature/code.py
```diff
@@ -397,6 +408,20 @@ def prefix(self) -> str:
...
         return "|" in self.name
 
+    @property
+    def long_description(self) -> str | None:
+        """Return a long description of the region code, including countries and ISO3 codes."""
+        if self.description is None:
+            return self.description
+        description = self.description
+        if self.countries:
+            description += f"; Countries: {', '.join(self.countries)}"
```
danielhuppmann, line 418: Some native-regions have only one country (or you implement a str-or-list check)
```suggestion
            description += f"; Countries: {', '.join(to_list(self.countries))}"
```

danielhuppmann (approved):
This looks good, just a few suggestions inline and one concern about regions created from pysquirrel: Those currently do not get a long-description, so either merge #599 or do a case-handling where no description but alpha-3 code returns a long description.

nomenclature/code.py
phackstock: We're already converting to a list using a pydantic validator: https://github.com/IAMconsortium/nomenclature/blob/2cb041718757350ae320e0ea1065467bd6031da0/nomenclature/code.py#L336-L351

PR #599, opened 2026-07-03
Add country name explicitly as the description

danielhuppmann:
This extends #598 because otherwise, regions that are created from pysquirrel won't get a long description with the alpha-3 code.


phackstock (approved):
Ah, good point, thanks

Issue #603, opened 2026-07-21
Refactor `RequiredDataValidator` to harmonise with `DataValidator`


dc-almeida:
While working on the refactor for the `MetaValidator`, I took a look at the `RequiredDataValidator` and thought it could be refactored to follow a similar structure to other validators.
In addition, there is a `# TODO consider merging with IamcDataFilter` which might suggest something in the same vein.
This isn't a priority at the moment, but decided to leave this comment here in case it might be worth pursuing later.

PR #604, opened 2026-07-23
Refactor `MetaValidator`

### LINKED ISSUE #597: Rework the **MetaValidator** class

dc-almeida:
Closes #597
This PR includes the following:
- [x] Refactors and expands `MetaValidator`
  - [x] `from_codelist`: uses meta codelists to validate against allowed values
  - [x] `from_file`: uses config files following a similar structure to `DataValidator`
    - Meta validation config files allow for a `meta` field that specifies the column(s) to validate, and `name` refers to the output column with validation results
  - [x] `apply` follows a similar implementation to `DataValidator`, but for meta indicators
    - the `value`/`values` field allows specifying a list of values or a single numerical value. Lists are hard checks, single numbers are assessed with `rtol`/`atol` (bounds).
- [x] Updates `MetaCode`, keeping a field alias for backwards compatibility
  - If backwards compatibility is not necessary, best to remove the alias, as it has a small caveat with Code `extra_attributes`
- [x] Refactors `DataValidator` by extracting generic `Validator` and `ValidatorCriteria` parent classes from which meta and data validators inherit
  - this avoid code duplication and makes #603 easier (if relevant)
- [x] Adds meta validation step in CLI command `validate-project`
- [x] Updates tests
- [x] Updates documentation


docs/api/metavalidator.rst
```diff
@@ -0,0 +1,56 @@
...
+Standard usage
+--------------
+
+.. code-block:: python
+
+  from nomenclature import MetaValidator
+
+  # ...setting directory/file paths and loading dataset
+
+  DataValidator.from_file(yaml_file_containing_meta_validation_criteria).apply(df)
```
danielhuppmann:
```suggestion
  MetaValidator.from_file(yaml_file_containing_meta_validation_criteria).apply(df)
```

nomenclature/processor/data.py
```diff
@@ -0,0 +1,194 @@
...
+    def validate_with_definition(self, dsd: DataStructureDefinition) -> None:
+        """Validate the criteria items against a :class:`DataStructureDefinition`.
+
+        Checks that all variables and regions referenced in the criteria
+        exist in the provided definition.
+
+        Parameters
+        ----------
+        dsd : DataStructureDefinition
+            Data structure definition to validate against.
```
danielhuppmann:
```suggestion
            Definition against which to check the items of the Validator
```

nomenclature/processor/meta.py
```diff
@@ -1,20 +1,140 @@
...
-    def __init__(self, path_to_meta_code_list_files: Path):
-        super().__init__(
-            meta_code_list=MetaCodeList.from_directory(
-                name="meta_code_list", path=path_to_meta_code_list_files
+    def validate_with_definition(self, dsd: DataStructureDefinition) -> None:
+        """Check meta indicators to validate against the DataStructureDefinition"""
+        codelist: MetaCodeList | None = getattr(dsd, "meta", None)
+        # No validation if codelist is not defined or filter-item is None
+        errors: list[NoTracebackException] = []
+        if codelist is None or getattr(self, "meta") is None:
```
danielhuppmann: Is this not a duplication of line 48?

nomenclature/processor/meta.py
```diff
@@ -1,20 +1,140 @@
...
+    @property
+    def criteria(self):
+        return self.model_dump(exclude_none=True, exclude_unset=True)
 
-    def __init__(self, path_to_meta_code_list_files: Path):
-        super().__init__(
-            meta_code_list=MetaCodeList.from_directory(
-                name="meta_code_list", path=path_to_meta_code_list_files
+    def validate_with_definition(self, dsd: DataStructureDefinition) -> None:
+        """Check meta indicators to validate against the DataStructureDefinition"""
```
danielhuppmann:
```suggestion
        """Check criteria items against the DataStructureDefinition"""
```

nomenclature/processor/meta.py
```diff
@@ -1,20 +1,140 @@
...
+
+class MetaValidationItem(ValidationItem, MetaFilter):
+    """Validation item for meta indicator validation"""
+
+    validation: list[MetaValidationValue | ValidationBounds | ValidationRange]
+
+    def apply(self, df: IamDataFrame, fail_list: list, output_list: list):
+        """Apply meta validation to IamDataFrame."""
+        error = False
+        per_item_df = df.meta.filter(self.meta, axis="columns")
```
danielhuppmann, line 90: Probably a question of personal preference, but I find the following easier to read.
```suggestion
        per_item_df = df.meta[self.meta]
```

nomenclature/processor/meta.py
```diff
@@ -1,20 +1,140 @@
...
+
+        # If name is given, set a meta indicator for the item being processed
+        if self.name is not None:
+            meta_index: pd.MultiIndex = per_item_df.index.copy()
```
danielhuppmann: This probably doesn't need to copy?
```suggestion
            meta_index: pd.MultiIndex = per_item_df.index
```

nomenclature/processor/meta.py
```diff
@@ -71,24 +258,85 @@ def apply(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
...
+    pd.DataFrame | None
+        A DataFrame of failing scenarios if any, otherwise None
+
+    Raises
+    ------
+    ValueError
+        *If a meta indicator in the 'df' is not listed in the .yaml
+        definition file
+    """
+    _df = df.copy()
```
danielhuppmann: Maybe move that to below the fastpass?

nomenclature/processor/meta.py
```diff
@@ -71,24 +258,85 @@ def apply(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
...
+    value = kwargs.get("value")
+    upper_bound = kwargs.get("upper_bound")
+    lower_bound = kwargs.get("lower_bound")
+    if df.empty:
+        logger.warning("No data matches filters, skipping validation.")
```
danielhuppmann: Would be more useful when the name of the column is also written here, but then the check would have to happen in the main function, I guess...
```suggestion
        logger.warning("Column does not exist in `meta`, skipping validation.")
```

danielhuppmann (commented):
Very nice PR - see a first round of comments inline.

nomenclature/processor/meta.py
dc-almeida: Well spotted, and the second check was nonsensical, an artifact from copying the code from IamcDataFilter

nomenclature/processor/meta.py
dc-almeida: I added the column names, the df passed is already filtered by the meta columns to validate, so just checking `df.columns` gives the desired list.
A question that arose from this: if a validation config file specifies more than one column but the df doesn't include one of them, should it raise or ignore the missing column(s)?

danielhuppmann: The **DataValidator** follows the logic that all available datapoints must satisfy the criteria, but it does not matter if no datapoint exists. So I would apply the same logic here in the case of multiple columns:
- if no value for any of the given columns exists for a run, the result is `None`
- if any value exists, the validation-outcome from the available values if assigned to that run


model | scenario | column 1 | column 2
------|----------|----------|----------
m_a    |  scen_a |                 |
m_a    |  scen_b |              1 |
m_a    |  scen_a |              1 |              3

where the validation-config is 

```yaml
- name: Validation item
  meta: [column 1, column 2]
  warning_level: high
  upper_bound: 2
```

then the outcome should be
- scen_a -> n/a
- scen_b -> "ok"
- scen_c -> "high"

If the user would want to check that both column 1 and column. 2 individually, they would have to specify separate criteria

nomenclature/utils.py
```diff
@@ -23,3 +23,7 @@ def handle_remove_readonly(func, path, excinfo):
         func(path)
     else:
         raise
+
+
+def single_input_to_list(v):
```
phackstock, line 28: I don't see this function being used anywhere. Please remove unless I've missed something

```suggestion
```

docs/api/metavalidator.rst
```diff
@@ -0,0 +1,56 @@
+.. _meta-validation:
+
+.. currentmodule:: nomenclature
+
+**MetaValidator**
+=================
+
+**Meta validation** checks if meta indicators follow allowed values and ranges.
```
phackstock: Meta validator sounds too much like validating validation to me rather than validating meta indicators:

```suggestion
**Meta indicator validation** checks if meta indicators follow allowed values and ranges.
```

docs/api/metavalidator.rst
```diff
@@ -0,0 +1,56 @@
...
+    validation:
+      - warning_level: high
+        upper_bound: 1490
+      - warning_level: medium
+        upper_bound: 1290
+  - meta: Project
+    values: [Project Name 1, Project Name 2]
+
+
+Each criteria item contains **meta indicator filter arguments** and **validation arguments**.
```
phackstock: I find this phrasing quite confusing. 

```suggestion
Each criteria item contains the **name of the meta column to be validated** and **validation arguments**.
```

docs/api/metavalidator.rst
```diff
@@ -0,0 +1,56 @@
...
+
+The meta indicator filter argument ``meta`` (also allowed as ``meta_columns_to_validate``)
```
phackstock:
```suggestion
The name of the column is specified in ``meta`` (also allowed as ``meta_columns_to_validate``).
```

docs/api/metavalidator.rst
```diff
@@ -0,0 +1,56 @@
...
+refers to the column(s) in the meta indicator table that will undergo validation.
```
phackstock:
```suggestion
```

nomenclature/processor/data.py
phackstock: I find the renaming of this file a bit confusing. I'd suggest to rename this file to `data_validator.py` then it's clear that this file contains, amongst other things, the `DataValidator` class.
The file that's currently called `validator.py`, I'd suggest to rename to `validation_criteria.py` and finally the file that contains the meta validator should be called `meta_validator.py`

nomenclature/processor/meta_validator.py
```diff
@@ -1,20 +1,140 @@
...
+                    + "\n"
+                )
+        return error, fail_list, output_list
+
+
+class MetaValidator(Validator):
+    """Meta indicator validation and processing class"""
+
+    criteria_items: list[MetaValidationItem]
+    file: Path | str
```
phackstock: Since you're inheriting from `Validator` you don't need this:
```suggestion
```

nomenclature/processor/meta_validator.py
```diff
@@ -1,20 +1,140 @@
...
+    output_path: Path | None = None
```
phackstock: Same as above, since you're inheriting from Validator you don't need this:
```suggestion
```

nomenclature/processor/meta.py
```diff
@@ -50,6 +170,73 @@ def _values_allowed(self, values, allowed_values, meta_indicator) -> bool:
...
+        Returns
+        -------
+        MetaValidator
+        """
+        with open(file, "r", encoding="utf-8") as f:
+            content = yaml.safe_load(f)
+        criteria_items = []
+        for item in content:
+            # Simple case where filter and criteria args are all given at top level
+            if "validation" not in item:
```
phackstock: I get the idea of wanting to offer the option to have a simple and flat structure but I think the price for this is too high in this case. I find this function hard to read and understand what's going on. I'd suggest to align the class more closely with the yaml structure. I'd suggest to only allow validation criteria to be defined under the `validation` label in yaml and rename the attribute in `MetaValidator` as well. That way you could remove almost this entire function.

nomenclature/processor/meta.py
```diff
@@ -50,6 +170,73 @@ def _values_allowed(self, values, allowed_values, meta_indicator) -> bool:
...
+            for criterion in criteria:
+                value = item.pop(criterion)
+                for criteria_item in item["validation"]:
+                    criteria_item[criterion] = value
+            criteria_items.append(item)
+
+        return cls(file=file, criteria_items=criteria_items, output_path=output_path)  # type: ignore
+
+    @classmethod
+    def from_codelist(
```
phackstock: I know we've said in IAMconsortium/nomenclature#597 that we want two constructors, directly from a yaml file and from a codelist. However, looking at this PR I think this adds unnecessary complexity. In the past, to my knowledge, we have **never** used the metacodelist. My suggestion therefore would be to eliminate one of the sources for the meta validator. Either only have it created from a metacodelist (that would then be the only job of a metacodelist) or have it specified in a yaml file in a `validate_meta` folder. I find keeping both options is unnecessary an complex.

nomenclature/processor/meta_validator.py
```diff
@@ -71,24 +258,88 @@ def apply(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
...
+    if lower_bound is not None:
+        failed_index.update(_df[_df < lower_bound].dropna(how="all").index)
+    if not failed_index:
+        return
+    _df = df.loc[sorted(failed_index)]
+
+    if not _df.empty:
+        msg = "{} of {} meta indicators do not satisfy the criteria"
+        logger.warning(msg.format(len(_df), len(df)))
+        return _df
```
phackstock: My mypy complains about a missing return statement here since all the return statements are part of if-clauses. I'd add a final `return None` outside and after the if here.

nomenclature/processor/meta_validator.py
```diff
@@ -71,24 +258,88 @@ def apply(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
...
+        fail_list: list[str] = []
+        output_list: list[pd.DataFrame] = []
 
-        for meta_indicator in df.meta.columns:
-            self._values_allowed(
-                list(set(df.meta[meta_indicator].values)),
-                self.meta_code_list.mapping[meta_indicator].allowed_values,
-                meta_indicator,
-            )
+        with adjust_log_level():
```
phackstock: I prefer to make this explicit because as it's imported here it's not obvious which logger is affected and at what level
```suggestion
        with adjust_log_level(logger="pyam", level="ERROR"):
```

nomenclature/processor/meta_validator.py
```diff
@@ -71,24 +258,88 @@ def apply(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
...
+            for item in self.criteria_items:
+                error, fail_list, output_list = item.apply(df, fail_list, output_list)
+                error_list.append(error)
+            if self.output_path:
+                pd.concat(output_list).to_excel(self.output_path, index=False)
+            fail_msg = f"(file {get_relative_path(self.file)}):\n"
```
phackstock: This should not be its own variable but just part of the string interpolation on line 276

nomenclature/processor/data.py
```diff
@@ -0,0 +1,194 @@
+import logging
```
danielhuppmann, line 1: Suggest to keep the file name as `data_validator.py` for clarity

nomenclature/processor/meta_validator.py
```diff
@@ -1,20 +1,140 @@
-from pathlib import Path
+import logging
```
danielhuppmann, line 1: Suggest to name `meta_validator.py` for clarity.

nomenclature/testing.py
```diff
@@ -151,3 +156,4 @@ def assert_valid_structure(
         path, dsd, RequiredDataValidator, "required_data", required_data
     )
     _check_processor_directory(path, dsd, DataValidator, "validate_data", validate_data)
+    _check_processor_directory(path, dsd, MetaValidator, "validate_meta", validate_meta)
```
danielhuppmann: Consider putting this in a loop like
```python
for _type, name, arg in [
    (DataValidator, "validate_data", validate_data),
    ....
]:
    _check_processor_directory(path, did, _type, name, arg)
```

danielhuppmann (commented):
Thanks @dc-almeida, very useful.

I think that the `MetaValidator.from_codelist()` doesn't actually work as expected, so I suggest to remove this and implement in a follow-up PR.

Can you implement these changes as a first step, and I will do another review tomorrow?

phackstock:
> I think that the MetaValidator.from_codelist() doesn't actually work as expected, so I suggest to remove this and implement in a follow-up PR.

I've started reviewing this PR but then got sidetracked and didn't finish the review yet but I arrived at the same conclusion. Being able to create a `MetaValidator` both from a metacodelist as well as a dedicated validator file is unnecessary and adds overhead and complexity that we don't need. I'd even go so far as to suggest to rip out `MetaCodeList` outright as it does not serve any purpose that I'm aware of. As a side benefit, this should clean up the `MetaValidator` constructor code nicely.

danielhuppmann:
@phackstock, not quite what I meant - I think that the feature to create a `MetaValidator` from a codelist and execute it as part of `DataStructureDefinition.validate()` will be quite useful - see what I did at [https://github.com/IAMconsortium/climate-definitions/pull/6](https://github.com/IAMconsortium/climate-definitions/pull/6/changes#diff-4aecb2c7d26ab3a93236b5a46c703a6579e5c28890d0cd723db97fe3d3e4c08d).

This will allow us to have the name of a meta-indicator together with a definition (to be used in the documentation-page or mouse-over in a Scenario Explorer app).

My point was just that this doesn't work as I expected yet in this PR, so my suggestion was to yank it out now and then discuss later how to proceed.

docs/api/metavalidator.rst
```diff
@@ -0,0 +1,56 @@
...
+Standard usage
+--------------
+
+.. code-block:: python
+
+  from nomenclature import MetaValidator
+
+  # ...setting directory/file paths and loading dataset
+
+  MetaValidator.from_file(yaml_file_containing_meta_validation_criteria).apply(df)
```
danielhuppmann, line 53: Maybe for better readability?
```suggestion
  MetaValidator.from_file("path/to/file.yaml").apply(df)
```

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,317 @@
...
+    """Meta indicator validation and processing class"""
+
+    criteria_items: list[MetaValidationItem]
+    file: Path | str
+    output_path: Path | None = None
+    exception_cls: type[NoTracebackException] = MetaValidationError
+
+    def _values_allowed(self, values, allowed_values, meta_indicator) -> bool:
+        """Checks if the values within a meta indicator column are
+        listed in model mapping
```
danielhuppmann, line 141: It's not clear what the "model mapping" refers to here?

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,317 @@
...
+class MetaValidator(Validator):
+    """Meta indicator validation and processing class"""
+
+    criteria_items: list[MetaValidationItem]
+    file: Path | str
+    output_path: Path | None = None
+    exception_cls: type[NoTracebackException] = MetaValidationError
+
+    def _values_allowed(self, values, allowed_values, meta_indicator) -> bool:
+        """Checks if the values within a meta indicator column are
```
danielhuppmann, line 140:
```suggestion
        """Checks if the values in a meta indicator column are
```

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,317 @@
...
+                    textwrap.indent(
+                        failed_validation.iloc[:, :-1].to_string(), prefix="  "
+                    )
+                    + "\n"
+                )
+        return error, fail_list, output_list
+
+
+class MetaValidator(Validator):
+    """Meta indicator validation and processing class"""
```
danielhuppmann, line 132: The docstrings are not consistent between
- meta indicator
- meta-indicator
- meta_indicator

I suggest to use the hyphen consistently

```suggestion
    """Meta-indicator validation and processing class"""
```

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,317 @@
...
+                errors,
+            )
+
+
+def repr_list(x):
+    return "'" + "', '".join(map(str, x)) + "'"
+
+
+def _validate_meta(df: pd.DataFrame, **kwargs) -> pd.DataFrame | None:
+    """Validate meta indicator values in IamDataFrame.
```
danielhuppmann, line 272: Not quite consistent whether to use a stop at the end of the first one of the docstring
```suggestion
    """Validate meta indicator values in IamDataFrame
```

tests/test_meta.py
```diff
@@ -1,37 +1,212 @@
...
 
-def test_MetaValidator(simple_df):
-    meta_validator = MetaValidator(MODULE_TEST_DATA_DIR / "definitions1" / "meta")
-    exp = simple_df.copy()
-    pyam.testing.assert_iamframe_equal(exp, meta_validator.apply(df=simple_df))
+def test_MetaValidator_from_file():
+    """
+    Test MetaValidator can be created from a YAML file and validation criteria
+    are set correctly.
+    """
```
danielhuppmann, line 23: Detailed descriptions are usually good, but here, it would improve readability to make them fit in one line (also for the dosctrings below).
```suggestion
    """Check that a MetaValidator can be created from a yaml file"""
```

tests/test_meta.py
```diff
@@ -1,37 +1,212 @@
...
+    )
+    assert meta_validator.criteria_items[0].validation[0].upper_bound == 1.0
+    assert (
+        meta_validator.criteria_items[0].validation[0].warning_level == WarningEnum.high
+    )
+    assert meta_validator.criteria_items[0].validation[1].upper_bound == 0.0
+    assert (
+        meta_validator.criteria_items[0].validation[1].warning_level
+        == WarningEnum.medium
+    )
```
danielhuppmann, line 35: Again a suggestion to improve readability by avoiding the very messy multi-line assertions...
```suggestion
    criteria_item_0 = meta_validator.criteria_items[0]
    assert criteria_item_0.validation[0].upper_bound == 1.0
    assert criteria_item_0.validation[0].warning_level == WarningEnum.high
    assert criteria_item_0.validation[1].upper_bound == 0.0
    assert criteria_item_0.validation[1].warning_level == WarningEnum.medium
```

nomenclature/processor/data_validator.py
```diff
@@ -322,45 +165,6 @@ def from_codelist(
             file="definitions", criteria_items=criteria_items, output_path=output_path
         )
 
-    def apply(self, df: IamDataFrame) -> IamDataFrame:
```
danielhuppmann, line 325: More a question of style, but I think it would be more helpful to keep the `apply()` implementation in the **DataValidator** class - because the **MetaValidator** has a custom `apply()` method anyway, right?

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,317 @@
...
+        Returns
+        -------
+        True : boolean
+            If all column elements are listed in model mapping
+
+        Raises
+        ------
+        ValueError
+            *If any of the values in the meta indicator column are not
+            listed in model mapping
```
danielhuppmann, line 161: Again not clear to me what the model mapping is here?

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,317 @@
...
+            criteria_items.append(item)
+
+        return cls(file=file, criteria_items=criteria_items, output_path=output_path)  # type: ignore
+
+    def apply(self, df: pyam.IamDataFrame) -> pyam.IamDataFrame:
+        """Apply meta indicator validation processing
+
+        Parameters
+        ----------
+        df (pyam.IamDataFrame)
```
danielhuppmann, line 217: This is not our usual formatting, right?

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,317 @@
...
+            Input data whose meta indicators will be validated
+
+        Returns
+        -------
+        df (pyam.IamDataFrame)
```
danielhuppmann, line 222:
```suggestion
        pyam.IamDataFrame
```

danielhuppmann (approved):
I have a few more comments about style to improve readability - overall, this looks good to me and will be helpful for the SCI evaluation.

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,317 @@
...
+                    + ", ".join(f"'{item}'" for item in invalid)
+                )
+            )
+            raise NoTracebackExceptionGroup(
+                f"Errors in {self.__class__.__name__}", errors
+            )
+
+
+class MetaValidationValue(ValidationValue):
+    value: float | list[Any] = Field(..., alias="values")
```
phackstock: Does this type hint make sense? Looking further down, I think it should be this:
```suggestion
    value: float | str | list[float|str] = Field(..., alias="values")
```

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,288 @@
...
+
+    @computed_field
+    @property
+    def lower_bound(self) -> float | None:
+        return self.value - self.tolerance if isinstance(self.value, float) else None
+
+    @property
+    def validation_args(self):
+        # In case of list of values, validation is a hard equality check
+        if isinstance(self.value, list):
```
phackstock, line 92: This is in disagreement with the type hint above. According to the type hint `value` can only be of type `float` but here you're checking if it's a list.

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,288 @@
...
+            return {"value": self.value}
+        # Else, return the bounds for tolerance check
+        return self.model_dump(
+            exclude_none=True,
+            exclude_unset=True,
+            exclude=["warning_level", "value", "rtol", "atol"],
```
phackstock: From what I gather, you only want to return the not None of upper and lower bound so I think the following is more readable:
```suggestion
            include=["upper_bound", "lower_bound"],
```

nomenclature/processor/meta_validator.py
dc-almeida: Thanks for these comments, this was an orphaned method from the original implementation that I had forgotten to remove!

phackstock:
> @phackstock, not quite what I meant - I think that the feature to create a MetaValidator from a codelist and execute it as part of DataStructureDefinition.validate() will be quite useful - see what I did at [https://github.com/IAMconsortium/climate-definitions/pull/6](https://github.com/IAMconsortium/climate-definitions/pull/6/changes#diff-4aecb2c7d26ab3a93236b5a46c703a6579e5c28890d0cd723db97fe3d3e4c08d).
This will allow us to have the name of a meta-indicator together with a definition (to be used in the documentation-page or mouse-over in a Scenario Explorer app).

Fair point, it still adds a bit of overhead in having to support two different constructor methods on the side of the `MetaValidator` but I see that the use cases are important enough so that we pay this price.

docs/api/metavalidator.rst
```diff
@@ -0,0 +1,56 @@
+.. _meta-validation:
+
+.. currentmodule:: nomenclature
+
+**MetaValidator**
+=================
+
+**Meta validation** checks if meta-indicators follow allowed values and ranges.
```
phackstock: I'd add the word "_indicator_" because the way that I read it right now, "meta validation" sound like we're validating the validation, which is not what we're doing
```suggestion
**Meta indicator validation** checks if meta-indicators follow allowed values and ranges.
```

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,285 @@
...
+
+        Returns
+        -------
+        MetaValidator
+        """
+        with open(file, "r", encoding="utf-8") as f:
+            content = yaml.safe_load(f)
+        criteria_items = []
+        for item in content:
+            # Simple case where filter and criteria args are all given at top level
```
phackstock: This option might make for slightly cleaner files but I'd suggest against it. The example that you're giving in the docs contains the "validation" keyword and for the sake not violating the KISS principle, I'd suggest to strictly enforce the format where validation criteria have to be given under the "validation" keyword in the yaml file.

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,285 @@
...
+                f"Error in MetaValidator (file {get_relative_path(self.file)})",
+                errors,
+            )
+
+
+def repr_list(x):
+    return "'" + "', '".join(map(str, x)) + "'"
+
+
+def _validate_meta(df: pd.DataFrame, **kwargs) -> pd.DataFrame | None:
```
phackstock: I wouldn't use kwargs here since the function expects three named inputs:

```suggestion
def _validate_meta(df: pd.DataFrame, value, upper_bound, lower_bound) -> pd.DataFrame | None:
```
(plus the correct type hints)
In my opinion the **kwargs pattern only fits when you then forward the **kwargs to some other function and don't want to explicitly have to replicate the interface of the second function.

`_validate_meta` should also be part of the `MetaValidationItem` class since it's used there and only there.

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+    @property
+    @abc.abstractmethod
+    def validation_args(self):
+        """Attributes used for validation."""
+        pass
+
+    @property
+    @abc.abstractmethod
+    def criteria(self):
+        """Attributes used for validation (as specified in the file)."""
```
phackstock: The way that I understand it, `criteria` are only ever used in the communication with the user since they mirror 1:1 what's in the validation file. This should be reflected in the docstring, currently it reads too similar to `validation_args`

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+        pass
+
+    def __str__(self):
+        return ", ".join([f"{key}: {value}" for key, value in self.criteria.items()])
```
phackstock, line 63: This should be (for all intents and purposes at least) equivalent:
```suggestion
        return str(self.criteria)
```

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+
+    @computed_field
+    @property
+    def lower_bound(self) -> float | None:
+        return self.value - self.tolerance if isinstance(self.value, float) else None
+
+    @property
+    def validation_args(self):
+        # In case of list of values, validation is a hard equality check
+        if isinstance(self.value, list):
```
phackstock, line 92: This is not consistent with the type hint where value has to be a float. Which one is correct? Does it have to be a float or can it be a list too?

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+            return {"value": self.value}
+        # Else, return the bounds for tolerance check
+        return self.model_dump(
+            exclude_none=True,
+            exclude_unset=True,
+            exclude=["warning_level", "value", "rtol", "atol"],
```
phackstock: I think the other way round is more readable:

```suggestion
            include=["upper_bound", "lower_bound"],
```

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+        )
+
+    @property
+    def criteria(self):
+        return self.model_dump(
+            exclude_none=True,
+            exclude_unset=True,
+            exclude=["warning_level", "lower_bound", "upper_bound"],
```
phackstock: Same suggestion as above, I think it's easier to understand what this function returns by listing what it returns rather than listing what it does not return:
```suggestion
            include=["value", "atol", "rtol"],
```

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+    @property
+    def lower_bound(self) -> float:
+        return self.range[0]
+
+    @property
+    def validation_args(self):
+        return self.model_dump(
+            exclude_none=True,
+            exclude_unset=True,
+            exclude=["warning_level", "range"],
```
phackstock:
```suggestion
            exclude=["upper_bound", "lower_bound"],
```

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+        )
+
+    @property
+    def criteria(self):
+        return self.model_dump(
+            exclude_none=True,
+            exclude_unset=True,
+            exclude=["warning_level", "lower_bound", "upper_bound"],
```
phackstock:
```suggestion
            include=["range"],
```

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+        return self
+
+    @property
+    def validation_args(self):
+        return self.criteria
+
+    @property
+    def criteria(self):
+        return self.model_dump(
+            exclude_none=True, exclude_unset=True, exclude=["warning_level"]
```
phackstock:
```suggestion
            exclude_none=True, exclude_unset=True, include=["upper_bound", "lower_bound"]
```

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+        ):
+            raise ValueError(
+                f"Validation criteria for {self.criteria} not sorted"
+                " in descending order of severity."
+            )
+        else:
+            return self
+
+    @property
+    def filter_args(self):
```
phackstock: If you exclude `name` and `validation` what is left of the class? From looking at it quickly, this function looks like it would just return an empty dictionary. EDIT: I saw that it is used in `MetaValidationItem` and `DataValidationItem` but there it only returns the filter arguments, as the function name suggests. I think this function can and should be removed from here and the corresponding `criteria` functions in `MetaFilter` and `IamcDataFilter` renamed to `filter_args`.

nomenclature/processor/validator.py
```diff
@@ -0,0 +1,269 @@
...
+            exclude_none=True, exclude_unset=True, exclude=["validation", "name"]
+        )
+
+    @abc.abstractmethod
+    def apply(self, df: IamDataFrame, fail_list: list, output_list: list):
+        """Apply validation to IamDataFrame."""
+        pass
+
+    def __str__(self):
+        return ", ".join([f"{key}: {value}" for key, value in self.filter_args.items()])
```
phackstock: I'm not sure if a `ValidationItem` is sufficiently represented by just its filter args. On an unrelated note, if you want to keep it this way, I'd suggest:

```suggestion
        return str(self.filter_args)
```

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,285 @@
...
+
+    @property
+    def criteria(self):
+        return self.model_dump(exclude_none=True, exclude_unset=True)
+
+    def validate_with_definition(self, dsd: DataStructureDefinition) -> None:
+        """Check criteria items against the DataStructureDefinition."""
+        codelist: MetaCodeList | None = getattr(dsd, "meta", None)
+        # No validation if codelist is not defined or filter-item is None
+        errors: list[NoTracebackException] = []
```
phackstock: Suggest to put this after checking for `None`
EDIT: After reading the rest of the class, I don't think this is needed at all:

```suggestion
```

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,285 @@
...
+        if codelist is None:
+            return
+        if invalid := codelist.validate_items(getattr(self, "meta")):
+            errors.append(
+                NoTracebackException(
+                    "The following meta-indicators are not defined in the "
+                    "DataStructureDefinition:\n   "
```
phackstock:
```suggestion
                    "MetaCodeList:\n   "
```

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,285 @@
...
+                    + ", ".join(f"'{item}'" for item in invalid)
+                )
+            )
+            raise NoTracebackExceptionGroup(
```
phackstock: I don't think that this needs the ExceptionGroup, since you don't have a loop, you'll only ever have a single `NoTracebackException` so not really a group. Instead you can raise the above error directly.

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,285 @@
...
+        Parameters
+        ----------
+        df : pyam.IamDataFrame
+            Input data whose meta-indicators will be validated
+
+        Returns
+        -------
+        pyam.IamDataFrame
+            If all meta-indicators and their values are listed in the
+            model mapping, the same df is returned.
```
phackstock: I don't fully understand this sentence and the last part is not true, if the validation specifies it, new meta columns with the result of the validation will be added.

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,285 @@
...
+
+        Raises
+        ------
+        ValueError
```
phackstock: I don't see a line of code that directly raises a `ValueError` I only see a `MetaValidationError`.

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,285 @@
...
+            If a meta-indicator in the 'df' is not listed in the .yaml
```
phackstock: If this explanation is supposed to fit to `MetaValidationError` I think it's false. The error is raised if the validation fails. The validation can fail because the meta indicator is missing but also because the values are wrong.

nomenclature/processor/meta_validator.py
```diff
@@ -0,0 +1,285 @@
...
+        Validation criteria
+
+    Returns
+    -------
+    pd.DataFrame | None
+        A DataFrame of failing scenarios if any, otherwise None
+
+    Raises
+    ------
+    ValueError
```
phackstock: This is wrong, this function does not raise an error at least not directly. Plus the definition is wrong, if the column does not exist the validation is skipped.

phackstock (changes requested):
Thanks for putting in the work for this rather large PR @dc-almeida. In terms of pure functionality, I think we're there, great work 👍
With regards to implementation and maintainability though, there's quite a few things to change in my opinion, comments in line below.

dc-almeida:
I am one commit away from re-requesting a review.
An example: I have a validation file with **3 criteria**. I have **2 meta-indicators**. After validation, the 2 meta-indicators fail, 1 of them fails for criteria 1, the other fails for criteria 2 and 3. What's the best message to summarise this information in the log warning?
Before, the message was simply "X of Y meta-indicators do not satisfy the criteria" and then it printed the flagged rows.

danielhuppmann:
For the data-validation, we show the failing rows for every criteria item and validation-level, right?

So I suggest something similar, like

```
Meta-indicator validation failed with error(s) (file: ...)
  Criteria: meta-column "Project", list: [Project 1, Project 2]
       model scenario               value warning_level
  0  model_a   scen_a  Not-A-Real-Project         error

```

PR #605, opened 2026-08-04
Separate run workflow for scse processing

phackstock:
This PR separates the "running a workflow" part out of the cli command so that it can be used by scse-processing to directly capture the `IamDataFrame` to save on an additional write and read operation.

### DIFF
--- a/nomenclature/cli.py
+++ b/nomenclature/cli.py
@@ -1,5 +1,3 @@
-import importlib.util
-import sys
 from pathlib import Path
 
 import pandas as pd
@@ -10,6 +8,7 @@ from typing_extensions import Annotated
 
 from nomenclature import __version__
 from nomenclature.codelist import CodeList, VariableCodeList
+from nomenclature.core import run_workflow
 from nomenclature.definition import SPECIAL_CODELIST, DataStructureDefinition
 from nomenclature.processor import RegionAggregationMapping, RegionProcessor
 from nomenclature.testing import assert_valid_structure, assert_valid_yaml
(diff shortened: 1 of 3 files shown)


nomenclature/cli.py
```diff
@@ -206,8 +206,26 @@ def list_missing_variables(
 # ---------------------------------------------------------
 # run-workflow
 # ---------------------------------------------------------
-@app.command()
 def run_workflow(
```
danielhuppmann: Would it make sense to move this to `core.py`? Better to import from the main package rather than from the cli module

phackstock: Yeah I was wondering where to put the function and agreed, it doesn't really feel at home in `cli.py`, `core.py` is for sure a more fitting place.

nomenclature/core.py
```diff
@@ -113,3 +116,23 @@ def process(
         )
 
     return df
+
+
+def run_workflow(
+    input_file: Path,
```
danielhuppmann: This could now also be an **IamDataFrame** for consistency with `process()`.
```suggestion
    df: IamDataFrame,
```

nomenclature/cli.py
```diff
@@ -231,16 +232,7 @@ def run_workflow(
...
-    spec = importlib.util.spec_from_file_location(module_name, workflow_file)
-    workflow = importlib.util.module_from_spec(spec)
-    sys.modules[module_name] = workflow
-    spec.loader.exec_module(workflow)
-
-    if not hasattr(workflow, workflow_function):
-        raise ValueError(f"{workflow} does not have a function `{workflow_function}`")
-
-    df: IamDataFrame = getattr(workflow, workflow_function)(IamDataFrame(input_file))
+    df = run_workflow(input_file, workflow_file, workflow_function)
```
danielhuppmann: Corresponding suggestion to my other comment in `core.py`.
```suggestion
    df = run_workflow(IamDataFrame(input_file), workflow_file, workflow_function)
```

danielhuppmann (approved):
One minor suggestion inline, then good to be merged.

nomenclature/core.py
phackstock: Ah yes, of course, great catch, thanks a lot. With this change we're now actually saving two read and one write operation. That should be a very nice performance improvement.

PR #606, opened 2026-08-04
Allow datetime-items without timezone

danielhuppmann:
While trying to fix a failed submission attempt by @lindnemi to the **ariadne-dev** instance, I become quite annoyed with the requirement that datetime-items *must* have a timezone even if no timezone is required in the instance config.

This PR removes the error when a datetime-dataset is validated where the datetime-items do not have a timezone. Instead, the datetime-validation only checks the timezone if that is specified in the project config.

### DIFF
--- a/nomenclature/config.py
+++ b/nomenclature/config.py
@@ -301,8 +301,8 @@ class TimeDomainConfig(BaseModel):
     def mixed_allowed(self) -> bool:
         return self.year_allowed and self.datetime_allowed
 
-    def check_datetime_format(self, df: IamDataFrame) -> None:
-        """Validate that datetime values conform to configured format and timezone."""
+    def check_datetime_timezone(self, df: IamDataFrame) -> None:
+        """Validate that datetime values use the configured timezone."""
 
         # Only check timezone if a specific timezone is required
         if not self.timezone:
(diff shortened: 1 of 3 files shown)


nomenclature/config.py
```diff
@@ -301,28 +301,23 @@ def validate_datetime_and_timezone(self) -> "TimeDomainConfig":
...
+    def check_datetime_timezone(self, df: IamDataFrame) -> None:
+        """Validate that datetime values use the configured timezone."""
+
+        # Only check timezone if a specific timezone is required
+        if not self.timezone:
+            return None
+
+        errors = [
+            TimeDomainError(f"{d} - invalid timezone") for d in
+            [d for d in df.time if isinstance(d, datetime)]
```
phackstock: I'm not a fan of using single character variables:
```suggestion
            [time for time in df.time if isinstance(time, datetime)]
```

nomenclature/config.py
```diff
@@ -301,28 +301,23 @@ def validate_datetime_and_timezone(self) -> "TimeDomainConfig":
...
-                errors.append(TimeDomainError(f"{d} - missing timezone"))
+    def check_datetime_timezone(self, df: IamDataFrame) -> None:
+        """Validate that datetime values use the configured timezone."""
+
+        # Only check timezone if a specific timezone is required
+        if not self.timezone:
+            return None
+
+        errors = [
+            TimeDomainError(f"{d} - invalid timezone") for d in
```
```suggestion
            TimeDomainError(f"{time} - invalid timezone") for time in
```

nomenclature/config.py
```diff
@@ -301,28 +301,23 @@ def validate_datetime_and_timezone(self) -> "TimeDomainConfig":
...
+            [d for d in df.time if isinstance(d, datetime)]
+            if d.tzname() != self.timezone
```
```suggestion
            if time.tzname() != self.timezone
```

phackstock (approved):
One small stylistic comment, either way good to be merged from my side 👍

